PR #19, opened 2015-01-22
Fta new bundles

DimHelander:
Add FTA bundles for Lakapa project.


ZakarFin:
Thanks for the pull request! Can you clean up the translations that are not valid (finnish in most of the en/sv localization). You can just remove the locale files and remove the link from packages/.../bundle.js to only support a finnish setup OR add the translations OR tell me if this is really the way it's supposed to work (having finnish text shown on other languages as well). 

Also some basic documentation about functionalities the bundles provide would be great so others can check if it suits their needs and also about the configuration/dependencies they require to work. You can see some examples in:
* https://github.com/nls-oskari/oskari.org/blob/master/md/documentation/bundles/framework/featuredata2.md ->  http://oskari.org/documentation/bundles/framework/featuredata2
* https://github.com/nls-oskari/oskari.org/blob/master/md/documentation/bundles/framework/infobox.md -> http://oskari.org/documentation/bundles/framework/infobox

Seems at least this plugin requires configuration and won't work without it: https://github.com/dimenteq/oskari/blob/fta-new-bundles/bundles/liikennevirasto/bundle/lakapa-transport-selector/plugin/maptransportselectorplugin/LakapaTransportSelectorPlugin.js#L71-L73. You can make a new pull request to the oskari.org repository with the docs.

ZakarFin:
Somehow the fta-new-bundles branch had lost the commit history of Oskari. Added the liikennevirasto/lakapa-* bundles as new files because of this.

Also upgraded the folder structure to reflect the current Oskari structure.

PR #26, opened 2015-12-08
Lupapiste import

jokimaki:
Latest source code of Lupapiste application.

Lupapiste uses Oskari 1.19.3. This PR includes only the files that could be merged cleanly into current master: applications/lupapiste/*, bundles/lupapiste/*, packages/lupapiste/* and resources/lupapiste/*.

All modifications are available at https://github.com/lupapiste/oskari/tree/lupapiste.


ZakarFin:
Sorry for the long wait. We're trying to figure out how to merge bundles that are application specific. The bundles would need to be documented and working at least on the current Oskari version so others can see what the code is about, see if it's something that could be part of other applications and try it out. This makes it somewhat possible to update the code with changes to the "core Oskari". Otherwise there isn't really point in merging application specific code to the Oskari repository.

There's propably some misconfigured line changes going on with changes showing as +86,538 −68,701. See this https://help.github.com/articles/dealing-with-line-endings/. ~18k new lines of code in 6 commits sounds like an svn dump :) Guess that's ok'ish since it's all in lupapiste directories, but the code would need to be atleast updated to current Oskari version to be worth the merge.

jokimaki:
No problem, I'll make a new PR if our fork catches up with the current Oskari version.

Note that the current lupapiste bundle (which this PR would update) works with even more obsolete version of oskari. You might want to consider cleaning it up from your repository entirely, it might be more clutter than a useful example in this day and age :)

ZakarFin:
Nice, thanks! We'll try to do some overall cleaning on the repository during the summer to make it more accessible :)

PR #27, opened 2015-12-15
Alert for backend failure

hatapitk:
Just using simple alert since I guess this will likely happen if backend is not working as it should (that is what happened for us). At least better than no error message at all.


ZakarFin:
Thanks! This needs localization, but we'll take a look at it.

kuosman:
Thanks! Merged into develop.

PR #28, opened 2016-01-14
Show email field when modifying user as admin

hatapitk:
This is the frontend code related to https://github.com/nls-oskari/oskari-server/pull/28


ZakarFin:
Thanks, we will test and merge this and the serverside counterpart shortly

kuosman:
Thanks. Tested and merged.

PR #29, opened 2016-02-24
Infobox flyout bundle selected-featuredata

Ruahonen:
New infobox layout/bundle


ZakarFin:
Thanks! This has been merged a while back with some manual tuning that lost the git history of the files. Sorry about that.

PR #30, opened 2016-04-18
IE support

keskival:
Changed the tables to DIV-elements to support Internet Explorer. Using special binding for dropdown list inputs for the same reason. Also commented out the placeholder text from the text field in data input, because IE does not support it and fails.


ZakarFin:
Merged, thanks!

PR #31, opened 2016-10-26
Develop / Feedback service (Open311) extensions

kessu:
Feedback service (Open311) management for geojson geometry and
feedback service url+api_key+[extensions] params to view metadata (was earlier in env properties)


ZakarFin:
Thanks!

PR #36, opened 2017-02-10
Added implementation for Button _setVisibleImpl

heikkiy:
Added function for handling the visibility of a Oskari.userinterface.component.Button.


ZakarFin:
Thanks! Merged into develop so this will be included in the next release.

PR #51, opened 2017-03-31
Maprotator finetuning

nxholm:


bundles/mapping/maprotator/instance.js
```diff
@@ -84,8 +84,6 @@ Oskari.clazz.define("Oskari.mapping.maprotator.MapRotatorBundleInstance",
       this.plugin = plugin;
     },
     stop: function() {
-      this.plugin.teardownUI(true);
-      this.plugin = null;
```
ZakarFin: This should stop the plugin as well if it's started.

ZakarFin:
Hide UI in publisher throws a JS error

ZakarFin:
Using Hide UI and leaving the publisher (with cancel or publishing) will result in the mainmap having the north arrow. Also the published map shows the UI even if Hide UI has been selected.

Editing the map with "hidden ui" doesn't check the Hide UI like it should.

PR #60, opened 2017-04-19
Ah 601 legend

nxholm:


ZakarFin:
This is a pull request for showing map legends on embedded maps.

PR #61, opened 2017-04-20
Feature/statsgrid vector regions

kuosman:
Some modifications:
- show statsgrid areas vector
- added OptionalStyle handling for VectorLayerPlugin ol2
- map region clicks scrolls datatable to show selected area

Fixes:
- ol2 featureevent to return GeoJSON to JSON format (previous string).
- drawplugin ol2 swallowing events.
- grid sorting when using column name renderers


bundles/mapping/mapmodule/mapmodule.ol2.js
```diff
@@ -466,6 +466,8 @@ Oskari.clazz.define('Oskari.mapframework.ui.module.common.MapModule',
             olStyle.graphicWidth = size;
             olStyle.graphicHeight = size;
 
+            olStyle.fillOpacity = 0.8;
```
ZakarFin: Opacity should be passed as parameter instead of hardcoded here.

bundles/statistics/statsgrid2016/service/StateService.js
```diff
@@ -68,16 +81,41 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.StateService',
...
+         * Gets the active region.
+         * @method  @public getRegion
+         * @return {null|Number} null or selected region
+         */
+        getRegion: function(){
+            if(this.activeRegion) {
+                // return selected region
+                return this.activeRegion;
+            }
+            return null;
```
ZakarFin: Any reason why not just always do "return this.activeRegion;"?

bundles/statistics/statsgrid2016/service/StateService.js
```diff
@@ -244,7 +282,11 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.StateService',
             var eventBuilder = Oskari.eventBuilder('StatsGrid.IndicatorEvent');
             this.sandbox.notifyAll(eventBuilder(ind.datasource, ind.indicator, ind.selections));
 
-
+            // if region is selected then highlight it also
+            var region = this.getRegion();
+            if(region) {
```
ZakarFin: Why do you need to set a selected region here?

bundles/statistics/statsgrid2016/components/RegionsetViewer.js
```diff
@@ -0,0 +1,150 @@
...
+        });
+
+        me.service.on('FeatureEvent', function(event){
+            if(event.getParams().operation === 'click' && event.hasFeatures()) {
+                // resolve region
+                var features = event.getParams().features[0];
+                var region = features.geojson.features[0].properties.id;
+
+                if(me.instance.conf.scrollDatatable) {
+                    state.setComponentState('datatable', {
```
ZakarFin: Can this be done without the component state tracking?

api/framework/statsgrid2016/bundle.md
```diff
@@ -28,13 +28,18 @@ Indicator selector, regionset selector and a datagrid with the indicator data ar
           "type": "system"
         }],
         "grid": true,
-        allowClassification : true
+        "allowClassification" : true,
+        "vectorViewer": true,
+        "scrollDatatable": true
```
ZakarFin: Need to think the config a bit, I think the scrollDatatable isn't very good configurable thing. Also vectorViewer could be tied to the map NOT having the StatsLayerPlugin registered OR RegionsetViewer.js could be an alternative implementation of StatsLayerPlugin.

bundles/statistics/statsgrid2016/components/Datatable.js
```diff
@@ -376,13 +379,26 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Datatable', function(sandbox, l
...
             me.grid.select(event.getRegion());
+
+            var gridEl = me.mainEl.find('table.oskari-grid:visible');
+            var parent = gridEl.parents('.oskari-flyoutcontentcontainer');
+            var compState = state.getComponentState('datatable');
+            if(compState.scroll && parent.length>0) {
+                parent.scrollTop(0);
+                var row = parent.find('tr[data-id="'+event.getRegion()+'"]');
+                if(row.length > 0) {
+                    parent.scrollTop(row.position().top);
```
ZakarFin: Maybe do this as a function in grid?

ZakarFin: grid.select(id) could take another optional param grid.select(id, true) with true to scroll to the selected row?

bundles/mapping/mapmodule/mapmodule.ol2.js
kuosman: This case set default opacity only. If current function styleDef parameter has styleDef.image.opacity then use this: https://github.com/kuosman/oskari-frontend/blob/65b94834de3bc5116bc18fbbeb337ee0d6f4f752/bundles/mapping/mapmodule/mapmodule.ol2.js#L487-L489

bundles/statistics/statsgrid2016/components/Datatable.js
kuosman: This is not so simple. Scrolling needs to know Grid parent scrollable element (for calculate position), in statsgrid2016 this is div.oskari-flyoutcontentcontainer -element, other bundle this maybe can be a different. 

Suggestions:
- second param is handler function and handler function get first param the grid element ? In this case handler actually do same thing than now.
- second param is scrollable element, if second param is setted then do scrolling

ZakarFin: Good point, the scrollable element as second parameter sounds better to me. Let's go with that.

bundles/statistics/statsgrid2016/components/RegionsetViewer.js
ZakarFin: As background info: This was used to detect where the selection originated (grid/map) to prevent scrolling grid if the user clicked on the grid. Let's solve this by replacing "setComponentState"  with adding a "component id" to selectRegion-function and the event it sends out. Could be "triggeredBy" with a value as one of "map", "grid" (and in the future possibly "graph" or "rpc").

PR #65, opened 2017-04-25
Feature/some fixes for release

kuosman:
Fixes:
- Grid: sort when using column name renderer
- DrawPlugin.ol2: not swallow events
- VectorLayerPlugin.ol2: FeatuereEvent now returns JSON formatted GeoJSON
- VectorLayerPlugin.ol2: optionalStyle support

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -2,11 +2,14 @@
 
 ## 1.42.0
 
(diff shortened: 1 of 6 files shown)


ZakarFin:
The same improvements was also part of #61. No need to merge these anymore. Thanks!

PR #68, opened 2017-05-03
Maplegend visual fixes

nxholm:
- Added support for lefthanded publishing.

- Adjust popup location based on plugin position.

- Removed styles that made plugin to be in a offset with other plugins.

- Added tooltip on hover.


kuosman:
Thanks merged to develop and also included changes 1.42.0

PR #80, opened 2017-06-01
Register bundle for pti

hvisuri:
New paikkatietoikkuna-specific bundle that creates login and registration links as well as logout link.
Bundle also creates registration popup to give information about registration before directing to registration page.


bundles/paikkatietoikkuna/register/instance.js
```diff
@@ -0,0 +1,133 @@
...
+         * @return {String} the name for the component
+         */
+        getName: function () {
+            return 'Register';
+        },
+
+        /**
+         * @method startPlugin
+         */
+        startPlugin: function () {
```
ZakarFin: use afterStart() instead of startPlugin()

bundles/paikkatietoikkuna/register/instance.js
```diff
@@ -0,0 +1,133 @@
...
+            var me = this;
+
+            //TODO get urls from properties
+            //Delete when url comes from config
+            if (!me.termsUrl) {
```
ZakarFin: Use me.getConfiguration().termsUrl, me.getConfiguration().registerUrl instead

PR #85, opened 2017-06-06
Feature/drawplugin fixes

okauppinen:
The drawn figures are removed from the map when PlaceForm is closed by clicking x-icon (cancel). Cancel and x-icon have now same functionality which is handled in infobox eventhandler.
https://github.com/oskariorg/oskari-docs/issues/12

DrawPlugin now checks preconditions before trying to save the drawn figures on the map. A line should have 2 points or finished figure (double click) and an area should have 3 points or finished figure (double click).
https://github.com/oskariorg/oskari-docs/issues/11

Added currentDrawing variable to avoid unnecessary getDrawing calls


ZakarFin:
Solves oskariorg/oskari-docs#12

PR #86, opened 2017-06-06
Feature/thematic map fixes

kuosman:
Some fixes:

- Grid.js sorting fixed when checking subtable
- Published maps now also get vectorviewer config
- When selected same region again then previously highlighted region disappears

### DIFF
--- a/bundles/statistics/statsgrid2016/instance.js
+++ b/bundles/statistics/statsgrid2016/instance.js
@@ -285,7 +285,7 @@ Oskari.clazz.define(
             }
 
             if(state.activeRegion) {
-                service.selectRegion(state.activeRegion);
+                service.toggleRegion(state.activeRegion);
             }
 
             // if state says view was visible fire up the UI, otherwise close it
--- a/bundles/statistics/statsgrid2016/instance.js
+++ b/bundles/statistics/statsgrid2016/instance.js
@@ -299,8 +299,7 @@ Oskari.clazz.define(
             var state = {
                 indicators: [],
                 regionset: service.getRegionset(),
-                view: me.visible,
-                vectorViewer: (me.regionsetViewer !== null) ? true : false
+                view: me.visible
             };
             service.getIndicators().forEach(function(ind) {
                 state.indicators.push({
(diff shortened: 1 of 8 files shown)


bundles/statistics/statsgrid2016/instance.js
```diff
@@ -297,9 +297,10 @@ Oskari.clazz.define(
             var me = this;
             var service = this.statsService.getStateService();
             var state = {
-                indicators : [],
-                regionset : service.getRegionset(),
-                view :me.visible
+                indicators: [],
+                regionset: service.getRegionset(),
+                view: me.visible,
+                vectorViewer: (me.regionsetViewer !== null) ? true : false
```
ZakarFin: Don't store this in state as it's already part of the config

bundles/statistics/statsgrid2016/publisher/ClassificationTool.js
```diff
@@ -128,7 +128,8 @@ function() {
                     statsgrid: {
                         conf : {
                             allowClassification: me.state.enabled,
-                            legendLocation : this.getPlugin().getLocation()
+                            legendLocation : this.getPlugin().getLocation(),
+                            vectorViewer: statsGridState.vectorViewer
```
ZakarFin: Access through statsgrid conf instead of state

bundles/statistics/statsgrid2016/publisher/StatsTableTool.js
```diff
@@ -120,7 +120,8 @@ function() {
                 statsgrid: {
                     state: statsGridState,
                     conf : {
-                        grid: me.state.enabled
+                        grid: me.state.enabled,
+                        vectorViewer: statsGridState.vectorViewer
```
ZakarFin: Access through statsgrid conf instead of state

bundles/statistics/statsgrid2016/service/StateService.js
```diff
@@ -76,9 +76,9 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.StateService',
         selectRegion : function(region, componentId) {
```
ZakarFin: Should this be renamed toggleRegion or use with null region as parameter instead?

PR #94, opened 2017-06-27
update ol to v4.3.1

nxholm:
### 23.8:  
- update ol to version 4.3.1
- update shrinkwrap
- update custom oskari buildfiles for openlayers


ZakarFin:
This should include a new npm-shrinkwrap.json as package.json has changed. All the Grunt-plugins worked ok with just updating Grunt itself?

libraries/ol3/base.js
```diff
@@ -1497,7 +1497,11 @@ goog.inherits = function(childCtor, parentCtor) {
  * @return {*} The return value of the superclass method.
  */
 goog.base = function(me, opt_methodName, var_args) {
-  var caller = arguments.callee.caller;
+	function caller (){
+		return agruments.callee.caller;
```
ZakarFin: agruments looks like a typo here?

libraries/ol3/oskari/oskariasynctileimage.js
```diff
@@ -85,10 +87,7 @@ ol.source.OskariAsyncTileImage.prototype.getWFSTileCache_ = function() {
 };
 
 ol.source.OskariAsyncTileImage.prototype.purgeWFSTileCache_ = function() {
-    var me = this,
-        wfsTileCache = this.getWFSTileCache_(),
-        layerTileInfos = wfsTileCache.tileInfos,
-        lastTileSetIdentifier =  wfsTileCache.tileSetIdentifier;
+
```
ZakarFin: Empty function? Is there a reason it should stay?

tools/conf/ol3/ol-custom.json
```diff
@@ -61,7 +61,10 @@
       "externs/olx.js",
       "externs/proj4js.js",
       "externs/tilejson.js",
-      "externs/topojson.js"
+      "externs/topojson.js",
+	  "externs/bingmaps.js",
```
ZakarFin: I think these have been removed previously by design to minimize the size. Some reason to add these?

libraries/ol3/base.js
```diff
@@ -1497,7 +1497,11 @@ goog.inherits = function(childCtor, parentCtor) {
  * @return {*} The return value of the superclass method.
  */
 goog.base = function(me, opt_methodName, var_args) {
-  var caller = arguments.callee.caller;
+	function caller (){
+		return arguments.callee.caller;
+	}
+  var caller = caller();
```
ZakarFin: In the previous version arguments refer to the original function and with the change the caller will always be 'goog.base'?

ZakarFin: I'm not sure why we should have this in our libs. Maybe try removing it?

ZakarFin:
Check the issue with libraries/ol3/base.js. Otherwise looks good for merging.

PR #104, opened 2017-08-11
Draw request fixes

data-ux:
Added shape type validation for DrawTools.StartDrawingRequest. Fixed StartDrawingRequest type "Square" to actually draw square.

In reference to:
https://jira.nls.fi/browse/AH-3886
https://jira.nls.fi/browse/AH-3899

### DIFF
--- a/bundles/mapping/drawtools/instance.js
+++ b/bundles/mapping/drawtools/instance.js
@@ -37,6 +37,12 @@ function() {
     __name : 'DrawTools',
 
     /**
+     * @static
+     * @property __validShapeTypes
+     */
+    __validShapeTypes : ["Point","Circle","Polygon","Box","Square","LineString"],
+
+    /**
      * @method getName
      * @return {String} the name for the component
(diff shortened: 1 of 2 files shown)


ZakarFin:
I'd prefer making these case-insensitive, but the change itself looks good. @kuosman can merge this next week if you want to go the extra mile for case-insensitivity. That would require changes in drawplugin as well and a check if we send out the shape-type in event. The event should have the CamelCase-formatting if it includes the type info for not breaking anything.

data-ux:
@ZakarFin Do you mean the DrawRequest shape type parameter should accept strings in a case-insensitive way? Accepting "linestring" and so on? If so, why would that be preferable?

ZakarFin:
Yes, I think it would be more developer-friendly.

ZakarFin:
Then again having case-insensitivity on some features, but not all is misleading as well. Let's roll with the console warnings instead.

PR #112, opened 2017-09-06
Ol3 featuredata2 drawtools

nxholm:
- update jsts so we can use WKTReader
- featuredata2 draw-functionality now uses requests
- new optional boolean parameter in stopdrawingrequest, which controls if event should be sent or not
- new parameter crs in drawing event


bundles/mapping/drawtools/event/DrawingEvent.js
```diff
@@ -11,11 +11,13 @@ Oskari.clazz.define('Oskari.mapping.drawtools.event.DrawingEvent',
...
  * @param {Object} additional info like length of line or area of a polygon?
  */
-function(id, geojson, data, isFinished) {
+function(id, geojson, data, crs, isFinished, olObject) {
     this._id = id;
     this._geojson = geojson;
     this._data = data;
+    this._crs = crs;
     this._isFinished = isFinished;
+    this._olObject = olObject;
```
ZakarFin: Events should be serializable as JSON and shouldn't include any mapping lib specific values.

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -166,14 +179,14 @@ Oskari.clazz.define(
          * @param {String} id
          * @param {boolean} clearCurrent: if true, all selection will be removed from the map
          */
-        stopDrawing : function(id, clearCurrent) {
+        stopDrawing : function(id, clearCurrent, sendEvent) {
```
ZakarFin: The default should be that event is sent. Use suppressEvent -> if true, don't send event.

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -208,22 +219,24 @@ Oskari.clazz.define(
...
             var bufferedGeoJson = me.getFeaturesAsGeoJSON(bufferedFeatures);
 
             var data = {
                 length : me._length,
                 area : me._area,
                 buffer: me._buffer,
                 bufferedGeoJson: bufferedGeoJson,
-                shape: me._shape
+                shape: me._shape,
+                geometry: me.geometry
```
ZakarFin: Why is geometry added here and as parameter for event? It shouldn't be sent at all since it isn't serializable and geojson has the same info. Also CRS could be included in the geojson data.

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -327,11 +339,23 @@ Oskari.clazz.define(
                    me.sendDrawingEvent(me._id, optionsForDrawingEvent);
                    return geometry;
                  }
+            } else if (shape === 'point' || shape === 'Point') {
```
ZakarFin: This should be done to all or none of the types. And all the places that reference the shapes. I think we should only use 'Point' as with the other shapes.

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -340,7 +364,10 @@ Oskari.clazz.define(
                      me.sendDrawingEvent(me._id, optionsForDrawingEvent);
                      return geometry;
                  }
-            } else if(shape === 'Polygon') {
+            } else if(shape === 'Circle' && ! options.buffer) {
+                geometryType = 'Circle';
+                geometryFunction = ol.interaction.Draw.createRegularPolygon(400);
+            } else if(shape === 'Polygon' || shape === 'area') {
```
ZakarFin: What's the reason for adding 'area'?

packages/framework/bundle/featuredata2/bundle.js
```diff
@@ -45,6 +45,9 @@ Oskari.clazz.define("Oskari.mapframework.bundle.featuredata2.FeatureDataBundle",
         }, {
             "type": "text/javascript",
             "src": "../../../../bundles/framework/featuredata2/plugin/MapSelectionPlugin.js"
+        },{
+            "type": "text/javascript",
```
ZakarFin: The same bundle includes ol2 and ol3 version?

bundles/framework/featuredata2/plugin/MapSelectionPlugin.ol3.js
```diff
@@ -0,0 +1,259 @@
...
+        me.multipart = (me._config && me._config.multipart === true);
+    }, {
+
+        /**
+         * @method addListener
+         * Registers a listener that will be notified when a selection has been made.
+         * The function will receive the selection geometry as parameter (OpenLayers.Geometry).
+         * @param {Function} listenerFunction
+         */
+        addListener: function (listenerFunction) {
```
ZakarFin: Should be replaced with Oskari.makeObservable(this) and callers using plugin.on('eventname', listenerFunction)

bundles/framework/featuredata2/plugin/MapSelectionPlugin.ol3.js
```diff
@@ -0,0 +1,259 @@
+/**
+ * @class Oskari.mapframework.bundle.metadata.plugin.MapSelectionPlugin
+ *
+ * Provides functionality to draw a selection box on the map
+ */
+Oskari.clazz.define(
+    'Oskari.mapframework.bundle.featuredata2.plugin.MapSelectionPlugin',
```
ZakarFin: Seems this doesn't need to be plugin anymore. Are there any references to mapmodule or mapping lib specific code? I think there's references to modifyControls and drawlayers that doesn't appear to be initialized anywhere. Also seems to include lots of unused code.

bundles/mapping/drawtools/event/DrawingEvent.js
```diff
@@ -11,11 +11,13 @@ Oskari.clazz.define('Oskari.mapping.drawtools.event.DrawingEvent',
  * @param {Object} geojson drawn shape (includes buffer as property?)
  * @param {Object} additional info like length of line or area of a polygon?
  */
-function(id, geojson, data, isFinished) {
+function(id, geojson, data, crs, isFinished, olObject) {
     this._id = id;
     this._geojson = geojson;
     this._data = data;
+    this._crs = crs;
```
ZakarFin: CRS could be set to geojson instead of using a new parameter.

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -60,15 +61,27 @@ Oskari.clazz.define(
...
+            var styleDef = jQuery.extend({}, me._defaultStyle, styleForType);
+            me._styles[type] = me.getMapModule().getStyle(styleDef);
+          });
+        },
+        /**
+         * @Return{String} reference system as defined in GeoJSON format
+         * @method _getSRS
+         */
+        _getSRS: function () {
+            return {
```
ZakarFin: Why the complex structure when it's not even merged to the geojson?

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -166,14 +173,17 @@ Oskari.clazz.define(
...
+            if( typeof supressEvent === undefined ) {
+                supressEvent = false;
+            }
             var options = {
                 clearCurrent: clearCurrent,
                 isFinished: true
             };
             if(me._functionalityIds[id]) {
-                me.sendDrawingEvent(id, options);
+                supressEvent === true ? me.clearDrawing(id) : me.sendDrawingEvent(id, options);
```
ZakarFin: So if event isn't sent the clearDrawing() isn't called? Looks like a bug. See this for previous example of using suppressEvent: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapmodule/AbstractMapModule.js#L693-L703

bundles/mapping/drawtools/request/StopDrawingRequest.js
```diff
@@ -16,11 +16,15 @@ Oskari.clazz
          *        id for drawing as given in StartDrawingRequest
          * @param {boolean} clearCurrent
          *        true to remove the current drawing
+         * @param {boolean} clearCurrent
+         *        true to remove the current drawing
          */
 
-        function(id, clearCurrent) {
+        function(id, clearCurrent, sendEvent) {
```
ZakarFin: Sending the event is the default thing to do and you don't want to have 99% cases of using this request to have to include true as a third param. sendEvent should be suppressEvent which can be used in the corner cases where you actually don't want an event about the finalized geometry. Also when changing requests/events make sure you update the API-docs regarding the request/event and mention the change in api/CHANGELOG.md

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
nxholm: sendDrawingEvent clears the drawing also if it's specified in the options: https://github.com/nxholm/oskari-frontend/blob/41d293710a66673f75b736569436434832bb3945/bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js#L244

PR #119, opened 2017-09-13
Improved timeseries animation

data-ux:
Improvements:

- Animation will wait for frame to load before advancing
- One frame is buffered before starting playback
- User can choose what time interval to skip ahead on each frame

Functionality is mostly in AbstractMapModule and AbstractLayer. MapLayerPlaybackRequest is sent to start/stop animatiosn. The layer animates itself and notifies the playback control UI with new TimeseriesAnimationEvent. 

In reference to:
https://jira.nls.fi/browse/AH-3994


bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -140,6 +140,14 @@ Oskari.clazz.define(
         me.tilesToLoad = 0;
         me.errors = 0;
 
+        me._lastFrameLoadTime = 0;
+        me._scheduleNextTimestep = false;
+        me._currentTime = null;
+        me._mapModule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
```
ZakarFin: Mapmodule should not be references here. At the very least it should be passed as parameter, but I don't think that's a good idea either. Maybe move the functionality to mapmodule or a plugin instead and keep the logic out of the domain object.

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -1199,6 +1212,156 @@ Oskari.clazz.define(
...
+        },
+        /**
+         * @method _getTileUrls
+         * Get tile urls for given time
+         * @param {String} time time as ISO-string 
+         */
+        _getTileUrls: function (time) {
+            var OLlayers = this._mapModule.getOLMapLayers(this._id);
+            var urls = [];
+            OLlayers[0].grid.forEach(function (a) {
```
ZakarFin: References to mapping lib specific structures should be moved to mapmodule or plugins. This probably only works for ol2. There's a request for updating url parameters as well. Maybe that could be used instead.

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -1199,6 +1212,156 @@ Oskari.clazz.define(
...
+         * @method _getNextTimestep
+         * Get next timestep from timeseries
+         * @param {Number} stepInterval time as ISO-string 
+         */
+        _getNextTimestep(stepInterval){
+            var times = this.getAttributes().times;
+            var nextTime;
+            var projection;
+            if(!times) {
+                console.warn('layer does not have "times" attribute');
```
ZakarFin: Use Oskari.log('[name]').warn() instead of direct console refs.

bundles/framework/timeseries/TimeseriesAnimationPlugin.js
```diff
@@ -0,0 +1,244 @@
...
+                }
+            }
+            me._scheduleNextTimestep = false;
+        },
+        /**
+         * @method _getTileUrls
+         * Get tile urls for given time
+         * @param {String} time time as ISO-string 
+         */
+        _getTileUrls: function (time) {
```
data-ux: There was no way to get the current tile urls from OpenLayers' public API, so had to resort to this somewhat hacky solution.

bundles/framework/timeseries/TimeseriesAnimationPlugin.js
```diff
@@ -0,0 +1,244 @@
...
+            var OLlayers = this._mapModule.getOLMapLayers(this._currentLayerId);
+            var urls = [];
+            OLlayers[0].grid.forEach(function (a) {
```
data-ux: There was no way to get the current tile urls from OpenLayers' public API, so had to resort to this somewhat hacky solution.

bundles/mapping/mapmodule/AbstractMapModule.js
```diff
@@ -940,7 +940,7 @@ Oskari.clazz.define(
 
           this.loadtimer = setTimeout( function() {
             var eventBuilder = Oskari.eventBuilder( 'ProgressEvent' );
-            var event = eventBuilder( done, 'maplayer' );
+            var event = eventBuilder(done, layerId);
```
ZakarFin: 'maplayer' is used to detect _what_ is being loaded. Need to take a better look at it if we'll just change it like this or add layerId as third param. Currently it's under mapmodule which is probably a wrong place if it could be used more commonly.

ZakarFin:
Looks good. Still have to figure out ProgressEvent change and check if the docs need to be updated:

- https://github.com/oskariorg/oskari-frontend/tree/master/api/mapping/timeseries (shown here http://oskari.org/api/bundles#1.43.0/mapping/timeseries)
- https://github.com/oskariorg/oskari-frontend/blob/master/api/mapping/mapmodule/event/ProgressEvent.md (shown here http://oskari.org/api/events#1.43.0/mapping/mapmodule/event/ProgressEvent.md)

data-ux:
Added docs & updated CHANGELOG with change to ProgressEvent. After the change in AbstractMapModule, the event is used as it is documented, but if someone relies on the ID field always having the string "maplayer", they can read about it in CHANGELOG.

bundles/framework/timeseries/TimeseriesAnimationPlugin.js
ZakarFin: We need to rethink this for OL3+, but for now it's ok.

PR #124, opened 2017-09-21
Feature/layerselector2 filter fix

kuosman:
Fixes:
- fixed filter buttons to show each tabs
- fixed undefined error for ShowFilteredLayerListRequest

Improvements:
- fastened layers filtering (before maded dublicate layer rendering)

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -8,15 +8,10 @@ Fixed filter buttons to show each tabs. Also fixed undefined error for ShowFilte
 
 Changed ``stats`` filter name to ``vectors`` (because it's filter vector layers).
 ```javascript
-var sb = Oskari.getSandbox();
-var reqBuilder = sb.getRequestBuilder('ShowFilteredLayerListRequest');
-if (reqBuilder) {
-    var request = reqBuilder(null, 'vectors', false);
-    sb.request('MainMapModule', request);
-}
+Oskari.getSandbox().postRequestByName('ShowFilteredLayerListRequest', [null, 'vectors']);
 ```
 
-Fastened layers filtering (before made duplicate layer rendering).
+Filtering performance has been improved.
 
 ### Timeseries improvements
 
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -13,6 +13,45 @@ Oskari.getSandbox().postRequestByName('ShowFilteredLayerListRequest', [null, 'fe
 
 Filtering performance has been improved.
 
+Requests should be serializable to JSON and shouldn't be used to pass functions. AddLayerListFilterRequest and ShowFilteredLayerListRequest refactored so at there is no anymore function parameters.
+Filter-functions can be registered to MapLayerService and it will have built-in filters for 'featuredata' and 'newest' ids.
+
+Use built-in filters:
+```javascript
+Oskari.getSandbox().postRequestByName('ShowFilteredLayerListRequest', ['newest', false]);
+```
+
+Register new filter and use this:
+```javascript
+// Register new filter
+var mapLayerService = Oskari.getSandbox().getService('Oskari.mapframework.service.MapLayerService');
+mapLayerService.registerLayerFilter('find_layers_name_start_a', function(layer){
+    var name = layer.getName().toLowerCase(),
+            nameFirstChar = name.substring(0,1);
+        return (nameFirstChar === 'a');
+});
+// Use new filter by request
+Oskari.getSandbox().postRequestByName('ShowFilteredLayerListRequest', ['find_layers_name_start_a', false]);
+```
+
+Add new layer lis filter button:
+```javascript
+// Add layer filter to map layer service
+var mapLayerService = Oskari.getSandbox().getService('Oskari.mapframework.service.MapLayerService');
+mapLayerService.registerLayerFilter('publishable', function(layer){
+    return (layer.getPermission('publish') === 'publication_permission_ok');
+});
+
+// Add layerlist filter button
+Oskari.getSandbox().postRequestByName('AddLayerListFilterRequest', [
+        'Publishable',
+        'Show publishable layers',
+        'layer-publishable',
+        'layer-publishable-disabled',
+        'publishable'
+]);
+```
+
 ### Timeseries improvements
 
 Animation now waits for frame to load before advancing.
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -16,6 +16,8 @@ if (reqBuilder) {
 }
 ```
 
+Fastened layers filterings (before made duplicate layer rendering).
+
 ### Timeseries improvements
 
 Animation now waits for frame to load before advancing.
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -16,7 +16,7 @@ if (reqBuilder) {
 }
 ```
 
-Fastened layers filterings (before made duplicate layer rendering).
+Fastened layers filtering (before made duplicate layer rendering).
 
 ### Timeseries improvements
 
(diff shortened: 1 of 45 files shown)


ReleaseNotes.md
```diff
@@ -2,6 +2,22 @@
...
+Fixed filter buttons to show each tabs. Also fixed undefined error for ShowFilteredLayerListRequest.
+
+Changed ``stats`` filter name to ``vectors`` (because it's filter vector layers).
+```javascript
+var sb = Oskari.getSandbox();
+var reqBuilder = sb.getRequestBuilder('ShowFilteredLayerListRequest');
+if (reqBuilder) {
+    var request = reqBuilder(null, 'vectors', false);
+    sb.request('MainMapModule', request);
+}
```
ZakarFin: The example should be:

     Oskari.getSandbox().postRequestByName('ShowFilteredLayerListRequest', [null, 'vectors']);

In the original example sandbox.getRequestBuilder() is deprecated and Oskari.requestBuilder() should be used instead. Also since we are not a module sending the request it's better to use postRequestByName() instead of saying that it's MapModule sending the request. I guess there's a case to be made that the code is much shorter as well :)

ReleaseNotes.md
```diff
@@ -2,6 +2,22 @@
...
+```
+
+Fastened layers filtering (before made duplicate layer rendering).
```
ZakarFin: "Layer filtering is now faster" or "Filtering performance has been improved"

api/CHANGELOG.md
```diff
@@ -11,6 +11,18 @@ Some extra tags:
 
 ## 1.44
 
+### [mod] [breaking] ShowFilteredLayerListRequest
```
ZakarFin: Why is this a breaking change? Provide an example of "before" and "after"?

bundles/framework/layerselector2/Flyout.js
```diff
@@ -289,12 +292,19 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
             me.tabContainer = Oskari.clazz.create(
                 'Oskari.userinterface.component.TabContainer'
             );
+
+            // Add filter tab change listener
+            me.tabContainer.addTabChangeListener(function(previousTab, newTab) {
```
ZakarFin: Adds a new listener with each filter request? Looks like a memory leak. Can you check this?

bundles/framework/layerselector2/Flyout.js
```diff
@@ -240,11 +241,13 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
...
          * Hande selected filter request
          * @method  public enableFilter
-         * @param  {String} selectedFilter selected filter, can be a 'stats', 'newest' or 'publishable'
+         * @param  {String} selectedFilter selected filter, can be a 'vectors', 'newest' or 'publishable'
          */
         enableFilter: function(selectedFilter) {
+            // Trigger first founded tab filter --> this activate/deactivate filterings also another tabs
             var me = this,
-                filterButton = jQuery('.layer-filter .filter-'+selectedFilter).first(),
+                filterContainer = (me.layerTabs.length>0) ? me.layerTabs[0].getTabPanel().getContainer().find('.layerselector2-layer-filter') : jQuery('<div></div>'),
```
ZakarFin: Why is this always done on the first of the layerTabs? Looks like it could be a bug. Should it be the selected tab?

ZakarFin: Also why would you wan't to create a new div if there are no layerTabs? Just stop the execution if there are none?

bundles/framework/layerselector2/Flyout.js
ZakarFin: Nevermind - this look ok. GitHub diff UI looked like this was done in enableFilter().

ZakarFin:
Requests should be serializable to JSON and shouldn't be used to pass functions. In a discussion not visible in GitHub it was agreed that the request will be refactored to only include filterId string and a boolean indicating if the layer listing UI should be shown automatically. Filter-functions can be registered to MapLayerService and it will have built-in filters for 'featuredata' and 'newest' ids. These can be used to get layers that have featuredata or are the latest layers registered to the instance.

bundles/framework/layerselector2/Flyout.js
```diff
@@ -510,75 +471,127 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
...
+        /**
+         * Get currently selected filter.
+         * @method  @private _getCurrentFilter
+         * @param  {String} filterName filter name
+         * @return {Object|null}  filter object if founded or null
+         */
+        _getCurrentFilter: function(filterName) {
+            var me = this;
+            // Checks at if filter founded then return this
+            var currentFilter = jQuery.grep(me.filters || [], function(filter){
```
ZakarFin: Maybe use the native forEach instead of jQuery as the logic is pretty much the same?

bundles/framework/layerselector2/Flyout.js
```diff
@@ -588,19 +601,26 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
...
+            me.layerTabs.forEach(function(tab, tabIndex){
+                var filterContainer = tab.getTabPanel().getContainer().find('.layerselector2-layer-filter');
+                jQuery.each(me.filters, function(index, filter) {
+                    if(!notDeactivateThisFilter || filter.filterName !== notDeactivateThisFilter) {
+                        var filterIcon = filterContainer.find('.filter-icon.' + 'filter-'+filter.filterName);
+                        filterIcon.removeClass(filter.iconClassActive);
+                        filterIcon.removeClass('active');
+                        filterIcon.addClass(filter.iconClassDeactive);
+                        // Set tooltip for one per filter
+                        if(tabIndex === 0) {
```
ZakarFin: Only set tooltips for the first tab?

api/mapping/layerselector/request/addlayerlistfilterrequest.md
```diff
@@ -10,19 +10,18 @@ The filter name parameter is used in ShowFilteredLayerListRequest request when a
...
+// Add layerlist filter button
+Oskari.getSandbox().postRequestByName('AddLayerListFilterRequest', [
         'Publishable',
         'Show publishable layers',
-        function(layer){
-            return (layer.getPermission('publish') === 'publication_permission_ok');
-        },
         'layer-publishable',
         'layer-publishable-disabled',
         'publishable'
```
ZakarFin: Can you add an example of the CSS referenced by 'layer-publishable' and 'layer-publishable-disabled'?

api/mapping/layerselector/request/showfilteredlayerlistrequest.md
```diff
@@ -9,69 +9,53 @@ or selected filter as well as optional paramater, which is open layer list.
...
-* stats , this shows only analysable layers on layerlist
-* publishable, this shows only publishable layers on layerlist
+* featuredata, this shows only featuredata layers on layerlist
+
+Bundles registered filters:
+* publishable, this shows only publishable layers on layerlist. Available only then if publisher2 bundle started and user is logged in.
 
 You can use default filters only when filter buttons are showed on layerlist.
 
 Filter layerlist to show only newest layers using the code below:
```
ZakarFin: I think the examples having false as second parameter should only send one param and have one of them use true with description what it does (opens the flyout?).

bundles/framework/layerselector2/Flyout.js
kuosman: _setFilterTooltip-function already adds tooltips for all tabs so there needs call this function once time per filter.

PR #140, opened 2017-10-03
Statsgrid/filter

nxholm:
## Desc ##
This pull request separates the thematic maps indicator selection and different types of result viewing to separate UI windows/flyouts.

### Changes ###
- Flyout has been replaced by a Flyout Manager that controls several extra flyouts
- Views (each flyout takes a view-field in options to map correct view to the flyout)
- Custom tile-implementation that determines onClick which flyout to open
- Search view, search indicators
- Tablevisualizer view, displays the data based on search results in table format.
- Diagramvisualizer view, displays the data based on search results as a diagram.
- Filter view, implemented later

Note! There are still some UI-tuning and code-refactoring that needs to be done after this PR.


bundles/statistics/statsgrid2016/FlyoutManager.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+    initFlyouts: function () {
+        var me = this;
+        me.flyouts = {
+            search: Oskari.clazz.create('Oskari.userinterface.extension.ExtraFlyout', this.instance.getLocalization().tile.title, {
+                width: 'auto',
+                cls: 'statsgrid-search-flyout',
+                view:'search',
+                pos: {
+                    x: 210,
```
ZakarFin: These should be calculated based on the size of the left bar or you can use the position of the map-container. The following flyouts should get their position from the previous x + width and detect if they are still on screen. The flyouts can overlap if they don't fit the screen side by side.

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -40,7 +40,7 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function(sandbox, lo
                 verticalCenterPadding: 8
             },
             width: function () {
-                return 960 - this.margin.left - this.margin.right;
+                return 380 - this.margin.left - this.margin.right;
```
ZakarFin: What is this magic number and shouldn't width be configurable?

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -49,58 +59,53 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function(sandbox, lo
...
     /**
      * initializes the chart skeleton without any specific line or bar options
      *
      * @method initChart
      *
      */
-    initChart: function() {
-        var selections = this.initSelection();
+    initChart: function () {
+        var svg = this.initSVG();
```
ZakarFin: I would change initSVG() to getSVGTemplate(), remove assignment of this.svg from the initSVG() implementation and assign it in here if it's needed.

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -132,27 +137,41 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function(sandbox, lo
...
+            .attr( "class", "x axis" )
+            .call( this.xAxis );
+    },
+    /**
+     * @method setColorScale
+     */
+    setColorScale: function ( colors ) {
+        this.colorScale = d3.scaleQuantize()
+            .domain( [0, this.data.length] )
+            .range( colors );
```
ZakarFin: When used with thematic maps this is problematic. This component needs a method where the colorScale can be passed by the code using this component.

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -49,58 +59,53 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function(sandbox, lo
...
-    initSelection: function() {
-            var svg = d3.select(this._template.graph[0]).append("svg")
-                .attr("width", this.dimensions.width() + this.dimensions.margin.left + this.dimensions.margin.right)
-                .attr("height", this.dimensions.height() + this.dimensions.margin.top + this.dimensions.margin.bottom)
-                .append("g")
-                .attr("transform", "translate(" + this.dimensions.margin.left + "," + this.dimensions.margin.top + ")"
-            );
-            this.svg = svg;
+    initSVG: function () {
+        var svg = d3.select( this._template.graph[0] ).append("svg")
```
ZakarFin: Use this._template.graph.clone() instead. That way you can remove any this._template.graph.empty() calls.

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -211,56 +233,54 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function(sandbox, lo
...
-                .call(me.createGridlines()
-                .tickSize(-me.dimensions.height() +30, 0, 0)
+                .call( me.createGridlines()
+                .tickSize( -me.dimensions.height() +30, 0, 0 )
                 .tickFormat("")
             );
         });
     },
+    clear: function () {
+        this._template.graph.empty();
```
ZakarFin: Use graph.clone() and remove this empty() call.

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -211,56 +233,54 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function(sandbox, lo
...
+            chart = this.createBarChart( this.data, opts );
         } else if( this.chartType === 'linechart' ) {
-            chart = this.createLineChart(this.data);
+            chart = this.createLineChart( this.data, opts );
         }
 
         return chart;
     },
     getGraph: function() {
         return this._template.graph;
```
ZakarFin: This variable should hold the template and not the finished graph.

bundles/framework/divmanazer/component/Grid.js
```diff
@@ -1049,6 +1035,26 @@ Oskari.clazz.define('Oskari.userinterface.component.Grid',
             }
         },
 
+        updateHeaderHeight: function(){
+            var me = this;
+            // autosize header
+            if(typeof me.autoHeightHeader === 'number' && me.table) {
```
ZakarFin: Reverse this if so instead of having 10+ rows of code inside the method just "fails fast"/returns at the beginning of the function if there's nothing to be done for the header.

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -60,86 +60,103 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
...
     },
-    /**@method selectFirstValue
-    *  makes the second value show instead of placeholder text
+    /** @method selectFirstValue
+    *   Select the first non-placeholder value
     */
-    selectFirstValue: function(){
+    selectFirstValue: function () {
       var chosen = this.element.find('select');
       chosen.find('option:nth-child(2)').attr('selected', 'selected');
```
ZakarFin: The second one is always the first? So there's always an "empty option" at first index?

bundles/framework/divmanazer/extension/ExtraFlyout.js
```diff
@@ -24,35 +24,40 @@ Oskari.clazz.define('Oskari.userinterface.extension.ExtraFlyout',
...
+	    		'	<div class="oskari-flyoutheading"></div>' +
+	    		'	<div class="oskari-flyout-title"><p></p></div>' +
+	    		'	<div class="oskari-flyouttools">' +
+	    		'		<div class="oskari-flyouttool-close icon-close icon-close:hover"></div>' +
+	    		'	</div>' +
+	    		'</div>' +
+	    		'<div class="oskari-flyoutcontentcontainer"></div>' +
+	    		'</div>')
+	    },
+        __temp:{
```
ZakarFin: Why is this under __temp and not in __templates?

bundles/framework/divmanazer/extension/ExtraFlyout.js
```diff
@@ -177,5 +182,99 @@ Oskari.clazz.define('Oskari.userinterface.extension.ExtraFlyout',
...
+        },
+        getElement: function(){
+            return this._popup;
+        },
+
+/************************************************************************************************
+* Side tool functions
+************************************************************************************************/
+         //this function collects the label and calls the default flyout function addSideTool sending the label and the callback function.
+        getSideLabel : function( text ) {
```
ZakarFin: createSideLabel() or createSideLabelElement() would be more fitting name.

bundles/framework/divmanazer/extension/ExtraFlyout.js
```diff
@@ -177,5 +182,99 @@ Oskari.clazz.define('Oskari.userinterface.extension.ExtraFlyout',
...
+            var sidelabel = jQuery(this.__temp.sideTool());
+            sidelabel.find('label').text(text);
+            return sidelabel;
+        },
+        _calcSideLabelPositions: function(){
```
ZakarFin: _refreshSideLabelPositions() could be more descriptive name?

ZakarFin: or _updateSideLabelPositions()

bundles/framework/divmanazer/component/Chart.js
nxholm: d3.select needs a string or a dom-node to append to, so jQuery won't work here.
https://github.com/d3/d3-selection/blob/master/README.md#select

ZakarFin: But if you clone() the DOM-element generated before passing it to d3 you stop modifying the template variable and won’t need to ”clear()” it in other pieces of the code. So either clone it or stop calling it template.

ZakarFin:
This needs a follow-up PR to fix UI issues and some code refactoring, but as the changes are massive already reviewing any changes will only get more difficult. We are determined to fix the issues before the next release and to make future changes easier to check the functionality is merged at this point even when it's not yet "production ready"

PR #145, opened 2017-10-09
OL3 metadatacatalogue

data-ux:
Removed `MetaData.FinishedDrawingEvent`. Uses `DrawTools.StartDrawingRequest` and `DrawTools.StopDrawingRequest` instead to draw search coverage area on map.

Can be tested in "Geoportal OL3" view.


ZakarFin:
This overwrites the OpenLayers 2 implementation. For this to be merged the whole geoportal needs to be upgraded to OpenLayers 3. Please restore the old instance.js and have the changes in for example  instance.ol3.js with another bundle.js under packages so we can merge this, use OpenLayers 2 for Geoportal for now AND develop ol3 version in the side before migrating the whole geoportal.

data-ux:
This still needs some updated handling of the coverage GeoJSON on the server side.

data-ux:
OpenLayers 2 implementation is back. Server side fixed in:
https://github.com/oskariorg/oskari-server/pull/105

PR #148, opened 2017-10-18
Added some more input sanitization to data fetched from backend apis.

tsallinen:
This prevents some security issues if invalid data is passed from backend.


ZakarFin:
Thanks!

PR #151, opened 2017-10-20
New paikkatietoikkuna specific bundle "coordinatetransformation"

nxholm:
- new bundle coordinatetransformation
- some unfinished functionality has been disabled for now

### This is not a generic functionality for oskari-frontend. When community modules repository has been established this can be moved there.


bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -0,0 +1,135 @@
...
+ *
+ */
+Oskari.clazz.define('Oskari.userinterface.component.FileInput', function (localization) {
+    this.loc = localization;
+    this.el = null;
+    this._template = {
+        fileBox: _.template('<div class="oskari-fileinput" style="display:none"> '+
+                    '<form method="post" action="", enctype="multipart/form-data" class="box" id="fileinput">'+
+                        '<div class="box__input">'+
+                            '<input type="file" name="file" id="file" class="box__file" />'+
```
ZakarFin: Remove id and name or atleast make them much more unique

bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -0,0 +1,135 @@
+/**
+ * @class Oskari.userinterface.component.FileInput
+ *
+ */
+Oskari.clazz.define('Oskari.userinterface.component.FileInput', function (localization) {
+    this.loc = localization;
+    this.el = null;
+    this._template = {
+        fileBox: _.template('<div class="oskari-fileinput" style="display:none"> '+
+                    '<form method="post" action="", enctype="multipart/form-data" class="box" id="fileinput">'+
```
ZakarFin: Remove id or make it much more unique

bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -0,0 +1,135 @@
...
+        },
+        /**
+         * @method canUseAdvancedUpload
+         *
+         * Checks if the browser supports drag and drop events aswell as formdata & filereader
+         * @return {boolean} true if supported 
+         */
+        canUseAdvancedUpload: function() {
+            var div = document.createElement('div');
+            return (('draggable' in div) || ('ondragstart' in div && 'ondrop' in div)) && 'FormData' in window && 'FileReader' in window;
```
ZakarFin: Maybe add some linebreaks here to make it more readable?

bundles/framework/divmanazer/component/FileInput.js
ZakarFin: That comma after action looks like it should be removed.

bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -0,0 +1,135 @@
...
+    this._template = {
+        fileBox: _.template('<div class="oskari-fileinput" style="display:none"> '+
+                    '<form method="post" action="", enctype="multipart/form-data" class="box" id="fileinput">'+
+                        '<div class="box__input">'+
+                            '<input type="file" name="file" id="file" class="box__file" />'+
+                            '<label>  <%= fileupload %> <label for="file" style="cursor: pointer;"> <a> <%= link %> </a> </label> </label> '+
+                        '</div>'+
+                        '<div class="box__uploading"> <%= uploading %>&hellip;</div>'+
+                        '<div class="box__success"><%= success %></div>'+
+                        '<div class="box__error"><%= error %></div>'+
```
ZakarFin: Are these some message placeholders that are toggled visible/hidden? Maybe use just one with an extra class signifying the severity?

bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -0,0 +1,135 @@
...
+                e.stopPropagation();
+            })
+            .on('dragover dragenter', function() {
+                form.addClass('is-dragover');
+            })
+            .on('dragleave dragend drop', function() {
+                form.removeClass('is-dragover');
+            })
+            .on('drop', function(e) {
+                droppedFiles = e.originalEvent.dataTransfer.files;
```
ZakarFin: If this component only handles single file, should we check that the user only dropped one file?

bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -0,0 +1,135 @@
+/**
+ * @class Oskari.userinterface.component.FileInput
+ *
```
ZakarFin: Document usage instructions and the required localization here. Also a brief description would be nice.

ZakarFin: The individual methods would also benefit from some comments.

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -41,6 +41,9 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
         if(!dataKey.id && !dataKey.title){
           option.val(dataKey).text(dataKey);
         }
+        if(dataKey.cls) {
```
ZakarFin: This should be documented on the method description.

bundles/paikkatietoikkuna/coordinateconversion/Flyout.js
```diff
@@ -0,0 +1,71 @@
+Oskari.clazz.define('Oskari.coordinateconversion.Flyout',
+
+    /**
+     * @method create called automatically on construction
+     * @static
+     * @param {Oskari.framework.bundle.admin-layerrights.AdminLayerRightsBundleInstance}
```
ZakarFin: admin-layerrights?

bundles/paikkatietoikkuna/coordinateconversion/Tile.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * @class Oskari.bundle.cooridinateconversion.Tile
```
ZakarFin: Does this bundle really need a custom tile implementation? If not you should just extend the defaultextension/defaulttile. Also a typo on the cooridinateconversion.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+            return this.dropdowns;
+        },
+        create: function () {
+            var json = this.instance.helper.getOptionsJSON();
+
+            var selections = [];
+            var dropdowns = [];
+            var selects = [];
+            var options = {}
+            jQuery.each( json, function ( key, value ) {
```
ZakarFin: Consider using json.forEach() or Object.keys(json).forEach() instead of jQuery

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                var size = Object.keys( value ).length;
+                 jQuery.each( value, function ( key, val ) {
+                    var valObject = {
+                        id : val.id,
+                        title : val.title,
+                        cls: val.cls
+                    };
+                    selections.push( valObject );
+                    if ( key === "0" ) {
```
ZakarFin: Is this the first element or what's going on here?

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                            width: '100%'
+                        };
+                    }
+                     if ( key == size -1 ) {
+                        var select = Oskari.clazz.create('Oskari.userinterface.component.SelectList', "id");
+                        var dropdown = select.create(selections, options);
+                        selections = [];
+                        options = {};
+
+                        dropdown.css( { width:'170px', float:'right' } );
```
ZakarFin: Shouldn't this be in the CSS?

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+         * [2] = map projection
+         * [3] = geodetic coordinate system
+         * [4] = heigth system
+         */
+        handleSelectionChanged: function ( element ) {
+            var me = this;
+            var values = [];
+            var instances = this.getSelect();
+            var dropdowns = this.getDropdown();
+            var rows = me.instance.inputTable.getElements().rows;
```
ZakarFin: Can these be just passed to the method instead of digging stuff from the instance?

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+        me.dropdowns = [];
+    }, {
+        getSelect: function () {
+            return this.selects;
+        },
+        getDropdown: function () {
+            return this.dropdowns;
+        },
+        create: function () {
+            var json = this.instance.helper.getOptionsJSON();
```
ZakarFin: Pass the JSON as parameter instead of doing it like this.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+         * [0] = geodetic datum,
+         * [1] = coordinate system
+         * [2] = map projection
+         * [3] = geodetic coordinate system
+         * [4] = heigth system
+         */
+        handleSelectionChanged: function ( element ) {
+            var me = this;
+            var values = [];
+            var instances = this.getSelect();
```
ZakarFin: The naming in here doesn't feel consistent. We get a select, but receives an array of instances as return value? Same on the next row with getDropdown() returns an array instead of a single dropdown.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+            var dropdowns = this.getDropdown();
+            var rows = me.instance.inputTable.getElements().rows;
+                element.on( "change", function() {
+                    for (var i = 0; i < instances.length; i++ ) {
+
+                    dropdowns[i].find('option').hide();
+                    dropdowns[i].find('.'+ instances[0].getValue()).show();
+
+                    // Koordinaatisto special cases
+                    if( instances[1].getValue() === "KOORDINAATISTO_MAANT_2D" ) {
```
ZakarFin: How about instances being an object with some readable names? This could be something like instances.something_readable.getValue()

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                        dropdowns[3].find('option').hide();
```
ZakarFin: Same goes for this dropdowns array. This would be much more readable than magic numbers.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                        jQuery(dropdowns[i].find('.'+instances[0].getValue()+'.'+ instances[1].getValue())).show();
+        
+                    }
+                    if( instances[1].getValue() === "KOORDINAATISTO_MAANT_3D" ) {
+                        dropdowns[3].find('option').hide();
+                        dropdowns[i].find('.'+ instances[1].getValue()).show();
```
ZakarFin: So some of the options have classes that are named after values in another dropdown? If it has to be like this please write a comment describing what's going on here.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                        dropdowns[3].find('option').hide();
+                        dropdowns[i].find('.'+ instances[2].getValue()).show();
+                    }
+                    if( instances[3].getValue() !== "COORDSYS_DEFAULT" ) {
+                        me.instance.startingSystem = true;
+                    } else {
+                        me.instance.startingSystem = true;
+                    }
+                    if( instances[4].getValue() === "KORKEUSJ_DEFAULT") {
+                        rows.each( function ( idx, row ) {
```
ZakarFin: Clean this up or add comments for what's going on here.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                    }
+                        
+                    if( instances[1].getValue() === 'KOORDINAATISTO_SUORAK_2D' ) {
+                        jQuery('.map-projection').show();
+                    } else {
+                        jQuery('.map-projection').hide();
+                        instances[2].resetToPlaceholder();
+                    }
+
+                    if( instances[0].getValue() !== me.instance.currentDatum ||me.instance.currentDatum === undefined ) {
```
ZakarFin: What's going on in here?

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                    } else {
+                        rows.each( function ( idx, row ) {
+                            var lastCell = jQuery(this).find('td:nth-last-child(2)');
+                            lastCell.attr("contenteditable", false);
+                            lastCell.removeClass('heightsystem');
+                        })
+                    }
+                        
+                    if( instances[1].getValue() === 'KOORDINAATISTO_SUORAK_2D' ) {
+                        jQuery('.map-projection').show();
```
ZakarFin: What is ".map-projection" and why is it searched from the whole document? Seems dangerous.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                        }
+                    }
+                    values = [];
+                    instances[i].update();
+                    for ( var j = 0; j < instances.length; j++ ) {
+                        var vl = instances[j].getValue();
+                        values.push( vl );
+                    }
+                    if( i == instances.length -1 ) {
+                        if( this.id === 'inputcoordsystem' ) {
```
ZakarFin: Is id set from outside the component? I can't find another reference to id in this components code, but somehow it's still referenced here.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,168 @@
...
+                            me.instance.inputTable.updateTitle( values );
+                            me.instance.inputTable.isEditable( me.instance.clipboardInsert );
+                        } else if( this.id === 'targetcoordsystem' ) {
+                            me.instance.outputTable.updateTitle( values );
```
ZakarFin: You should make this component observable and bind listeners outside of it for these kinds of things. Instead of triggering functions with deep reference structure outside this component.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
+Oskari.clazz.define('Oskari.coordinateconversion.component.table', function( instance, loc ) {
+    this.instance = instance;
+    this.loc = loc;
+    this.numrows = 1;
+    this.container = null;
+    this.template = {
+            tableWrapper: jQuery('<div class="table"></div>'),
```
ZakarFin: Maybe use lodash for all instead of jQuery for one and lodash for the rest?

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+                                    '<td class="cell lon" headers="north" style=" border: 1px solid black ;"> <%= coords.lon %> </td>'+
+                                    '<td class="cell lat" headers="east" style=" border: 1px solid black ;"> <%= coords.lat %> </td>'+
+                                    '<td class="cell height" headers="ellipse_height" style=" border: 1px solid black;"></td>'+
+                                    '<td class="cell control"> <div class="removerow"></div></td>'+
+                                '</tr> '),
+            table:_.template('<div class="coordinatefield-table" style="display:inline-block;">' +
+                                        '<h5> <%= input %> </h5>' +
+                                        '<div class="oskari-table-content">'+
+                                         '<div style="width:100%; overflow:auto;">'+
+                                            '<table class="hoverable" id="oskari-coordinate-table" style="border: 1px solid black;" cellpadding="0" cellspacing="0" border="0">'+
```
ZakarFin: Having two of these tables in the UI results in multiple DOM elements having the same id.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+                    jQuery(this).find('td').attr("contenteditable", false);
+                });
+            }
+            else {
+                rows.each( function ( row ) {
+                    jQuery(this).find('td').attr("contenteditable", true);
+                });
+            }
+        },
+        removeRow: function () {
```
ZakarFin: If this always removes the first row the name should be removeFirstRow.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+        /**
+         * @method populate
+         *
+         * {Object} data, each key need to have property lon & lat 
+         */
+        populate: function( cell, data ) {
+            var table = this.getElements().table;
+            var i = 0;
+            for ( var key in data ) {
+                if ( i <= 9 ) {
```
ZakarFin: Populates the first few rows? Seems like the limit should be somewhere on the calling side instead of this method. Or atleast make a variable like MAX_ROWS that's referenced here instead of a magic number.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+                    if ( data.hasOwnProperty( key ) ) {
+                        var row = this.template.row( { coords: data[key] } );
+                        jQuery( cell ).parent().after( row );
```
ZakarFin: I think the caller should send the cell.parent() instead. Would make a lot more sense to send "previous row" instead of a cell.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                }
+            }
+            table.trigger('rowCountChanged');
+        },
+        addRows: function (coords) {
+            var table = this.getElements().table;
+            for (var i = 0; i < coords.length; i++ ) {
+                var row = this.template.row( { coords: coords[i] } );
+                table.find('tr:first').after(row);
```
ZakarFin: Why tr:first? New row always comes to the top of the table? Maybe the function name should be renamed to indicate that.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+            var cells = this.getContainer().find('tr .cell:not(.cell.control)');
+            for(var i = 0; i < cells.length; i++) {
+                var trimmedCellValue = jQuery.trim( jQuery( cells[i] ).html() );
+                if( trimmedCellValue !== "" ) {
+                    jQuery( cells[i] ).parent().remove();
+                    this.decrementNumRows();
+                }
+            }
+        },
+        displayNumOfRows: function () {
```
ZakarFin: Seems "displayNumOfRows" doesn't really display anything. Instead it binds a listener that hides any row after the first 1000.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+            var me = this;
+            var table = this.getElements().table;
+            table.bind('rowCountChanged', function (evt) {
+                var rows = me.getElements().rows;
+                    if( rows.length >= 1000 ) {
+                        rows.slice(1000).css('display', 'none');
```
ZakarFin: Please find another way of doing this. Like removing excess rows when new ones are added. This seems like a very poor solution for the problem of having too many rows on the table.

bundles/paikkatietoikkuna/coordinateconversion/components/table.js
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                });
+        },
+        updateTitle: function (values) {
+            this.getElements().header.remove();
+            
+            var x = y = z = "";
+            if( values[3].indexOf("COORDSYS_KKJ" ) !== -1 ) {
```
ZakarFin: More magic numbers for the maintainer to enjoy? :)

bundles/paikkatietoikkuna/coordinateconversion/instance.js
```diff
@@ -0,0 +1,186 @@
...
+            var l = appSetup.startupSequence.length;
+            appSetup.startupSequence[l] = {
+                "bundleinstancename":"coordinateconversion",
+                "bundlename":"coordinateconversion" ,
+                "fi":"",
+                "sv":"",
+                "en":"",
+                "title":"coordinateconversion" 
+            }
+            appSetup.startupSequence[l].metadata= { "Import-Bundle": { "coordinateconversion": { "bundlePath": "/Oskari/packages/paikkatietoikkuna/bundle/" } } };
```
ZakarFin: Maybe add some comments what these are. Also startupSequence.push() would be more readable. You don't need fi, sv, en, title, bundleinstancename here.

bundles/paikkatietoikkuna/coordinateconversion/components/select.js
```diff
@@ -0,0 +1,164 @@
...
+                this.selectInstances = selects;
+        },
+        /**
+         * @method handleSelectionChanged
+         * which key corresponds to which dropdown in the array:
+         * [0] = geodetic datum,
+         * [1] = coordinate system
+         * [2] = map projection
+         * [3] = geodetic coordinate system
+         * [4] = heigth system
```
ZakarFin: The comment about the array indexes is now outdated

bundles/paikkatietoikkuna/coordinateconversion/resources/css/coordinateconversion.css
```diff
@@ -0,0 +1,322 @@
+.oskari-flyoutcontentcontainer{
```
ZakarFin: Make this affect only coordinateconversions flyout.

ZakarFin: Actually prefix all the styles so they only affect this flyout.

bundles/paikkatietoikkuna/coordinateconversion/service/ConversionService.js
```diff
@@ -0,0 +1,68 @@
...
+ * @method create called automatically on construction
+ * @static
+ */
+function(instance) {
+    this._instance = instance;
+    this._sandbox = instance.sandbox;
+    this.urls = {};
+
+    var ajaxUrl = this._sandbox.getAjaxUrl() + '&action_route=';
+    this.urls.reverseGeocode = (ajaxUrl + 'GetConversionResult');
```
ZakarFin: You can replace this with Oskari.urls.getRoute('GetConversionResult'). It will return the same as sandbox.getAjaxUrl() + '&action_route=' + GetConversionResult;

bundles/paikkatietoikkuna/coordinateconversion/resources/css/coordinateconversion.css
```diff
@@ -0,0 +1,322 @@
...
+}
+.dropdown_title {
+    float: left;
+    font-weight: bold;
+    word-break: break-word;
+    max-width:12em !important;
+}
+.infolink {
+    float:right;
+    background-image: url('../images/icon-info.png');
```
ZakarFin: You can use icon-info class for the component so you don't need to add this image to the bundle resources. The icon is part of the base set of icons.

bundles/paikkatietoikkuna/coordinateconversion/resources/css/coordinateconversion.css
```diff
@@ -0,0 +1,322 @@
...
+.hoverable tr:hover .cell.control { 
+    /* float: right; */
+    /* display: inline-block; */
+}
+
+.removerow {
+    border-radius: 50%;
+    height:1.3em;
+    width:1.4em;
+    background-image: url('../images/icon-close.png');
```
ZakarFin: You can use icon-close class for the component so you don't need to add this image to the bundle resources. The icon is part of the base set of icons.

PR #152, opened 2017-10-24
Generalized timeseries

data-ux:
- Adds generalized timeseries UI that any other bundle can use with TimeseriesService
- Adds support for registering layer animation implementations with TimeseriesLayerService (via TimeseriesService)
- Adds implementation for animating WMS layers (via TimeseriesLayerService)

In reference to: 
https://jira.nls.fi/browse/AH-3995


bundles/framework/timeseries/WMSAnimator.js
```diff
@@ -0,0 +1,180 @@
...
+            var imgCount = urls.length;
+            if (imgCount === 0) {
+                callback(true);
+                return;
+            }
+            var aborted = false;
+            var timeout = setTimeout(function () {
+                aborted = true;
+                callback(false);
+            }, 5000);
```
ZakarFin: Should the timeout be configurable with layer.attributes.imageTimeout or similar and default to 5 seconds if not configured?

bundles/framework/timeseries/WMSAnimator.js
```diff
@@ -0,0 +1,180 @@
...
+        requestNewTime: function (newTime, nextTime, doneCallback) {
+            var me = this;
+            this._currentTime = newTime;
+            var requestBuilder = this._sandbox.getRequestBuilder('MapModulePlugin.MapLayerUpdateRequest');
+            if (requestBuilder) {
+                this._isLoading = true;
+                this._doneCallback = doneCallback;
+                if (nextTime) {
+                    this._bufferImages(this._mapModule.getLayerTileUrls(this._layer.getId()), nextTime, function (success) {
+                        me._isBuffering = false;
```
ZakarFin: This isBuffering handling should be moved to _bufferImages() so it keeps track of the status instead of external code that's not actually doing the buffering.

bundles/framework/timeseries/WMSAnimator.js
ZakarFin: Maybe not since the timeout is just for buffering. The caller doesn't seem to mind if buffering timed out.

bundles/framework/timeseries/WMSAnimator.js
```diff
@@ -0,0 +1,180 @@
...
+         * @method requestNewTime
+         * Requests change in current selected time
+         * @param {String} newTime change current time to this value(ISO string)
+         * @param {String} nextTime time value at next animation frame(ISO string). Can be null if not animating
+         * @param {function} doneCallback callback that will be called after new time has been loaded
+         */
+        requestNewTime: function (newTime, nextTime, doneCallback) {
+            var me = this;
+            this._currentTime = newTime;
+            var requestBuilder = this._sandbox.getRequestBuilder('MapModulePlugin.MapLayerUpdateRequest');
```
ZakarFin: Use Oskari.requestBuilder() instead as sandbox.getRequestBuilder() has been deprecated. If you want to know if someone is actually handling the request you can check with sandbox.hasHandler('MapModulePlugin.MapLayerUpdateRequest').

bundles/framework/timeseries/WMSAnimator.js
```diff
@@ -0,0 +1,180 @@
...
+         */
+        getSubsetRange: function () {
+            return this._subsetRange.slice();
+        },
+        /**
+         * @method getSubsetRange
+         * Set current selected time range subset
+         * @param {String[]} range Array with 2 elements, start & end, ISO-string
+         */
+        setSubsetRange: function (range) {
```
ZakarFin: Should this check that the parameter is an array with 2 elements?

bundles/framework/timeseries/instance.js
```diff
@@ -80,161 +41,78 @@ Oskari.clazz.define("Oskari.mapframework.bundle.timeseries.TimeseriesToolBundleI
...
-         * @method initTimeSeries
-         * inits the timeline ui with values of a layer supporting times
-         */
-        initTimeseries: function(layer) {
-            var me = this,
-                times = layer.getAttributes().times,
-                layerId = layer.getId();
+            me._timeseriesLayerService = Oskari.clazz.create('Oskari.mapframework.bundle.timeseries.TimeseriesLayerService', sandbox, me._timeseriesService);
+            sandbox.registerService(me._timeseriesLayerService);
+            me._timeseriesLayerService.registerLayerType('wms', function (layerId) {
```
ZakarFin: This should probably be in the wmslayerplugin, but since wms has "hardcoded" support in maplayerservice it's fine in here too.

bundles/framework/timeseries/resources/css/timeseriesplayback.css
```diff
@@ -1,135 +1,197 @@
...
-  div.mapplugin.timeseries .oskari-timeslider-empty {
-    margin-top: 10px;
-    float: left;
-    width: 10px;
-    height: 40px; }
+/* override these to theme component */
+.timeseriescontrolplugin {
+  background-color: rgba(255, 255, 255, 1);
+}
+g.full-axis .tick line {
```
ZakarFin: These g.* should probably have .timeseriescontrolplugin in front of them to not affect any g.*, but just the ones under the timeseries control?

bundles/framework/timeseries/resources/css/timeseriesplayback.css
```diff
@@ -1,135 +1,197 @@
...
-        border-left: 1px solid #fff;
-        height: 8px;
-        position: absolute;
-        bottom: 0; }
-  div.mapplugin.timeseries .oskari-timeslider-empty {
-    margin-top: 10px;
-    float: left;
-    width: 10px;
-    height: 40px; }
+/* override these to theme component */
```
ZakarFin: Nice thinking with the theming although the scss needs to be removed because of this.

bundles/framework/timeseries/service/TimeseriesLayerService.js
```diff
@@ -0,0 +1,146 @@
...
+         */
+        _checkMultipleLayers: function () {
+            if (this._timeseriesService.getCountByType('layer') > 1) {
+                this._popupService.closeAllPopups(false);
+                var popup = this._popupService.createPopup();
+                var closeBtn = Oskari.clazz.create('Oskari.userinterface.component.Button');
+                closeBtn.setTitle(this.loc('alert.ok'));
+                closeBtn.setHandler(function () {
+                    popup.close(true);
+                });
```
ZakarFin: You can create a close button for a popup with just:

       var closeBtn = popup.createCloseButton(this.loc('alert.ok'));

The label parameter is optional defaulting to a "Close" label.

bundles/framework/timeseries/service/TimeseriesLayerService.js
```diff
@@ -0,0 +1,146 @@
...
+            }
+        },
+        /**
+         * @method _checkMultipleLayers
+         * @private
+         * Show popup if there are more than one selected timeseries layers
+         */
+        _checkMultipleLayers: function () {
+            if (this._timeseriesService.getCountByType('layer') > 1) {
+                this._popupService.closeAllPopups(false);
```
ZakarFin: This will close any other popup that is currently open. Doesn't seem very userfriendly, but I guess it's ok here.

bundles/framework/timeseries/service/TimeseriesLayerService.js
```diff
@@ -0,0 +1,146 @@
...
+        },
+        /**
+         * @method updateTimeseriesLayers
+         * Update timeseries service state based on current selected layers
+         */
+        updateTimeseriesLayers: function () {
+            var me = this,
+                layers = me._sandbox.findAllSelectedMapLayers();
+            for (var i = 0; i < layers.length; i++) {
+                if (layers[i].hasTimeseries()) {
```
ZakarFin: I prefer using continue when all the code inside a loop is in a single if statement to prevent having large blocks and/or }}} kinds of rows, but it's mostly an matter of opinion.

    if (!layers[i].hasTimeseries()) {
         continue;
    }

bundles/framework/timeseries/service/TimeseriesLayerService.js
```diff
@@ -0,0 +1,146 @@
...
+                    var series = this._timeseriesService.getTimeseries(layer.getId(), 'layer');
+                    if (!series) {
+                        var delegate = me._layerDelegateFactory(layer.getId(), layer.getLayerType());
+                        if(!delegate) {
+                            continue;
+                        }
+                        this._timeseriesService.registerTimeseries(layer.getId(), 'layer', -i, delegate);
+                    } else {
+                        this._timeseriesService.updateTimeseriesPriority(layer.getId(), 'layer', -i);
+                    }
```
ZakarFin: Same here, I would do:

    if (series) {
        this._timeseriesService.updateTimeseriesPriority(layer.getId(), 'layer', -i);
        continue;
    }
    var delegate = me._layerDelegateFactory(layer.getId(), layer.getLayerType());
    if(delegate) {
        this._timeseriesService.registerTimeseries(layer.getId(), 'layer', -i, delegate);
    }

bundles/framework/timeseries/service/TimeseriesService.js
```diff
@@ -0,0 +1,177 @@
...
+         * @method getActiveTimeseries
+         * Get timeseries that should have UI visible (has lowest priority)
+         * @return {Object/undefined} timeseries config or undefined
+         */
+        getActiveTimeseries: function () {
+            var me = this;
+            var things = Object.keys(this._timeseriesThings).map(function (key) { return me._timeseriesThings[key] });
+            var targets = things.filter(function (thing) {
+                return thing.type !== 'layer';
+            });
```
ZakarFin: Looks like map() and filter() are supported by the IE 9+ and modern browsers. We need to start using these in other places as well 👍

bundles/framework/timeseries/view/TimeseriesControlPlugin.js
```diff
@@ -0,0 +1,642 @@
...
+                        }
+                    }
+                }
+            },
+            buttonGroup: 'mobile-toolbar'
+        };
+    }, {
+        __fullAxisYPos: 35,
+        __localeData: {
+            fi: {
```
ZakarFin: These should be in locale files.

bundles/framework/timeseries/view/TimeseriesControlPlugin.js
```diff
@@ -0,0 +1,642 @@
...
+                "shortMonths": ["Jan", "Feb", "Mar", "Apr", "Maj", "Jun", "Jul", "Aug", "Sep", "Okt", "Nov", "Dec"]
+            }
+        },
+
+        // Returns a function, that, when invoked, will only be triggered at most once
+        // during a given window of time. Normally, the throttled function will run
+        // as much as it can, without ever going more than once per `wait` duration;
+        // but if you'd like to disable the execution on the leading edge, pass
+        // `{leading: false}`. To disable execution on the trailing edge, ditto.
+        _throttle: function (func, wait, options) {
```
ZakarFin: Feels like throttle should be a function in the Oskari(.util?)-global instead of implemented in a bundle.

bundles/framework/timeseries/WMSAnimator.js
data-ux: I would add the configuration in this case when/if it's needed in the future. _bufferImages() is a private method and in the way it's currently used, it doesn't really make a difference if the buffering failed or succeeded. There just needs to be some reasonable timeout so that the animation can progress in any case.

bundles/framework/timeseries/WMSAnimator.js
data-ux: Good catch. Actually I would solve the problem the other way around: move `this._isBuffering = true;` from `_bufferImages` to `requestNewTime`. In this way _bufferImages would not mutate the WMSAnimator instance state, and therefore it could be called from many call sites independently if needed.

bundles/framework/timeseries/view/TimeseriesControlPlugin.js
```diff
@@ -0,0 +1,642 @@
...
+                formatMonth = formatterFunction("%b"),
+                formatYear = formatterFunction("%Y");
+
+            return function multiFormat(date) {
+                return (d3.timeSecond(date) < date ? formatMillisecond
+                    : d3.timeMinute(date) < date ? formatSecond
+                    : d3.timeHour(date) < date ? formatMinute
+                    : d3.timeDay(date) < date ? formatHour
+                    : d3.timeMonth(date) < date ? formatDay
+                    : d3.timeYear(date) < date ? formatMonth
```
ZakarFin: We are using both moment.js and d3 for date/time operations?

ZakarFin: I guess it's ok to use d3 here as well since there's other d3 stuff in this particular file.

bundles/framework/timeseries/WMSAnimator.js
data-ux: Good question. Do we generally want to check that arguments are of correct type, or do we trust that the developer has read the docs?

bundles/framework/timeseries/instance.js
data-ux: Yes, the thinking was that since 'Oskari.mapframework.bundle.timeseries.WMSAnimator' class is part of timeseries bundle, we should register it for use here. Another option would be to move WMSAnimator to be a part of the general WMS layer functionality, but then the code would be loaded even if it cannot be used (requires timeseries bundle). I think it's better this way.

bundles/framework/timeseries/WMSAnimator.js
ZakarFin: Ok, lets keep it as is

bundles/framework/timeseries/WMSAnimator.js
ZakarFin: ”It’s complicated” :) With simple values usually no checking is required, but with complicated values/object I’d prefer some validation. I guess this falls somewhere in the middle. I think the question can be answered by asking ”how difficult is it to find the root cause if someone passes an invalid value causing an error later on in the code”. That being said, its not mandatory to have a check here.

bundles/framework/timeseries/instance.js
ZakarFin: True, lets keep it like this.

bundles/framework/timeseries/view/TimeseriesControlPlugin.js
data-ux: moment.js is used for time duration/interval etc. handling, d3 for time format presentation.

bundles/framework/timeseries/service/TimeseriesLayerService.js
```diff
@@ -0,0 +1,147 @@
...
+        /**
+         * @method _checkMultipleLayers
+         * @private
+         * Show popup if there are more than one selected timeseries layers
+         */
+        _checkMultipleLayers: function () {
+            if (this._timeseriesService.getCountByType('layer') > 1) {
+                var popup = this._popupService.createPopup();
+                var closeBtn = popup.createCloseButton(this.loc('alert.ok'));
+                closeBtn.setHandler(function () {
```
ZakarFin: popup.createCloseButton() creates the handler also https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/divmanazer/component/Popup.js#L180-L190

ZakarFin:
Pretty much a rewrite of everything concerning timeseries in Oskari.

ZakarFin:
Nice work!

PR #154, opened 2017-10-24
New RPC tests

Miselius:
Added new rpc test specs: 
-location_spec	
-features_spec
-drawing_spec
-markers_spec


tests/rpc/spec/drawing_spec.js
```diff
@@ -0,0 +1,93 @@
...
+               'LineString',
+               'Polygon',
+               'Circle',
+               'Box',
+               'Square'
+               ],
+                function(drawRequests, done) {
+                    channel.postRequest('DrawTools.StartDrawingRequest', ['my functionality id', drawRequests]);
+                    channel.log("StartDrawingRequest:", drawRequests)
+                    eventCounter++;
```
ZakarFin: I think this variable should be scoped from global to test scope (introduced / test file). In here it's not counting events so the naming is very misleading.

tests/rpc/spec/drawing_spec.js
```diff
@@ -0,0 +1,93 @@
+describe('Drawing', function(){
+
+    function handleEvent(name, handler) {
```
ZakarFin: This function should be in some common helpers file.

tests/rpc/spec/drawing_spec.js
```diff
@@ -0,0 +1,93 @@
...
+                    eventCounter++;
+                    done();
+                }
+            );
+        });
+
+        describe('Stop Drawing', function() {
+
+            it("Cancels Drawing request", function(done) {
+                handleEvent('DrawingEvent', function(data) {
```
ZakarFin: Maybe do a function at the start of this file checks the event and call it in here and in the next it() as the code is the same in both.

tests/rpc/spec/features_spec.js
```diff
@@ -0,0 +1,109 @@
...
+                });
+             });
+        });
+    });
+
+    afterEach(function() {
+        // Spy callback.
+        expect(eventCounter).toEqual(1, "Event count does not match");
+        // Reset event handlers.
+        while (handlersToClean.length) {
```
ZakarFin: This seems to be repeated in each spec. Maybe move the code to a common teardown function that can be called from each spec.

tests/rpc/spec/features_spec.js
```diff
@@ -0,0 +1,109 @@
...
+            eventCounter++;
+            done();
+        });
+        // Remove all features which 'test_property' === 'line feature' from the layer id==='VECTOR'
+        channel.postRequest('MapModulePlugin.RemoveFeaturesFromMapRequest', ['test_property', lineGeojsonObject.features[0].properties.test_property, addLineFeatureParams[1].layerId]);
+        channel.log('RemoveFeaturesFromMapRequest test_property:', lineGeojsonObject.features[0].properties.test_property, addLineFeatureParams[1].layerId);
+    });
+
+    it("Removes all features", function(done) {
+        //annihilate everything - does not trigger featureEvent
```
ZakarFin: This seems like a bug in the implementation. In my opinion removing everything should trigger an event.

tests/rpc/src/common_parameters.js
```diff
@@ -10,35 +10,227 @@ var handlersToClean = [];
...
+var testLayerOptions = {
+  'minResolution': 0,
+  'maxResolution': 10000
+};
+
+// # Requests: Change map layer style
+var styleParams = {
+  SLD_BODY: '<StyledLayerDescriptor xmlns=\"http://www.opengis.net/sld\" xmlns:ogc=\"http://www.opengis.net/ogc\" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0.0" xsi:schemaLocation="http://www.opengis.net/sld StyledLayerDescriptor.xsd">' +
+    '    <NamedLayer>' +
+    '    <Name>oskari:kunnat2013</Name>' +
```
ZakarFin: I guess the name doesn't really have any meaning here?

tests/rpc/spec/location_spec.js
```diff
@@ -0,0 +1,104 @@
...
+       "TRANSIT,WALK"
+       ],
+        function(routeRequest, done) {
+
+            var myRouteRequest = {
+              "fromlat": defaultPosition.centerY,
+              "fromlon": defaultPosition.centerX,
+              "srs": defaultPosition.srsName,
+              "tolat": defaultPosition.centerY+1000,
+              "tolon": defaultPosition.centerX+1000,
```
ZakarFin: These probably won't work in a non-metric projection.

tests/rpc/src/common_parameters.js
ZakarFin: Doesn't seem like a reasonable default so maybe use some other name or document if tests assume a layer with this name.

tests/rpc/spec/drawing_spec.js
Miselius: Agreed, it is used to spy rpc callback in general. It was originally used to only spy event callbacks. Name is misleading.

tests/rpc/spec/drawing_spec.js
Miselius: Agreed, eventhandler is identical and repeated.

tests/rpc/spec/location_spec.js
Miselius: There might be a problem with routing rpc in demo.oskari.org maps, since this didn't work with +10 either. Regardless, it's true that if the test value would be +10 it would work in both meters and degrees.
"The valid range of latitude in degrees is -90 and +90 for the southern and northern hemisphere respectively. Longitude is in the range -180 and +180 specifying coordinates west and east of the Prime Meridian, respectively."

tests/rpc/src/common_parameters.js
Miselius: Will check.

tests/rpc/spec/features_spec.js
Miselius: Bug reported.

tests/rpc/spec/features_spec.js
Miselius: Agreed, eventHandler() and resetEventhandler() needs to be defined only once.

tests/rpc/spec/drawing_spec.js
Miselius: Will move.

ZakarFin:
Nice, thanks!

PR #162, opened 2017-11-01
Fix feature event not sending event when removing all features in ol3-implementation

nxholm:


ZakarFin:
FeatureEvent is now triggered correctly on published maps/ol3 implementation when layer is cleared with MapModulePlugin.RemoveFeaturesFromMapRequest without parameters.

PR #174, opened 2017-11-07
Added move map to extent functionality to MapMoveByLayerContentRequest

okauppinen:
Added zoomToExtent (optional, default false) parameter to MapMoveByLayerContentRequest. Handler moves and zooms the map to layer extent if zoomToExtent is true and layer has geometry. 

Moved openlayers dependent getCentroid to layerplugin ol2/3 getGeometryCenter. Also added getGeometryBounds method which is used by zoomToExtnent functionality.

Fixed LayersPlugin.ol3 _parseGeometryForLayer to set geometry to array of geometries.


ZakarFin:
Looks good. The request seems to be missing it's API documentation. Can you add a simple md-file describing the request in here https://github.com/oskariorg/oskari-frontend/tree/master/api/mapping/mapmodule/request

PR #181, opened 2017-11-21
New bundle: geometrycutter

data-ux:
Geometrycutter is a more bare bones replacement for the "geometryeditor" bundle. Geometrycutter has only two editing modes:
- splitting a feature by a user drawn linestring as if it was a cutting "blade"
- removing a part of a feature that overlaps a user drawn polygon (difference)

Benefits compared to old "geometryeditor" bundle:

- Map library agnostic: uses DrawRequest/AddFeaturesToMapRequest
- Only one dependency: JSTS
- More maintainable structure & size: 500 loc vs. 2500 loc


bundles/framework/geometrycutter/BundleModule.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * @class Oskari.mapframework.ui.module.common.geometrycutter.BundleModule
+ */
+Oskari.clazz.define('Oskari.mapframework.bundle.geometrycutter.BundleModule', function () {
+    this.sandbox = null;
+}, {
+    getName: function () {
+        return this.__name
```
ZakarFin: I can't find definition for __name here.  This functions as a "basemodule" to reduce boilerplate-code in actual instance.js?

bundles/framework/geometrycutter/GeometryProcessor.js
```diff
@@ -0,0 +1,89 @@
...
+                    parts.push(splitted.getGeometryN(i));
+                }
+                break;
+            default:
+                throw new Error('Unsupported split geometry result!');
+        }
+        return parts.map(function (part, i) {
+            return {
+                type: 'Feature',
+                geometry: this._writer.write(part),
```
ZakarFin: Is the scope of "this" correct here?

bundles/framework/geometrycutter/instance.js
```diff
@@ -0,0 +1,109 @@
...
+     * @method startEditing
+     * @param {String} operationId unique id for geometry editing operation
+     * @param {org.geojson.Feature} feature the target feature to be edited
+     * @param {String} mode cutting mode: "lineSplit" or "polygonClip"
+     */
+    startEditing: function (operationId, feature, mode) {
+        var drawId = this.__idPrefix + operationId;
+        this.stopEditing(operationId, false); // cleanup any previous edits with same operationId
+        function requestFunction(requestName, args) {
+            var builder = this.sandbox.getRequestBuilder(requestName);
```
ZakarFin: Use Oskari.requestBuilder(requestName) instead. It will return the builder function if the request is available in this appsetup. If you want to know if the request is being handled in given sandbox you can do sandbox.hasHandler(requestName)

bundles/framework/geometrycutter/request/StartGeometryCuttingRequestHandler.js
```diff
@@ -0,0 +1,10 @@
+Oskari.clazz.define('Oskari.mapframework.bundle.geometrycutter.StartGeometryCuttingRequestHandler', function (sandbox, geometryCutterInstance) {
```
ZakarFin: Sandbox isn't really needed here.

bundles/framework/geometrycutter/request/StopGeometryCuttingRequestHandler.js
```diff
@@ -0,0 +1,10 @@
+Oskari.clazz.define('Oskari.mapframework.bundle.geometrycutter.StopGeometryCuttingRequestHandler', function (sandbox, geometryCutterInstance) {
```
ZakarFin: You can remove the sandbox reference here as well.

bundles/framework/geometrycutter/instance.js
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+    },
+    requestHandlers: {
+        'StartGeometryCuttingRequest': function () {
+            return Oskari.clazz.create('Oskari.mapframework.bundle.geometrycutter.StartGeometryCuttingRequestHandler', this.sandbox, this);
+        },
+        'StopGeometryCuttingRequest': function () {
+            return Oskari.clazz.create('Oskari.mapframework.bundle.geometrycutter.StopGeometryCuttingRequestHandler', this.sandbox, this);
+        }
```
ZakarFin: As the requesthandlers are so simple you could also just add a handleRequest() function in the instance.js and return "this" as the handler for the requests. You would need to check the request.getName() in handleRequest() to check what to do. Just a thought - the current implementation is ok too.

ZakarFin:
Nice work. There's one reference to "this" that might be a bug, but in general looks good for merging.

bundles/framework/geometrycutter/GeometryProcessor.js
data-ux: Context is given as second parameter of `.map()`.

PR #183, opened 2017-11-23
Featuredata2 improved feature map selection and finalized drawplugin ol4 support

kuosman:
# DrawPlugin.ol3
Now users can also draw circle.
Circle now gets area measurement.

# WfsLayerPlugin.ol3
Fixed assertion error when trying to add highlighted layer to map when highlighted layer already exists.

# Featuredata2
Users can now really select map features on the map and featuredata grid shows them selected. Previously maps selection not update featuredata grid correctly.


bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -405,7 +412,13 @@ Oskari.clazz.define(
          */
         getFeatures: function (layerId) {
             var me = this,
-                features = [];
+                features = [],
+                layer = me.getLayer(layerId);
+
+            if(!layer) {
+                return features;
```
data-ux: If no layer with `layerId` is found, it would be better to return `null` to distinguish from the case where the layer was found but it had zero features. Other code calling this method would need to be updated to handle the case, of course.

PR #185, opened 2017-11-27
Empty WFS layer selections only if there are those.

hvisuri:
When map is clicked and there is WFS layer on map, WFS selections were emptied even if there were no selections. This caused some problems because of the event that was sent.

Now WFS selections are emptied only if there are those.

### DIFF
--- a/bundles/mapping/mapwfs2/service/Mediator.js
+++ b/bundles/mapping/mapwfs2/service/Mediator.js
@@ -395,7 +395,11 @@ Oskari.clazz.category('Oskari.mapframework.bundle.mapwfs2.service.Mediator', 'ge
         } else {
             // FIXME: pass coordinates from server in response, but not like this
             data.data.lonlat = this.lonlat;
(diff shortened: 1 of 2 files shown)


bundles/mapping/mapwfs2/service/Mediator.js
```diff
@@ -395,7 +395,11 @@ Oskari.clazz.category('Oskari.mapframework.bundle.mapwfs2.service.Mediator', 'ge
         } else {
             // FIXME: pass coordinates from server in response, but not like this
             data.data.lonlat = this.lonlat;
-            me.WFSLayerService.emptyWFSFeatureSelections(layer);
+
+            if (me.WFSLayerService.getWFSSelections().length > 0) {
+                me.WFSLayerService.emptyWFSFeatureSelections(layer);
```
ZakarFin: This if should be inside the emptyWFSFeatureSelections() implementation. Otherwise looks good.

PR #188, opened 2017-11-30
Printout2

nxholm:
- new version of print
- you can add tools to the print area by implementing a protocol (Oskari.mapping.printout2.Tool) in the target tool


ZakarFin:
Current problems:
- Preview:
    - Preview image on sidepanel is not updating on map move. 
    - The panel for the preview image starts up closed.
    - Options are not reset when the user leaves and returns the functionality, but the darkened preview area on the map assumes default paper size in this case.
    - Do we really want to have 2 previews? Maybe we should atleast tell the difference to the user (image is from the actual print server <> preview area on map is a "best guess" what kind of area fits the paper)
- Tools protocol isn't ready:
    - Tools should not return an element that is included on the map. Instead they should receive a reference to the preview element and attach any possible preview elements to it directly.
    - Tools should be able to "turn off" and  remove the ui from the preview area.
    - Tools could have a setEnabled(boolean, previewAreaRef) function that should be called from the printout with true/false based on the checkbox tick and tools should attach/remove the preview part they provide.
    - Tools need to have a way to return a value that is passed to the server. Otherwise any selection the user does is not reflected on the end product.
- The Oskari classes defined in printout2 are lowercase where in other parts of Oskari they are CamelCased.
- WFS layers are not included in printing

ZakarFin:
The thing that we saved this for was the preview overlay on the map for printable area that could be used instead of the smaller preview image generated on server. We decided the server generated preview is better because it can't go out of sync in any case. So closing this as rest of the code is so much behind the current code base that it's easier to start from scratch than continuing from this.

PR #198, opened 2017-12-07
Fix initial state for publisher

ZakarFin:
Fix for PR #194 since the publisher now ALWAYS uses StateRequest to setup the app for preview. The "default data" only had configuration to setup the initial publisher tools and this change constructs the missing parts of initial publisher data using current app state.

### DIFF
--- a/bundles/framework/publisher2/instance.js
+++ b/bundles/framework/publisher2/instance.js
@@ -148,8 +148,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
 
             if (blnEnabled) {
                 var stateRB = Oskari.requestBuilder('StateHandler.SetStateRequest');
-                var req = stateRB(data.configuration);
-                this.getSandbox().request(this, req);
+                this.getSandbox().request(this, stateRB(data.configuration));
 
                 me.getService().setNonPublisherLayers(deniedLayers);
                 me.getService().removeLayers();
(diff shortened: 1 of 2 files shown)


data-ux:
I would move the new method `getDefaultData` into instance.js and call it from within `setPublishMode` if no data is given. In this way `setPublishMode` would be a self contained entry point to the publisher functionality. Idea being it could be called with or without data and it would do the proper thing in both cases. What do you think?

ZakarFin:
Good point! Initial data for publisher is now handled in instance.js instead of the request handler.

bundles/framework/publisher2/instance.js
```diff
@@ -94,14 +93,17 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
...
             var loc = this.getLocalization();
 
             this.__service = Oskari.clazz.create('Oskari.mapframework.bundle.publisher2.PublisherService', sandbox);
             // create and register request handler
             var reqHandler = Oskari.clazz.create(
-                    'Oskari.mapframework.bundle.publisher.request.PublishMapEditorRequestHandler',
-                    this);
+                'Oskari.mapframework.bundle.publisher.request.PublishMapEditorRequestHandler',
+                function(data) {
+                    me.setPublishMode(true, data);
```
data-ux: We could just pass bound `setPublishMode` to the PublishMapEditorRequestHandler constructor:
`this.setPublishMode.bind(this, true)`
but maybe your version is more readable.

data-ux:
Looks good! Good idea about making the deniedLayers an optional parameter.

PR #210, opened 2017-12-12
Statsgrid extra features changes

kuosman:
## Search flyout extra features
- added ``Open table`` checkbox, defaults unchecked
- added ``Open diagram`` checkbox, defaults unchecked
- changed ``Show map layers`` --> ``Hide other map  layers``

## Other fixes
- also fixed tile icons and published map toggle icons


bundles/statistics/statsgrid2016/components/ExtraFeatures.js
```diff
@@ -1,26 +1,94 @@
-Oskari.clazz.define('Oskari.statistics.statsgrid.ExtraFeatures', function(sandbox, locale) {
+Oskari.clazz.define('Oskari.statistics.statsgrid.ExtraFeatures', function(sandbox, locale, instance) {
```
ZakarFin: I can't find a reference to instance on the code that was added. Why is it passed here?

bundles/statistics/statsgrid2016/Tile.js
```diff
@@ -129,6 +129,23 @@ function(instance, service) {
             el.addClass('material-selected');
         }
     },
+    /**
+     * @method  @public openExtension opens extension
+     * @param  {String} type  flyout type
+     */
+    openExtension: function(type) {
```
ZakarFin: If you want to add an "open" function it would be nice to have a "close" function and refactor all code using toggle to use these instead.

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -139,8 +140,18 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function(
                     }
                     me.service.getStateService().setRegionset(regionSelect.value());
 
-                    // FIXME need this anymore ?
-                    //me.instance.getFlyout().closePanels();
+                    var flyoutManager = (me.instance && me.instance.getTile() && me.instance.getTile().getFlyoutManager()) ? me.instance.getTile().getFlyoutManager() : null;
+                    var searchFlyout = (flyoutManager && flyoutManager.getFlyout('search')) ? flyoutManager.getFlyout('search') : null;
```
ZakarFin: You can take out the last ifs to make it more readable as the check for searchFlyout will basically be the same if the value is undefined or null. I would consider refactoring this to make it work without going through so many references. Like using some internal event through the service to trigger a flyout to open or some other way.

ZakarFin: Is this really something that should be the responsibility of indicator parameters in the first place? Maybe you should just call a method in extrafeatures that an indicator was added. It should decice what to do with the information, not this component.

PR #211, opened 2017-12-13
Don't send mapClickedEvent while drawing.

hvisuri:
Now it is possible to tell mapmodule that drawing is on or off with setDrawingMode(). Drawing mode can be asked from mapmodule with getDrawingMode(). DrawPlugin uses this to tell mapmodule about drawing state and mapmodule then uses this information when map is clicked to know whether mapClickedEvent should be sent or not. While drawing, the event is not sent.

All this was made to make GFI popup handling easier and to avoid unnecessary event handling.

### DIFF
--- a/bundles/mapping/mapmodule/AbstractMapModule.js
+++ b/bundles/mapping/mapmodule/AbstractMapModule.js
@@ -105,6 +105,8 @@ Oskari.clazz.define(
         this._cursorStyle = '';
         this.log = Oskari.log('AbstractMapModule');
 
+        this.isDrawing = false;
+
 
         this.templates = {
             'crosshair': jQuery(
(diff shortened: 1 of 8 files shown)


bundles/mapping/mapmodule/AbstractMapModule.js
```diff
@@ -477,6 +480,18 @@ Oskari.clazz.define(
         getProjection: function () {
             return this._projectionCode;
         },
+
+        setDrawingMode: function (mode) {
+            if (mode) {
```
ZakarFin: You can do the same with just:

    this.isDrawing = !!mode;

bundles/mapping/mapmodule/mapmodule.ol3.js
```diff
@@ -104,6 +104,9 @@ Oskari.clazz.define('Oskari.mapframework.ui.module.common.MapModule',
             });
 
             map.on('singleclick', function (evt) {
+                if (me.getDrawingMode()) {
+                    return;
```
ZakarFin: The same handling should be added to ol2 mapmodule to make the functionality consistent.

PR #216, opened 2017-12-18
publisher2/custom ui

nxholm:
- add function to check if a element reference has been passed in config, if true, disable tile and use custom element as starting point

### DIFF
--- a/bundles/framework/publisher2/instance.js
+++ b/bundles/framework/publisher2/instance.js
@@ -165,7 +165,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
             // trigger an event letting other bundles know we require the whole UI
             var eventBuilder = Oskari.eventBuilder('UIChangeEvent');
             this.sandbox.notifyAll(eventBuilder(this.mediator.bundleId));
-            if ( !!this.configurationHasCustomElement ) {
+            if ( !!this.configurationHasCustomElement() ) {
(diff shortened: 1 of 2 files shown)


ZakarFin:
Should custom element receive some class so it can be styled based on publisher being open/closed? You could do this in setPublishMode() based on the boolean param it receives.

bundles/framework/publisher2/instance.js
```diff
@@ -130,7 +134,20 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
         getService : function() {
             return this.__service;
         },
-
+        /**
+         * @return {STRING} reference to element-id to use instead of tile as bundle ui-element
```
ZakarFin: For consistency change {STRING} -> {String} like it's in other places

ZakarFin:
Also update https://github.com/oskariorg/oskari-frontend/blob/master/api/framework/publisher2/bundle.md

bundles/framework/publisher2/instance.js
```diff
@@ -148,7 +165,9 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
             // trigger an event letting other bundles know we require the whole UI
             var eventBuilder = Oskari.eventBuilder('UIChangeEvent');
             this.sandbox.notifyAll(eventBuilder(this.mediator.bundleId));
-
+            if ( !!this.configurationHasCustomElement ) {
```
ZakarFin: This is always true

PR #221, opened 2017-12-20
Admin layers import/export

kuosman:
### Admin layerselector

Added ``Export`` link to layers. For clicking Export-button new window opened where is showed current layer JSON.

### admin/admin

Added ``Layers Import`` new tab for admins where layers can be imported (JSON-format).

>This is not ready yet, some minor changes are needed when server side code is ready to use.

Needs least following fixes:
- actionroute url checks
- layer import success message showing


ZakarFin:
Too many changes has passed since this. It's easier to do this from scratch than merging stuff

PR #223, opened 2017-12-22
Statsgrid legend small screen fix

kuosman:
Now statsgrid legend showed correctly also small screens.

### DIFF
--- a/bundles/statistics/statsgrid2016/components/Legend.js
+++ b/bundles/statistics/statsgrid2016/components/Legend.js
@@ -46,7 +46,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function(sandbox, loca
     //   Legend
     //
     // Alternatively note about no indicator selected
-    render : function(el) {
+    render : function(el, handler) {
         if(this._renderState.inProgress) {
             // handle render being called multiple times in quick succession
             // previous render needs to end before repaint since we are doing async stuff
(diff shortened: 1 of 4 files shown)


bundles/statistics/statsgrid2016/components/Legend.js
```diff
@@ -69,17 +69,24 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function(sandbox, loca
             // attach container to parent if provided, otherwise updates UI in the current parent
             el.append(container);
         }
+
+        var renderDone = function(){
+            me._renderDone();
+            if(typeof handler === 'function') {
```
ZakarFin: I'd say you should send the optional handler to me._renderDone(handler) OR attach it to renderState. Having 2 similar function names isn't too much fun when trying to figure out what's going on. Should handler be called even with immediate repaint? If you attach the handler to renderState you can call it after rendering (including repaint) is totally finished. The again having another call to render with handler would overwrite the current... So maybe the best option would be to make the Legend component observable and trigger 'change' etc events after render is completed:

    _renderDone : function() {
        var state = this._renderState;
        this._renderState = {};
        this._restorePanelState(this._accordion, state.panels);
        if(state.repaint) {
            this.render(state.el);
        } else {
            this.trigger('change');
       }
    },

bundles/statistics/statsgrid2016/plugin/ClassificationToolPlugin.js
```diff
@@ -108,7 +110,9 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.plugin.ClassificationToolPlugin
                 return me._element;
             }
             me._element = me._templates.main.clone();
-            this.__legend.render(me._element);
+            this.__legend.render(me._element, function(){
+                me._calculatePluginSize();
+            });
```
ZakarFin: With the Legend as observable change this would be:

     this.__legend.on('change', function(){
           me._calculatePluginSize();
     });

bundles/statistics/statsgrid2016/plugin/ClassificationToolPlugin.js
```diff
@@ -166,6 +170,35 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.plugin.ClassificationToolPlugin
...
         stopPlugin: function(){
             this.teardownUI(true);
+        },
+        _createEventHandlers: function () {
+            return {
+                /**
+                 * @method AfterMapMoveEvent
+                 * Shows map center coordinates after map move
+                 */
+                AfterMapMoveEvent: function () {
```
ZakarFin: Why does moving the map trigger size check? Shouldn't this be something like MapSizeChangedEvent?

ZakarFin:
Thanks! I can actually make the changes that I recommended.

PR #225, opened 2018-01-05
display number of layers in a category with a badge instead of text

nxholm:
# Changes
- accordion header is now a flexbox so we can simply align the items in the header.


bundles/framework/divmanazer/scss/badge.scss
```diff
@@ -21,16 +21,26 @@
...
+    transform: translate(0,-50%);
     @include border-radius(9px);
 }
-
+.badge-wrapper {
+    display: inline-block;
+    position: absolute;
+    right: 8%;
+    height: inherit;
+    max-width: 50px;
```
ZakarFin: I'd say "right: 8%" or "max-width" defs should NOT be part of the badge ui-component. It should be defined in for example layerselector bundles css and specific to the layerlisting UI.

ZakarFin:
Thanks! This PR changes how the badge component renders AND changes the way layerselector bundle shows the number of layers under a group so it could have been split to 2 PRs. 

When layers are being filtered the layerselector shows the layercount as before (after the group title) AND with the badge: 
badge

Could the badge be updated while filtering and the title could remain as is?

bundles/framework/divmanazer/component/Badge.js
```diff
@@ -42,10 +44,15 @@ Oskari.clazz
                 }
 
                 var txtspan = this.compiledTemplates[status || 'default'].clone();
+                var wrapper = this.compiledTemplates['wrapper'].clone();
                 txtspan.append(pContent);
```
ZakarFin: Use txtspan.html(pContent) instead of append() to prevent possible XSS issues.

bundles/framework/layerselector2/view/LayersTab.js
```diff
@@ -356,8 +360,11 @@ Oskari.clazz.define(
                 if (group.layerListPanel.isVisible()) {
                     visibleGroupCount += 1;
                 }
-                group.layerListPanel.setTitle(group.getTitle() + ' (' +
-                    visibleLayerCount + '/' + layers.length + ')');
+                group.layerListPanel.setTitle(group.getTitle());
```
ZakarFin: Title doesn't need to be updated anymore?

bundles/framework/layerselector2/view/LayersTab.js
```diff
@@ -655,9 +662,8 @@ Oskari.clazz.define(
                 }
                 group.layerListPanel.setVisible(true);
                 group.layerListPanel.close();
-                group.layerListPanel.setTitle(
-                    group.getTitle() + ' (' + layers.length + ')'
-                );
+                group.layerListPanel.setTitle( group.getTitle() );
```
ZakarFin: Title doesn't need to be updated in here either?

bundles/framework/layerselector2/view/LayersTab.js
nxholm: It only needs to know the group title, the filtered layers / layers is shown in the badge.

ZakarFin: Sure, I meant that you can probably drop the whole setTitle() call from here.

PR #227, opened 2018-01-12
DrawTools now enable multigeometries

hvisuri:
It is now possible to give value 'multiGeom' to parameter allowMultipleDrawing in StartDrawingRequest. By giving that value DrawPlugin gathers drawn shapes into a single feature with multigeometry instead of sending multiple features in DrawingEvent.

Bug related to parameter allowMultipleDrawing = false is also fixed.


ZakarFin:
This introduces a problem where after drawing multigeometries it's no longer possible to NOT draw them.

bundles/mapping/drawtools/plugin/DrawPlugin.ol3.js
```diff
@@ -584,6 +665,11 @@ Oskari.clazz.define(
             var optionsForDrawingEvent = {
                 isFinished: false
             };
+
+            if (options.allowMultipleDrawing === 'multiGeom') {
+                me.drawMultiGeom = true;
```
ZakarFin: Nothing resets me.drawMultiGeom so after it has been set true it will remain so until the page is reloaded.

PR #230, opened 2018-01-15
Terrain profile

data-ux:
The bundle provides UI for querying and showing terrain profiles from Paikkatietoikkuna action route "TerrainProfile". A new tool is loaded into the toolbar. Clicking on the tools starts map drawing. After the user has drawn a linestring on the map, the terraing profile is queried from the action route.


ZakarFin:
The code looks good for merging. There's some functional bug that still needs work so not merging yet.

PR #231, opened 2018-01-16
Set parameter 'drawMultiGeom' to false by default

hvisuri:
Parameter 'drawMultiGeom' is now set to false by default when drawing is started. It is set to true if it is defined so in request options.


ZakarFin:
Fixes an issue introduced in #227 where once multigeometry is requested it could not be turned off.

PR #241, opened 2018-01-23
Download basket changes

kuosman:
* Fixed bundle name in bundle documentation
* Removed duplicate download-basket bundle in hsy folder


ZakarFin:
Thanks!

PR #248, opened 2018-01-25
publisher2 reset mapdiv width on exit

nxholm:
- mapdiv has by default width 100%, publisher2 modifies this value into pixels and wasn't reset after exiting the publisher


ZakarFin:
This fixes the root cause db24a571f863e3875cef93b62c311390e9bf0715 so closing this without merging.

PR #250, opened 2018-01-26
Featuredata: Capture Control-click in Grid.js click handler instead of sandbox

data-ux:


ZakarFin:
Selecting additional features in featuredata table works with this change.

PR #255, opened 2018-02-01
Layer parsing improvement

ZakarFin:
When Oskari instance has lots of registered layers available the browser "hangs" for some time while parsing them. This change improves responsiveness during layer parsing by yielding cpu time between layers instead of doing it on a tight loop.


ZakarFin:
Good point! This is for parsing the layers that are not on the initial map so yielding time to render the rest of the UI and map staying responsive should be priority in my opinion. But I guess yielding after every layer might be an overkill and introduce unnecessary overhead.

PR #256, opened 2018-02-02
Statsgrid / publish diagram

nxholm:
- Now there is an option to publish a diagram in the publisher if a thematic maplayer is present when publishing a map
- Added a diagram element to the toggleplugin, which controls the diagram flyout in embedded map.


bundles/statistics/statsgrid2016/instance.js
```diff
@@ -318,23 +324,61 @@ Oskari.clazz.define(
...
-            var sandbox = me.getSandbox();
+        getEmbeddedTools: function () {
+            return this.embeddedTools;
+        },
+        /**
+         * Check if the tool has been added to the array containing tool names opened in publisher
+         * @method publisherHasTool
+         * @param {String} tool tool name as string
+         */
+        publisherHasTool: function (tool) {
```
ZakarFin: This isn't really related to publisher. Could just as well be named hasTool()

bundles/statistics/statsgrid2016/instance.js
```diff
@@ -318,23 +324,61 @@ Oskari.clazz.define(
...
+         */
+        showEmbeddedTools: function (tool, enabled) {
+            var sandbox = this.getSandbox();
             var mapModule = sandbox.findRegisteredModuleInstance('MainMapModule');
-            if(!enabled) {
-                if(this.togglePlugin) {
-                    mapModule.unregisterPlugin(me.togglePlugin);
-                    mapModule.stopPlugin(me.togglePlugin);
+
+            if ( !enabled ) {
```
ZakarFin: This has so many branches that it's difficult to follow. Maybe add some comments at least. Which branch do you expect to activate, when and why.

bundles/statistics/statsgrid2016/instance.js
```diff
@@ -47,15 +50,24 @@ Oskari.clazz.define(
             // disable tile if we don't have anything to show or enable if we do
             // setup initial state
             this.setState();
+            
+            me.getTile().getFlyoutManager().init();
```
ZakarFin: How big of an operation would it be to move flyoutmanager to be in the instance? Since it's called a lot from here. Maybe we should refactor this so tile only handles the ui state of the tile, but the actual flyout handling would be handled in the instance.

bundles/statistics/statsgrid2016/instance.js
```diff
@@ -47,15 +50,24 @@ Oskari.clazz.define(
...
-                me.showToggleButtons(conf.grid !== false, this.state.view);
+
                 // Always show legend on map when embedded
                 me.showLegendOnMap(true);
                 // Classification can be disabled for embedded map
                 me.enableClassification(conf.allowClassification !== false);
+
+                // TODO? better way to call than strings
+                if( me.conf.grid ) {
+                    me.showEmbeddedTools('table', true);
```
ZakarFin: Are these "shown" as in toggled on or just added to the ui so user can open them?

bundles/statistics/statsgrid2016/plugin/TogglePlugin.js
```diff
@@ -30,26 +36,49 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.TogglePlugin', function(sandbox
             return this.element;
         }
         var me = this;
-        var toggleButtons = jQuery('<div class="statsgrid-published-toggle-buttons mapplugin"><div class="map"></div><div class="table"></div>');
+        var toggleButtons = jQuery('<div class="statsgrid-published-toggle-buttons mapplugin"><div class="map"></div><div class="table" style="display:none"></div><div class="diagram" style="display:none"></div>');
         var map = toggleButtons.find('.map');
         var table = toggleButtons.find('.table');
+        var diagram = toggleButtons.find('.diagram');
+        var stats = this.sb.findRegisteredModuleInstance('StatsGrid');
```
ZakarFin: Just pass the reference instead if you need it. But you could just pass the tool "id"s for the UI, make this observable and trigger events so you don't need to have a reference of statsgrid in this class.

bundles/statistics/statsgrid2016/plugin/TogglePlugin.js
```diff
@@ -30,26 +36,49 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.TogglePlugin', function(sandbox
             return this.element;
         }
         var me = this;
-        var toggleButtons = jQuery('<div class="statsgrid-published-toggle-buttons mapplugin"><div class="map"></div><div class="table"></div>');
+        var toggleButtons = jQuery('<div class="statsgrid-published-toggle-buttons mapplugin"><div class="map"></div><div class="table" style="display:none"></div><div class="diagram" style="display:none"></div>');
```
ZakarFin: You could also create the template by looping the tools.

ZakarFin:
I feel the tools handling in toggleplugin is not as good as it could be. There's hardcoded/duplicated information about the tools when it could be just an UI implementation to show with tools that can be passed as a parameter without the plugin knowing what they mean. The naming of functions still feels somewhat misleading regarding this. While this is mergeable with a note that it could/should be improved later on, I do think the functionality is complex as is even without misleading naming etc and will cause problems in the future. I think we should put more effort in to thinking what's the so called "correct place" for a functionality/responsibilities within components that makes sense instead of "just working" code so it's easier to continue working on this. Do you think you can still work on improving this or should I try to find some time to refactor it a bit myself?

bundles/statistics/statsgrid2016/Tile.js
```diff
@@ -63,22 +63,32 @@ function(instance, service) {
      */
     startPlugin : function() {
         this._addTileStyleClasses();
+    },
+    extend: function () {
```
ZakarFin: I would rename this to something like setupTools and pass a reference to the flyoutmanager instead of fetching it through instance.

bundles/statistics/statsgrid2016/instance.js
```diff
@@ -44,18 +47,36 @@ Oskari.clazz.define(
                 me.statsService.setMapModes(['wms','vector']);
             }
             statsService.addDatasource(conf.sources);
+
+            // initialize flyoutmanager
+            this.flyoutManager = Oskari.clazz.create('Oskari.statistics.statsgrid.FlyoutManager', this, statsService);
+            this.flyoutManager.init();
+            //call extend to initialize the flyoutmanager in tile
+            this.getTile().extend();
```
ZakarFin: This would become: 
this.getTile().setupTools(this.flyoutManager);

bundles/statistics/statsgrid2016/instance.js
```diff
@@ -318,23 +339,11 @@ Oskari.clazz.define(
...
-                this.togglePlugin = Oskari.clazz.create('Oskari.statistics.statsgrid.TogglePlugin', this.getSandbox(), this.getLocalization().published);
-            }
-            mapModule.registerPlugin(me.togglePlugin);
-            mapModule.startPlugin(me.togglePlugin);
-            me.togglePlugin.showTable(visible || me.visible);
+        getEmbeddedTools: function () {
+            return this.embeddedTools;
+        },
+        addEmbeddedTool: function ( tool ) {
+            this.embeddedTools.push(tool);
```
ZakarFin: Wasn't embeddedTools removed from this file?

bundles/statistics/statsgrid2016/instance.js
```diff
@@ -82,6 +103,12 @@ Oskari.clazz.define(
         getLayerService : function() {
             return this.getSandbox().getService('Oskari.mapframework.service.MapLayerService');
         },
+        getFlyoutManager: function () {
+            return this.flyoutManager;
+        },
+        getFlyout: function (type) {
+            return this.getFlyoutManager().getFlyout(type);
+        },
```
ZakarFin: Are getFlyoutManager() and getFlyout() needed here? Where are they used?

bundles/statistics/statsgrid2016/plugin/TogglePlugin.js
```diff
@@ -60,19 +74,18 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.TogglePlugin', function(sandbox
...
     redrawUI: function() {
-        var me = this;
-        me._element = me._createControlElement();
-        this.addToPluginContainer(me._element);
+        this.element = this._createControlElement();
+        this.addToPluginContainer( this.element );
     },
     teardownUI : function(stopping) {
-        this.sb.postRequestByName('userinterface.UpdateExtensionRequest',[null, 'close', 'StatsGrid']);
+        // this.sb.postRequestByName('userinterface.UpdateExtensionRequest',[null, 'close', 'StatsGrid']);
```
ZakarFin: Remove commented out code

bundles/statistics/statsgrid2016/instance.js
ZakarFin: SearchFlyout.js references getFlyoutManager(), but the getFlyout() isn't used?

PR #260, opened 2018-02-08
prevent scrollbar while searching

nxholm:
- added overflow-y: hidden to class metadataSearching


ZakarFin:
Doesn’t this remove the scrollbar also when there’s more results than can fit the container?

nxholm:
It should not affect the result list since the div is only visible for the searching duration.
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/catalogue/metadatacatalogue/instance.js#L820

nxholm:
I'll do it in a different way since there appears to be some problems with overriding css instead.

PR #261, opened 2018-02-08
line-height problem asdi overriden css

nxholm:
- remove the line-height specification in the asdi body overriden css.

there is a problem with the metadatacatalogue searching element `.metadataSearching`, it receives scrollbars when the line height is too narrow.


nxholm:
Alternative solution is to add a own class to `.metadataSearching` in asdi overriden.css and give it some padding

PR #263, opened 2018-02-09
Hierarchical layerlist

kuosman:
# New bundles
## hierarchical-layerlist
This bundles adds new layerlis what shows layers in hierarchical list.
hierarchical-layerlist

## admin-hierarchical-layrelist
This bundle extends hierarchical-layerlist bunde for admin tools (group adding/modifying/deleting and layer adding/modifying/deleting).
admin-hierarchical-layerlist

# Others
## Renamings
InpireTheme key is renamed so this need merged same time as oskari-server PR https://github.com/oskariorg/oskari-server/pull/145
Action route also renamed new name.
## Maplayerservice
Added support to groups.

This includes also https://github.com/oskariorg/oskari-frontend/pull/270

# Start bundles
```javascript
Oskari.app.playBundle(
    {
    bundlename : 'hierarchical-layerlist',
        metadata : {
            "Import-Bundle" : {
                "hierarchical-layerlist" : {
                    bundlePath : '/Oskari/packages/framework/bundle/'
                }
            }
        }
});

Oskari.app.playBundle(
    {
    bundlename : 'admin-hierarchical-layerlist',
        metadata : {
            "Import-Bundle" : {
                "admin-hierarchical-layerlist" : {
                    bundlePath : '/Oskari/packages/admin/bundle/'
                }
            }
        }
});
```


ZakarFin:
Implementation for oskariorg/oskari-docs#61. Kinda big change for a single PR :scream:

ZakarFin:
Please update this with the latest develop. Hopefully the amount of commits, files and lines changed will drop dramatically.

api/framework/hierarchical-layerlist/bundle.md
```diff
@@ -0,0 +1,120 @@
...
+  <td> `ChangeMapLayerStyleRequest` </td>
+  <td>Sent when a layer style is selected from the dropdown.</td>
+</tr>
+<tr>
+  <td> `ChangeMapLayerOpacityRequest` </td>
+  <td>Sends out when user changes the opacity for a layer</td>
+</tr>
+<tr>
+  <td> `MapModulePlugin.AddFeaturesToMapRequest` </td>
+  <td> Adds layer extent feature to map and zoomed in it's extent</td>
```
ZakarFin: Should use MapModulePlugin.MapMoveByLayerContentRequest instead for this.

bundles/admin/admin-hierarchical-layerlist/components/Group.js
```diff
@@ -0,0 +1,266 @@
...
+        if (id) {
+            var btnDelete = Oskari.clazz.create('Oskari.userinterface.component.Button');
+            btnDelete.setTitle(me.locale('buttons.delete'));
+            btnDelete.addClass('delete');
+            btnDelete.setHandler(function() {
+                // check at group has no layers or subgroups
+                var group = me.layerService.getAllLayerGroups(id);
+                var isLayersOrGroups = (group && group.layers && group.layers.length > 0) ? true : false;
+                if (group.groups && group.groups.length > 0) {
+                    isLayersOrGroups = true;
```
ZakarFin: This reads like Q: "Are these layers or groups?" A: "Yes" :) If I understand this correctly you want to check if the group has layers or subgroups? These could be functions in the group object: hasLayers() and hasSubgroups(). It would make this more readable:

    if (group.hasLayers() || group.hasSublayers()) { showError(); }

bundles/admin/admin-hierarchical-layerlist/components/Group.js
```diff
@@ -0,0 +1,266 @@
...
+                }
+                if (isLayersOrGroups) {
+                    var dialog = Oskari.clazz.create('Oskari.userinterface.component.Popup');
+                    dialog.show(me.locale('errors.groupnameDeleteCheckLayers.title'), me.locale('errors.groupnameDeleteCheckLayers.message'));
+                    dialog.fadeout(5000);
```
ZakarFin: If you would return here you wouldn't need the 20-line long else block.

bundles/admin/admin-hierarchical-layerlist/components/Group.js
```diff
@@ -0,0 +1,266 @@
...
+                    var confirmBtnCancel = Oskari.clazz.create('Oskari.userinterface.component.Button');
+
+                    confirmBtnOk.addClass('primary');
+                    confirmBtnOk.setTitle(me.locale('buttons.ok'));
+                    confirmBtnOk.setHandler(function() {
+                        confirmDialog.close();
+                        me._deleteGroup(data, popup, opts.type, tool);
+                    });
+
+                    confirmBtnCancel.setTitle(me.locale('buttons.cancel'));
```
ZakarFin: If you would use OkButton and CancelButton (https://github.com/oskariorg/oskari-frontend/tree/master/bundles/framework/divmanazer/component/buttons) you wouldn't need to have the localizations in this bundle.

bundles/admin/admin-hierarchical-layerlist/components/Group.js
```diff
@@ -0,0 +1,266 @@
...
+                    confirmBtnCancel.setHandler(function() {
+                        confirmDialog.close();
+                    });
+
+                    var groupName = me.sandbox.getLocalizedProperty(group.name);
```
ZakarFin: This is deprecated. Use Oskari.getLocalized(property, lang) instead.

bundles/admin/admin-hierarchical-layerlist/components/Layer.js
```diff
@@ -0,0 +1,246 @@
...
+                        me.dataProviders.push({
+                            id: org.id,
+                            name: Oskari.getLocalized(org.name)
+                        });
+                    });
+
+                    me.dataProviders.sort(function(a, b) {
+                        return Oskari.util.naturalSort(a.name, b.name);
+                    });
+                    callback();
```
ZakarFin: I think you should always send me.dataProviders as a parameter to the callback function and always use this function to get a reference to them. That way you can't have a situation where me.dataProviders is empty just because this function hasn't been called.

bundles/admin/admin-hierarchical-layerlist/components/Layer.js
```diff
@@ -0,0 +1,246 @@
...
+    *******************************************************************************************************************************/
+    /**
+     * Get dataproviders
+     * @method getDataproviders
+     * @param  {Boolean}        emptyCache empty cache
+     * @param  {Function}       callback   callback function
+     */
+    getDataproviders: function(emptyCache, callback) {
+        var me = this;
+        if (me.dataProviders === null || emptyCache === true) {
```
ZakarFin: I would reverse this with:

    if(me.dataProviders !== null && !emptyCache) {
        callback(me.dataProviders);
        return;
    }

This way you can get rid of one indentation level for the rest of the function and the if-statement is easier to follow with just 2 lines of code inside it rather than 20.

bundles/admin/admin-hierarchical-layerlist/components/Layer.js
```diff
@@ -0,0 +1,246 @@
...
+     * @param  {String}          tool    tool identifier
+     * @param  {String}          layerId layerId
+     * @param  {String}          groupId groupId
+     */
+    showLayerAddPopup: function(tool, layerId, groupId) {
+        var me = this;
+        me.getDataproviders(false, function() {
+            me._extraFlyout.move(100, 100, true);
+            me._extraFlyout.makeDraggable({
+                handle: '.oskari-flyouttoolbar, .statsgrid-data-container > .header',
```
ZakarFin: Why would this handle reference statsgrid? I'm not sure you need to send any opts as param for this.

bundles/admin/admin-hierarchical-layerlist/instance.js
```diff
@@ -0,0 +1,364 @@
...
+
+                var popupConf = me.group.getGroupAddingPopupConf(tool, groupId, parentId, options);
+                var popup = popupConf.popup;
+                var message = popupConf.message;
+                popupConf.popup.show(me.locale('groupTitles.editSubgroup'), message, popupConf.buttons);
+                popupConf.popup.makeModal();
+            }, {
+                cls: 'edit-subgroup-subgroup',
+                tooltip: me.locale('tooltips.editSubgroup')
+            });
```
ZakarFin: You don't need to change anything here, but did I get this right?

- On the main level you can add a group
- On group level you can add a layer, edit the group and add a subgroup
- On subgroup level you can add a layer, edit the subgroup and add a subgroup subgroup
- On subgroup subgroup level you can add a layer, edit the subgroup subgroup

Does adding a group differ from adding a subgroup or subgroup subgroup? Seems this could have been written without hardcoded levels:

- on main level you can only add a group
- each group can add a layer, edit the group and add a subgroup (except for adding subgroup is not allowed on configurable max depth for group)

bundles/admin/admin-hierarchical-layerlist/instance.js
```diff
@@ -0,0 +1,364 @@
...
+
+        /**
+         * Assigns event handlers to the admin hierarchical layer list.
+         */
+        _addEventHandlers: function() {
+            var me = this;
+            jQuery(document).on("dnd_stop.vakata", function(event, data) {
+                //If the drag target group is not open, we have to open it.
+                //Otherwise we can't get the necessary information of the drag operation.
+                var targetGroup = data.data.origin.get_node(jQuery(data.event.target).prop("id").split("_")[0]);
```
ZakarFin: What's the structure of "data" here? What's the difference between event and data.event? Are dnd-targets restricted to elements that always have id (and with a underscore to split from)? Can you add a comment about expected content of data and event target. Is this the same dnd we use in layerselection2 or another lib?

bundles/admin/admin-hierarchical-layerlist/instance.js
```diff
@@ -0,0 +1,364 @@
...
+         * Save group or list ordering and grouping.
+         * @method  _saveOrder
+         * @param   {Object}   data  data for saving
+         * @param   {Oskari.userinterface.component.Popup}   popup group adding/editing popup
+         * @param   {String}   type  jstree type
+         * @private
+         */
+        _saveOrder: function(event, data) {
+            var me = this;
+            var draggedNode = data.data.origin.get_node(data.data.nodes[0]);
```
ZakarFin: Does data need to be passed around or could you just send draggedNode as param?

ZakarFin: You should restrict the usage of "data" and event param so they are not needed here. Dig out the information you need in the calling function and pass it here as param.

bundles/admin/admin-hierarchical-layerlist/instance.js
```diff
@@ -0,0 +1,364 @@
...
+                    var successDialog = Oskari.clazz.create('Oskari.userinterface.component.Popup');
+                    successDialog.show(me.locale('succeeses.nodeDropSave.title'), me.locale('succeeses.nodeDropSave.message'));
+                    successDialog.fadeout();
+                    me.service.trigger('jstree-order-changed', {
+                        ajax: true,
+                        success: true
+                    });
+                }
+            });
+            me.service.trigger('jstree-order-changed', data);
```
ZakarFin: Why reference library in the event name?

ZakarFin: It's better that we hide the implementation details so it's easier to make changes in the future than promote what 3rd party library is used currently.

bundles/admin/admin-hierarchical-layerlist/instance.js
```diff
@@ -0,0 +1,364 @@
...
+                return this.locale(key);
+            }
+            return this.locale;
+        },
+        getName: function() {
+            return "AdminHierarchicalLayerList";
+        },
+        start: function() {
+            var me = this;
+            if (!me.service) {
```
ZakarFin: Add a comment here like "Only works with hierarchical layer selector, if the service it provides is not available we cannot proceed".

bundles/admin/admin-hierarchical-layerlist/models/layerModel.js
```diff
@@ -0,0 +1,433 @@
...
+                return fToBind.apply(this instanceof fNOP && oThis ? this : oThis,
+                    aArgs.concat(Array.prototype.slice.call(arguments)));
+            };
+
+        fNOP.prototype = this.prototype;
+        fBound.prototype = new fNOP();
+
+        return fBound;
+    };
+}
```
ZakarFin: Bind polyfill is now part of the core, you can remove it from here.

bundles/admin/admin-hierarchical-layerlist/models/layerModel.js
```diff
@@ -0,0 +1,433 @@
...
+// actual model - uses bind to make Oskari layer object functions call BackBone model.attributes
+(function() {
```
ZakarFin: Are there any changes in this file with the one in the current layer admin functionality? Could it be just linked here so a copy wouldn't be needed?

bundles/admin/admin-hierarchical-layerlist/resources/locale/en.js
```diff
@@ -0,0 +1,256 @@
...
+            "dataproviderSave": {
+                "title": "Dataprovider saving failed",
+                "message": "Dataprovider saving failed, try again later"
+            },
+            "maplayerGroups": {
+                "title": "Check maplayer groups",
+                "message": "Maplayer must have least one group"
+            }
+        },
+        "succeeses": {
```
ZakarFin: successMessages?

bundles/admin/admin-hierarchical-layerlist/views/adminLayerSettingsView.js
```diff
@@ -0,0 +1,1545 @@
...
+                        errorDialog.show(me.instance.locale('errors.dataproviderSave.title'), me.instance.locale('errors.dataproviderSave.message'));
+                        errorDialog.fadeout();
+                    },
+                    success: function(response) {
+                        popup.close();
+                        var successDialog = Oskari.clazz.create('Oskari.userinterface.component.Popup');
+                        successDialog.show(me.instance.locale('succeeses.dataproviderSave.title'), me.instance.locale('succeeses.dataproviderSave.message'));
+                        successDialog.fadeout();
+
+                        jQuery('#select-dataprovider').append('<option value="' + response.id + '">' + Oskari.getLocalized(response.name) + '</option>');
```
ZakarFin: Do you remove the previous option while editing somewhere? Or does editing leave the old option and add the new option?

bundles/admin/admin-hierarchical-layerlist/views/adminLayerSettingsView.js
```diff
@@ -0,0 +1,1545 @@
...
+                jQuery.ajax({
+                    type: 'POST',
+                    dataType: 'json',
+                    data: {
+                        name: sldName,
+                        xml: encodeURIComponent(sldXml)
+                    },
+                    url: me.instance.sandbox.getAjaxUrl('SldStyles'),
+                    success: function(resp) {
+                        me._showDialog("title", "New sld saved success / " + sldName);
```
ZakarFin: Hardcoded message. Should use localization.

bundles/admin/admin-hierarchical-layerlist/views/adminLayerSettingsView.js
```diff
@@ -0,0 +1,1545 @@
...
+                        //Update UI
+                        me._sldStylesAppendUI(resp.id, sldName);
+
+                    },
+                    error: function(jqXHR) {
+                        if (jqXHR.status !== 0) {
+                            me._showDialog("title", "Save of new sld xml failed");
```
ZakarFin: Hardcoded message. Should use localization.

bundles/admin/admin-hierarchical-layerlist/views/adminLayerSettingsView.js
```diff
@@ -0,0 +1,1545 @@
...
+             * @method  _sldStylesUI
+             * @param   {Object}     elem jQuery elemen
+             * @private
+             */
+            _sldStylesUI: function(elem) {
+                var me = this,
+                    sldSele = elem.find('#add-layer-sld-style');
+
+                for (var i = 0; me.sldStyles != null && i < me.sldStyles.length; i++) {
+                    sldSele.append('<option value=' + me.sldStyles[i].id + ' >' + me.sldStyles[i].name + '</option>');
```
ZakarFin: Call the below method instead so the same thing is not done in two places. You can send the select element as optional third parameter for it.

bundles/framework/hierarchical-layerlist/Flyout.js
```diff
@@ -0,0 +1,749 @@
...
+
+            me.service.on('option.added', function() {
+                me.populateLayers();
+            });
+
+            me.service.on('admin.changed', function() {
+                me.populateLayers();
+            });
+
+            me.service.on('admin-layer', function(data) {
```
ZakarFin: You should use the same naming scheme instead of some events having words separated with dots while others with dash.

bundles/framework/hierarchical-layerlist/Flyout.js
```diff
@@ -0,0 +1,749 @@
...
+            if (me.instance._selectedLayerGroupId[layerData.id]) {
+                var currentGroupId = me.instance._selectedLayerGroupId[layerData.id];
+                var founded = jQuery.grep(layerData.groups, function(group) {
+                    return group.id === currentGroupId;
+                });
+
+                if (founded.length === 0) {
+                    // update groups to first
+                    me.instance._selectedLayerGroupId[layerData.id] = layerData.groups[0].id;
+                }
```
ZakarFin: You can write this with array.find() (it's polyfilled in core now):

    _updateBreadcrumbGroups: function(layerData) {
        var me = this;
        var currentGroupId = me.instance._selectedLayerGroupId[layerData.id];
        if (!currentGroupId) {
            // requested group not found - ignoring
            return;
        }
        var group = layerData.groups.find(function(group) {
            return group.id === currentGroupId;
        });
        
        if(group) {
            // update groups to first
            me.instance._selectedLayerGroupId[layerData.id] = layerData.groups[0].id;
        }
    }

Though looking at the code it seems kinda fishy. You check if the layer is included in me.instance._selectedLayerGroupId (dictionary of layer id -> group id). If it is you check if the layer has the group that was linked to the layer id included as one of the layers groups. If it is you overwrite the layer to group mapping value with the id of first group for the layer? Does this have something to do with breadcrumbs? The dictionary value for the layer is always the first of the layers groups, but it's tracked with another variable in instance? Maybe naming could be improved so this doesn't feel so disconnected from it's name?

bundles/admin/admin-hierarchical-layerlist/instance.js
kuosman: Yes these are right.  In hierarchical-layerside side there is container for tools. Jstree has also typed nodes. Main level can contains also other tools, for example need i add update all getcapabilities here ?

bundles/admin/admin-hierarchical-layerlist/instance.js
kuosman: dnd_stop.vakata is jstree event and not used layerselection2 bundle.
Jstree bind this event in document (not jstree element) so need be there, handles drag-n drop.
See: https://www.jstree.com/api/#/?f=dnd_stop.vakata 

I add comments.

bundles/admin/admin-hierarchical-layerlist/components/Group.js
```diff
@@ -0,0 +1,258 @@
...
+                parentId: parentId,
+                id: id
+            };
+            var hasValidLocales = true;
+            popup.getJqueryContent().find('.oskari-textinput.group-name').each(function() {
+                var el = jQuery(this);
+
+                var value = el.find('input').val().trim();
+                data.locales[el.attr('data-locale')] = value;
+                if (value.length < 4) {
```
ZakarFin: The minimum allowed length for trimmed group name is 4 characters? Is this some requirement for something? Isn't a single character enough?

bundles/admin/admin-hierarchical-layerlist/components/Group.js
```diff
@@ -0,0 +1,258 @@
...
+            var value = (opts.locale && opts.locale[locale]) ? opts.locale[locale] : null;
+            if (value) {
+                input.setValue(value);
+            }
+            input.addClass('group-name');
+            var el = jQuery(input.getElement());
+            el.find('input').bind('keyup', function() {
+                var inputEl = jQuery(this);
+                var value = inputEl.val().trim();
+                if (value.length < 4) {
```
ZakarFin: Same here with the 4 character requirement. If one character isn't enough move the magic number 4 to a constant so it isn't repeated in multiple places (easier to change afterwards and find the places it affects).

bundles/framework/hierarchical-layerlist/service/LayerlistExtenderService.js
```diff
@@ -0,0 +1,424 @@
+/**
+ * @class Oskari.framework.bundle.hierarchical-layerlist.OskariEventNotifierServic
```
ZakarFin: Update this documentation. It says 'Oskari.framework.bundle.hierarchical-layerlist.LayerlistExtenderService' below.

bundles/framework/hierarchical-layerlist/service/LayerlistExtenderService.js
```diff
@@ -0,0 +1,424 @@
...
+
+        /**
+         * @method create called automatically on construction
+         * @static
+         */
+        function() {
+            var me = this;
+            this.sb = Oskari.getSandbox();
+            this._options = {
+                /* Default options */
```
ZakarFin: Are these options for jstree?

bundles/framework/hierarchical-layerlist/service/LayerlistExtenderService.js
```diff
@@ -0,0 +1,424 @@
...
+                });
+            },
+            /**
+             * Sets admin role
+             * @method setAdmin
+             * @param  {Boolean} isAdmin is admin
+             */
+            setAdmin: function(isAdmin) {
+                this._hasAdmin = isAdmin;
+                this.trigger('admin.changed', this._hasAdmin);
```
ZakarFin: Why does this need to be tracked separately? Why not just react to layer tool changes when admin bundle adds them?

bundles/framework/hierarchical-layerlist/service/OskariEventNotifierService.js
```diff
@@ -0,0 +1,47 @@
...
+            /**
+             * Used to propate Oskari events for files that have reference to service, but don't need to be registered to sandbox.
+             * Usage: service.on('Event', function(evt) {});
+             *
+             * instance.js registers eventhandlers and calls this to let components know about events.
+             * @param  {Oskari.mapframework.event.Event} event event that needs to be propagated to components
+             */
+            notifyOskariEvent: function(event) {
+                this.trigger(event.getName(), event);
+            }
```
ZakarFin: You could move all event handling to go through this service by adding onEvent() function and passing this to registerForEventByName() instead of the instance.

bundles/integration/admin-layerselector/views/adminLayerSettingsView.js
```diff
@@ -53,18 +53,211 @@ define([
...
+                btnOk.addClass('add');
+                btnOk.setHandler(function() {
+                    var selected = [];
+                    var ids = [];
+                    popup.getJqueryContent().find('.admin-maplayer-group input:checked').each(function() {
+                        var el = jQuery(this).parents('label');
+                        var name = el.attr('data-name');
+                        var id = el.attr('data-id');
+                        selected.push({
+                            id: parseFloat(id),
```
ZakarFin: Float id?

bundles/framework/hierarchical-layerlist/service/LayerlistExtenderService.js
kuosman: Yes, these are jstree options.

bundles/framework/hierarchical-layerlist/service/LayerlistExtenderService.js
kuosman: Only thing what use this is at when user is admin then shows also empty groups in layers so admin can add edit/delete group and also add layers.

bundles/framework/hierarchical-layerlist/service/LayerlistExtenderService.js
ZakarFin: Maybe add a comment or rename variable to clarify? Many files have similar structure where these kind of options (and state) are those of the Oskari class.

bundles/framework/hierarchical-layerlist/service/LayerlistExtenderService.js
ZakarFin: Maybe rename this for clarity? Show/IncludeEmptyGroups() or add a comment

ZakarFin:
Note! This requires at least 1.10.2 version of jQuery. So any custom oskari-server-extension might need to modify the app-specific JSP file to link a newer one:

    -    <script type="text/javascript" src="//code.jquery.com/jquery-1.7.2.min.js">
    +    <script type="text/javascript" src="//code.jquery.com/jquery-1.10.2.min.js">

ZakarFin:
jQuery.live() has been removed in version 1.9. Myplaces functionality uses $.live() so this cannot be used with myplaces.

ZakarFin:
While I merged this to make future improvements on it easier to review I will be downgrading the jQuery version of the sample app to fix myplaces functionality.

PR #266, opened 2018-02-12
Fix/drawtools stopdrawing

okauppinen:
Now DrawTools tries to finish currently drawn geometry (sketch) on stopdrawing request. Fixes the bug where unfinished (not dblclicked) feature were added to drawing event and not to draw layer. Now stopdrawing request adds unfinished feature to draw layer before sending event if  2 points were drawn for lines or 3 points for area. Also updates measurements on map.

Now clearDrawing removes features from all drawlayers if id isn't given. Fixes the bug where another tool/bundle haven't cleared drawings before new tool/bundle starts to draw. StopDrawing request must be send before StartDrawing request to be sure that the drawing layers are cleared.

Publisher custom tile element configuration is now checked in init method instead of constructor.

### DIFF
--- a/bundles/framework/myplaces3/view/MainView.js
+++ b/bundles/framework/myplaces3/view/MainView.js
@@ -123,6 +123,7 @@ Oskari.clazz.define("Oskari.mapframework.bundle.myplaces3.view.MainView",
                             this._setMeasurementResult(event.getData());
                         }
                     }else if(this.instance.isFinishedDrawing()){
+                        console.log(event);
                         this._handleFinishedDrawingEvent (event);
                     }
                 }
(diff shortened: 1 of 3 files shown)


bundles/framework/myplaces3/view/MainView.js
```diff
@@ -123,6 +123,7 @@ Oskari.clazz.define("Oskari.mapframework.bundle.myplaces3.view.MainView",
                             this._setMeasurementResult(event.getData());
                         }
                     }else if(this.instance.isFinishedDrawing()){
+                        console.log(event);
```
ZakarFin: Remove console.log. You can use Oskari.log('loggername').debug(msg) instead if you add a little context to the message.

PR #267, opened 2018-02-13
Added data protection description

jheijari:


ZakarFin:
These are for Paikkatietoikkuna.fi specific bundle

PR #268, opened 2018-02-13
coordinatetransformation/functionality refactor

nxholm:
- Transformed and non-transformed coordinates are now kept track of in a object instead of reading from DOM table.
- Fixed a lot of logic in when to show what options for the dropdowns
- Added helper functions to SelectList component to show and hide their content.
- Renamed a bunch of functions for clarity


bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -82,10 +82,27 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
...
     },
-    /**  @method addOption appends a new option to the select
-     *   @param { Object } object with keys id and title
+    /**  
+     * @method hideOptions hide all options except placeholder
+     */
+    hideOptions: function () {
+      this.resetToPlaceholder();
+      var chosen = this.element.find('select');
+      chosen.find('option').not(":first-child").css('display', 'none');
```
ZakarFin: I'd say this is bound to cause problems. How about we just hide the whole dropdown and possibly replace it with a text message where used if applicable.

bundles/paikkatietoikkuna/coordinatetransformation/components/CoordinateSystemSelection.js
```diff
@@ -0,0 +1,244 @@
...
+        populateSelect: function () {
+            var me = this;
+            var wrapper = this.getElement();
+            this.select.create();
+            this.selectInstance = this.select.getSelectInstances();
+            this.dropdowns = this.select.getDropdowns();
+            var i = 0;
+            Object.keys( this.dropdowns ).forEach( function( key ) {
+                jQuery( wrapper.find( '.transformation-system' ).find( '.select' )[i] ).append( me.dropdowns[key] );
+                i++;
```
ZakarFin: This seems kinda fishy. Could the key be used to find the corresponding placeholder instead?

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -82,10 +82,25 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
     update: function() {
       this.element.find('select').trigger('chosen:updated');
     },
-    /**  @method addOption appends a new option to the select
-     *   @param { Object } object with keys id and title
+    /**  
+     * @method disable or enable select
+     * @param { boolean } true = disable
+     */
+    disable: function ( disableFlag ) {
```
ZakarFin: It's more common to use setEnabled(boolean) name with this kind of function.

bundles/paikkatietoikkuna/coordinatetransformation/components/CoordinateSystemSelection.js
```diff
@@ -0,0 +1,241 @@
...
+         * and fills it with data
+         */
+        populateSelect: function () {
+            var me = this;
+            var wrapper = this.getElement();
+            this.select.create();
+            this.selectInstance = this.select.getSelectInstances();
+            this.dropdowns = this.select.getDropdowns();
+            Object.keys( this.dropdowns ).forEach( function( key ) {
+                jQuery( wrapper.find( '.transformation-system' ).find( me.makeClassSelector(key) ).append( me.dropdowns[key] ));
```
ZakarFin: Wouldn't it make more sense to bind the changelisteners here while creating the dropdowns instead of another function (handleSelectionChanged();) that finds new references for the dropdowns that are added here?

bundles/paikkatietoikkuna/coordinatetransformation/components/CoordinateSystemSelection.js
```diff
@@ -0,0 +1,241 @@
...
+                return;
+            }
+            if ( typeof keyToEmpty === 'undefined') {
+                Object.keys( dropdowns ).forEach( function ( key ) {
+                    dropdowns[key].find( 'option' ).hide();
+                    dropdowns[key].find( valueClass ).show();
+                });
+            } else {
+                Object.keys( dropdowns ).forEach( function ( key ) {
+                    dropdowns[keyToEmpty].find( 'option' ).hide();
```
ZakarFin: Looks like this does the same thing on each loop round. Can it be moved outside the loop like just before it?

bundles/paikkatietoikkuna/coordinatetransformation/components/CoordinateSystemSelection.js
```diff
@@ -0,0 +1,241 @@
...
+            var current = this.selectInstance[e.currentTarget.dataset.system];
+            var currentValue = current.getValue();
+            this.handleSelectValueChange(currentValue);
+        },
+        /**
+         * @method handleDropdownOptions
+         * @param {string} valueClass - class selector to show options for 
+         * @param {string} keyToEmpty - optional param to empty only one specific key in the dropdown 
+         */
+        handleDropdownOptions: function (valueClass, keyToEmpty) {
```
ZakarFin: If I understand this correctly the dropdowns have multiple sets of options. Sometimes all the options are shown but sometimes some of them are hidden from the user. Option-tags have a class that is used to identify sets of options. Options belonging to a given set (valueClass) can be easily shown after all options have been hidden. "keyToEmpty" is used if only options from one dropdown needs to be hidden before showing all options in all the dropdowns having valueClass. So it assumes the other dropdowns have the non-valueClass options hidden already OR it doesn't matter if they are hidden as long as the valueClass options are visible?

Could this be written more clearly as updateDropdownOptions(dropdownId, valueSetId) where it only ever updates a single dropdown, dropdown is key/keyToEmpty from this method and valueSetId is valueClass? When you need to do a full update you could do:

                Object.keys( dropdowns ).forEach( function ( key ) {
                    // for showing all options don't send valueSetId parameter
                    me.updateDropdownOptions(key, valueSetId); 
                });

bundles/paikkatietoikkuna/coordinatetransformation/components/SourceSelect.js
```diff
@@ -0,0 +1,70 @@
...
+                    '<input type="radio" id="clipboard" name="load" value="2"><label for="clipboard"><span></span> ${clipboard} </label>'+
+                    '<input type="radio" id="file" name="load" value="1"><label for="file"> <span></span> ${file} </label>'+
+                    '<input type="radio" id="mapselection" name="load" value="3"><label for="mapselection"> <span></span> ${map} </label>'+
+                '</form>'+
+                '</div>'
+            ),
+            info: _.template(
+                '<div class="datasource-info">' +
+                    '<div class="coordinateconversion-clipboardinfo" style=display:none;">'+
+                        '<div class="clipboardinfo"> <i><%= clipboardupload %><i> </div>'+
```
ZakarFin: You are using two different styles for placeholders in the same template: <%= clipboardupload %>and ${mapinfo}. Please choose one. Preferrably ${mapinfo} as it's more common in the codebase.

bundles/paikkatietoikkuna/coordinatetransformation/components/SourceSelect.js
```diff
@@ -0,0 +1,70 @@
...
+                    '<div class="coordinateconversion-mapinfo" style=display:none;">'+
+                        '<div class="mapinfo">'+
+                            '<i>${mapinfo}<i> </br>'+
+                            '<input type="button" class="selectFromMap" name="load" value="${map}">' +
+                        '</div>' +
+                    '</div>' +
+                '</div>'
+            )
+        }
+        if (this.element === null) {
```
ZakarFin: This is always true so you can remove it. Or add if (this.element !== null) { return; } at the beginning of createUi()

bundles/paikkatietoikkuna/coordinatetransformation/components/table.js
```diff
@@ -42,45 +41,93 @@ Oskari.clazz.define('Oskari.coordinatetransformation.component.table', function(
...
-
                 table.find('.coordinatefield-table h5').append(rowcounter);
 
                 this.setElement( table );
 
+                this.bindRowCountListener();
+
                 var coords = {};
                 for( var i = 0; i < 10; i++ ) {
                     table.find("#oskari-coordinate-table").append(this.template.row( { coords: coords } ) );
```
ZakarFin: Did I read correctly: "table" is a container where you insert "coordinatefield" that contains the actual table? What is the magic number 10? You could find reference to the table outside the loop and only append the rows in the loop.

bundles/paikkatietoikkuna/coordinatetransformation/util/helper.js
```diff
@@ -63,7 +63,7 @@ Oskari.clazz.define('Oskari.coordinatetransformation.helper', function(instance,
                 1: { "id":"DATUM_KKJ", "title":"KKJ", "cls":"DATUM_KKJ DATUM_EUREF-FIN DATUM_DEFAULT"},
                 2: { "id":"DATUM_EUREF-FIN", "title":"EUREF-FIN", "cls":"DATUM_KKJ DATUM_EUREF-FIN DATUM_DEFAULT"}
                 },
-            "dimension": {
+            "coordinate": {
                 0: { "id":"KOORDINAATISTO_DEFAULT", "title":"Mikä tahansa" },
```
ZakarFin: Some hardcoded localization?

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -82,10 +82,25 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
     update: function() {
       this.element.find('select').trigger('chosen:updated');
     },
-    /**  @method addOption appends a new option to the select
-     *   @param { Object } object with keys id and title
+    /**  
+     * @method enable or disable select
+     * @param { boolean } true = enable
+     */
+    enable: function ( enableFlag ) {
```
ZakarFin: To be consistent I still would use setEnabled() as for example in https://github.com/nxholm/oskari-frontend/blob/71ad715ace33259dfef51fee2a0f9a4b15e5b86d/bundles/framework/divmanazer/component/FormComponent.js#L70

ZakarFin: As SelectList extends FormComponent it would have both enable() and setEnabled() so you can just call setEnabled and name this _setEnabledImpl(): https://github.com/nxholm/oskari-frontend/blob/71ad715ace33259dfef51fee2a0f9a4b15e5b86d/bundles/framework/divmanazer/component/FormComponent.js#L86

ZakarFin: To clarify:
- rename this function _setEnabledImpl
- rename any calls to enable -> setEnabled

bundles/paikkatietoikkuna/coordinatetransformation/components/CoordinateSystemSelection.js
ZakarFin: Would this work?

    var dropdown = me.dropdowns[key];
    jQuery( wrapper.find( '.transformation-system' ).find( me.makeClassSelector(key) ).append( dropdown ));
    dropdown.on('change', function () {
        var value = jQuery(this).val();
        me.handleSelectValueChange(value);
    });

Then you could remove the next loop and the handleSystemChange-function.

bundles/paikkatietoikkuna/coordinatetransformation/components/CoordinateSystemSelection.js
```diff
@@ -0,0 +1,232 @@
...
+            var me = this;
+            var dropdowns = this.dropdowns;
+            var instances = this.selectInstance;
+            var datum = instances.datum.getValue();
+            var coordinate = instances.coordinate.getValue();
+            var projection = instances.projection.getValue();
+            var clsSelector = this.makeClassSelector;
+
+            var table;
+            var system = this.element.attr('data-system');
```
ZakarFin: is this.element the wrapper for the whole ui or something else here? Does it gain the data-system attribute somewhere?

nxholm: this.element is the wrapper for the coordinate system selection dropdowns, it gains the data-system attribute on creation (https://github.com/nxholm/oskari-frontend/blob/coordtrans/functionality-fixes/bundles/paikkatietoikkuna/coordinatetransformation/view/transformation.js#L73)

PR #270, opened 2018-02-15
Renamings, layerlist filter service and layertools

kuosman:
Renamed action routes and keys to work https://github.com/oskariorg/oskari-server/pull/145.

Layerselector filters now created from service.
Layer tools changes, metadata/featuredata bundles add their layertools now.


bundles/framework/layerselection2/Flyout.js
```diff
@@ -873,9 +784,12 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselection2.Flyout',
...
             for (s = 0; s < laytools.length; s += 1) {
                 laytool = laytools[s];
-                if (laytool) {
+                if (laytool && laytool.getName() !== 'show-layer-legend-tool') {
                     // Icon or text link
-                    if (laytool.getIconCls()) {
+                    if (laytool.getIconCls() &&
+                        laytool.getName() !== 'ownStyle' &&
+                        laytool.getName() !== 'objectData' &&
+                        laytool.getName() !== 'content-editor') {
```
ZakarFin: What's going on in here?

bundles/catalogue/metadataflyout/resources/locale/en.js
```diff
@@ -1,9 +1,9 @@
-Oskari.registerLocalization(
-{
+Oskari.registerLocalization({
     "lang": "en",
     "key": "catalogue.bundle.metadataflyout",
     "value": {
         "title": "Metadata",
+        "tooltip": "Show metadatas",
```
ZakarFin: I think this should be "Show metadata".

bundles/framework/layerselection2/resources/css/style.css
```diff
@@ -55,11 +55,22 @@ div.layerselection2 .layer.selected {
           padding-top: 2px; }
         div.layerselection2 .layer.selected div.layer-tools div.right-tools div.object-data a {
           padding-left: 5px; }
+        div.layerselection2 .layer.selected div.layer-tools div.right-tools div.object-data div.show-metadata-tool {
```
ZakarFin: When metadata tool is added by metadataflyout bundle I'd say the icon/css for the icon should be on that bundle instead of layerselection.

bundles/framework/layerselection2/resources/css/style.css
```diff
@@ -55,11 +55,22 @@ div.layerselection2 .layer.selected {
...
+          width: 16px;
+          height: 16px;
+          background-image: url("../images/icon-info.png");
+          background-repeat: no-repeat; }
+          div.layerselection2 .layer.selected div.layer-tools div.right-tools div.object-data div.show-metadata-tool:hover {
+            background-image: url("../images/icon-info_hover.png"); }
+        div.layerselection2 .layer.selected div.layer-tools div.right-tools div.object-data div.show-filter-tool {
```
ZakarFin: Same here. Maybe add a flag if a tool that is added for a layer should be shown with an icon OR text. The containers for the icons could be styled here, but the actual icons should be in the bundles that add the tool.

bundles/integration/admin-layerselector/models/layersTabModel.js
```diff
@@ -191,19 +191,18 @@
...
                     url: me.baseURL + me.actions.load + "&iefix=" + (new Date()).getTime(),
-                    success: function (pResp) {
+                    success: function(pResp) {
                         // cleanup old layer Groups since we always refresh from server for now
                         me.layerGroups = [];
                         me.loadGroups(pResp, groupingMethod);
                     },
-                    error: function () {
-                    }
+                    error: function() {}
```
ZakarFin: There's really no point in having an error callback if it does nothing?

bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -5,32 +5,85 @@
...
+                toolOwnStyle.setIconCls('show-own-style-tool');
                 toolOwnStyle.setTooltip(locOwnStyle);
-                toolOwnStyle.setCallback(function () {
+                toolOwnStyle.setCallback(function() {
                     me.sandbox.postRequestByName('ShowOwnStyleRequest', [layer.getId()]);
                 });
                 layer.addTool(toolOwnStyle);
             }
 
+            if (layer.isLayerOfType("WFS") || layer.isLayerOfType("ANALYSIS")) {
```
ZakarFin: Analysislayers has it's own modelbuilder: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapanalysis/domain/AnalysisLayerModelBuilder.js move the analysis stuff there.

ZakarFin: Or maybe do something like layer.isFilterSupported() that returns true for WFS and ANALYSIS layers, but not for myplaces or userlayers.

bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -5,32 +5,85 @@
...
+
+                    var filterDialog = Oskari.clazz.create('Oskari.userinterface.component.FilterDialog', fixedOptions);
+                    filterDialog.setUpdateButtonHandler(function(filters) {
+                        // throw event to new wfs
+                        var evt = me.sandbox.getEventBuilder('WFSSetPropertyFilter')(filters, layer.getId());
+                        me.sandbox.notifyAll(evt);
+                    });
+
+                    if (me.service) {
+                        var aggregateAnalyseFilter = Oskari.clazz.create('Oskari.analysis.bundle.analyse.aggregateAnalyseFilter', null, filterDialog); // todo loc
```
ZakarFin: This should definitely be in the analysis modelbuilder.

bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -61,9 +114,9 @@ Oskari.clazz.define(
             layer.setWpsLayerParams(mapLayerJson.wps_params);
```
ZakarFin: This should be moved to analysis modelbuilder as well

bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -5,32 +5,85 @@
...
+        this.service = null;
     }, {
+
+        /**
+         * @private @method _checkIfAggregateValuesAreAvailable
+         * function gives value to addLinkToAggregateValues (true/false)
+         *
+         * @return {Boolean}
+         */
+        _checkIfAggregateValuesAreAvailable: function() {
```
ZakarFin: This goes to analysis layerbuilder as well.

bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -5,32 +5,85 @@
...
          * @param {Object} mapLayerJson JSON presentation of the layer
          * @param {Oskari.mapframework.service.MapLayerService} maplayerService not really needed here
          */
-        parseLayerData: function (layer, mapLayerJson, maplayerService) {
-            var me = this,
-                toolBuilder = Oskari.clazz.builder('Oskari.mapframework.domain.Tool');
+        parseLayerData: function(layer, mapLayerJson, maplayerService) {
+            var me = this;
 
             if (layer.isLayerOfType("WFS")) {
```
ZakarFin: This method is called from userlayer, myplaces, analysis modelbuilders. This is why it checks if the layer is WFS even when it's a WFS modelbuilder. While testing that nothing breaks with the changes this could be refactored. Looks like nothing but the default style is set for any non-WFS layertypes.

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -370,7 +374,7 @@ Oskari.clazz.define(
          * @param {String} param
          *          inspire theme name under which the layer is listed in UI
          */
-        setInspireName: function (param) {
+        setInspireName: function(param) {
```
ZakarFin: Shouldn't this be setLayerGroup()?

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -390,13 +394,13 @@ Oskari.clazz.define(
          * @param {String} lang language id like 'en' or 'fi' (optional)
          * @return {String} inspire theme name under which the layer is listed in UI
          */
-        getInspireName: function (lang) {
+        getInspireName: function(lang) {
```
ZakarFin: Shouldn't this be getLayerGroup()?

bundles/mapping/mapmodule/domain/AbstractLayer.js
ZakarFin: Looks like you've added get/setGroups(). Can these be removed then?

ZakarFin: Or at least set/get the first index of groups?

bundles/integration/admin-layerselector/views/tabPanelView.js
```diff
@@ -217,34 +229,34 @@ define([
...
+                            "cancel": {
+                                "title": instance.getLocalization('cancel')
                             }
                         }
-                };
+                    };
 
                 // override some UI texts for inspire theme form
-                if(tabId === 'inspire') {
+                if (tabId === 'inspire') {
```
ZakarFin: Should this be layergroup instead of inspire?

ZakarFin: I guess it can be inspire since that's what the localization says in this bundle. So no need to change it.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -64,20 +65,24 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
-                jQuery(me._newestLayers).each(function(index, layer){
-                   ids.push(layer.getId());
+                jQuery(me._newestLayers).each(function(index, layer) {
+                    ids.push(layer.getId());
                 });
                 return (jQuery.inArray(layer.getId(), ids) !== -1);
             }
         };
+
+        this.layerlistFilterButtons = {};
```
ZakarFin: I don't think it's a good idea to gather UI-stuff in this service. Make a new service provided by layerselector bundle instead?

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +346,222 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+                });
+                me._layerGroups = newGroups;
+            } else {
+                var editable = me.getAllLayerGroups(data.parentId);
+                editable.groups.forEach(function(group) {
+                    if (group.id != data.id) {
+                        newGroups.push(group);
+                    }
+                });
+                editable.groups = newGroups;
```
ZakarFin: Nothing is done for the "editable" variable so this whole else block doesn't really do anything.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +346,222 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+            var groups = [];
+            var group = null;
+
+            if (me._layerGroups.length === 0) {
+                return;
+            }
+
+            // update or insert
+            if (!deleteLayer) {
+                var isLayerInGroup = function(groupLayers) {
```
ZakarFin: Rename isLayerInGroup to something like layerIndexInGroup as that's what the value is.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +346,222 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+                            founded = i;
+                            break;
+                        }
+                    }
+                    return founded;
+                };
+
+
+                var isGroup = function(groupId) {
+                    var isInGroup = jQuery.grep(newLayerConf.groups, function(group) {
```
ZakarFin: Somewhere else in this PR array.filter() was used. Use that here as well to be consistent and as it's native/doesn't depend on jQuery.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +346,222 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+                            }
+                        }
+                        // check subgroup subgroups
+                        if (subgroup.groups) {
+                            for (var k = 0; k < subgroup.groups.length; k++) {
+                                var subgroupSubgroup = subgroup.groups[k];
+                                layerIndex = isLayerInGroup(subgroupSubgroup.layers);
+                                if (layerIndex >= 0) {
+                                    if (deleteLayer) {
+                                        group.groups[j].groups[k].layers.splice(layerIndex, 1);
```
ZakarFin: Refactor and use recursion instead of groups.groups.groups and i j k loops. This function implementation is too long even without multiple levels of ifs and loops inside each other. Maybe break it into smaller pieces?

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +346,222 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+                        me.getAllLayerGroups(group.id).layers.push(newLayerConf);
+                    }
+                });
+            }
+
+            // Also check if layer has removed from groups
+            if (deleteLayer) {
+                // find all groups where layer is
+                // check main groups
+                var hasInGroupLayers = function(group) {
```
ZakarFin: This and the layer index function below have already been implemented in the code above inside the same gigantic function. You can do this with just one generic helper function that takes an array and an id, loops the array to find if any item in the array has a property id with the value of the parameter and return the index or -1 if not found. This can be used for layers and groups as a single implementation instead of having the same thing written 4 times inside the monstrously large function :) Also consider moving atleast delete to it's own function.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +346,222 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+                var getLayerIndex = function(layers) {
+                    var returnIndex = -1;
+                    layers.forEach(function(layer, index) {
+                        if (layer.id === layerId) {
+                            returnIndex = index;
+                        }
+                    });
+                    return returnIndex;
+                };
+                me._layerGroups.forEach(function(group, index) {
```
ZakarFin: This is basically the same loop-construct that's in the insert/update part of the method as well. Implemented with forEach here and for(var i=0...) above. I'm pretty sure this can be written in a more maintainable style without that much duplication.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -437,11 +666,110 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+                        // continue with next layer
+                        return;
+                    }
+
+                    if (me._reservedLayerIds[mapLayer.getId()] !== true) {
+                        me.addLayer(mapLayer, true);
+                    }
+                });
+
+                group.groups.forEach(function(subgroup) {
```
ZakarFin: Consider using recursion to go from groups to subgroups to subsubgroups.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -437,11 +666,110 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
         },
 
+        /**
+         * @method loadAllLayerGroupsAjax
+         * Loads layers JSON using the ajax URL given on #create()
+         * and parses it to internal layer objects by calling #createMapLayer() and #addLayer()
+         * @param {Function} callbackSuccess method to be called when layers have been loaded succesfully
+         * @param {Function} callbackFailure method to be called when something went wrong
+         */
+        loadAllLayerGroupsAjax: function(callbackSuccess, callbackFailure) {
```
ZakarFin: If the layers are now loaded using this function the old one should be removed.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -450,9 +778,49 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+         * Returns an array of layer groups added to the service
+         * @param {String|Integer} id if defined return only wanted group
+         * @return {Mixed[]/Oskari.mapframework.domain.WmsLayer[]/Oskari.mapframework.domain.WfsLayer[]/Oskari.mapframework.domain.VectorLayer[]/Object[]}
+         */
+        getAllLayerGroups: function(id) {
+            if (id) {
+                var filterFunction = function(group) {
+                    return group.id == id;
+                };
+                var group = this._layerGroups.filter(filterFunction)[0];
```
ZakarFin: The Java-person inside me screams ArrayIndexOutOfBounds, but taking the first item of an empty array doesn't seem to bother JS too much so I guess this is ok.

bundles/mapping/mapmodule/service/map.layer.js
ZakarFin: The current code assumes that there's always a group.group.group structure in the response and each have the layers array as well. I guess most applications won't have that so you should check if the structure is there.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -437,11 +666,110 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+            pResp.forEach(function(group) {
+                group.layers.forEach(function(layer) {
+                    var mapLayer = me.createMapLayer(layer);
+                    if (!mapLayer) {
+                        // unsupported map type, skip
+                        // continue with next layer
+                        return;
+                    }
+
+                    if (me._reservedLayerIds[mapLayer.getId()] !== true) {
```
ZakarFin: Does this mean that if the layer is in multiple groups the same json-data for the layer is repeated for each group it is in?

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -450,9 +778,49 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+        getAllLayerGroups: function(id) {
+            if (id) {
+                var filterFunction = function(group) {
+                    return group.id == id;
+                };
+                var group = this._layerGroups.filter(filterFunction)[0];
+                // group not found
+                // try to get subgroup
+                if (!group) {
+                    for (var i = 0; i < this._layerGroups.length; i++) {
```
ZakarFin: Is there a reason why you have for (var i = 0;...) and forEach-loops taking turns? I think you should use forEach() if there's not some good reason for not using it.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -568,23 +936,52 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
             }
 
-            if(me.layerFilters[filterId]) {
+            if (me.layerFilters[filterId]) {
                 Oskari.log(this.getName()).warn('[MapLayerService] "' + filterId + '" -layer filter has allready defined. Not register layer filter.');
                 return;
             }
             me.layerFilters[filterId] = filterFunction;
         },
+        registerLayerlistFilterButton: function(text, tooltip, cls, filterId) {
```
ZakarFin: In my opinion UI-button references shouldn't be stored here.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -594,18 +991,56 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+
+            if (!filterFunction) {
+                Oskari.log(this.getName()).warn('[MapLayerService] not found layer filter "' + filterId + '". Returning all layer groups.');
+                return allLayerGroups;
+            }
+            var filteredLayerGroups = [];
+            allLayerGroups.forEach(function(group) {
+                var filteredLayers = [];
+                group.layers.forEach(function(layer) {
+                    var mapLayer = me.createMapLayer(layer);
```
ZakarFin: The name suggests some kind of filtering, but layers are being created here. What's going on?

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -951,17 +1387,20 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
-            if(mapLayerJson.created && isNaN(Date.parse(mapLayerJson.created)) === false){
+            if (mapLayerJson.created && isNaN(Date.parse(mapLayerJson.created)) === false) {
                 var created = new Date(mapLayerJson.created);
-                if(created) {
+                if (created) {
                     layer.setCreated(created);
                 }
             }
 
+            layer.setGroups(mapLayerJson.groups);
```
ZakarFin: Looks like the server response you are expecting is to have groups that have layers that have groups. Since layers need much more information than groups I would suggest that you review the datamodel that's being sent from the server. Maybe have layers in one array, groups in another and just use id's to link between them on the client-side.

applications/sample/servlet/minifierAppSetup.json
```diff
@@ -1,410 +1,434 @@
 {
-	"startupSequence": [
```
ZakarFin: I can see that there's 24 lines added, but about 3000 lines worth of changes... Not too easy to see what's been added. I guess nothing should have been added as the new bundles are in another PR?

applications/sample/servlet/minifierAppSetup.json
```diff
@@ -1,410 +1,434 @@
...
+		"bundlename": "content-editor",
+		"metadata": {
+			"Import-Bundle": {
+				"content-editor": {
+					"bundlePath": "../../../packages/tampere/bundle/"
+				}
+			}
+		}
+	}, {
+		"bundlename": "hierarchical-layerlist",
```
ZakarFin: These are not available on this PR.

bundles/mapping/mapmodule/service/map.layer.js
ZakarFin: Oh, it modifies the internals of the class variable returned by getAllLayerGroups(). If getAllLayerGroups(-1) would return this._layerGroups you wouldn't need the if-statement at all. Also the function can just take parentID and id instead of a data-object?

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +344,191 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+                        break;
+                    }
+
+                }
+
+
+                // Finally check at layer has all groups in dom
+                newLayerConf.groups.forEach(function(group) {
+                    var groupConf = me.getAllLayerGroups(group.id);
+                    var isInGroup = jQuery.grep(me.getAllLayerGroups(group.id).layers, function(layer) {
```
ZakarFin: You are using both Array.filter and jQuery.grep in this file. Can you change this to use filter?

packages/framework/bundle/layerselector2/bundle.js
```diff
@@ -42,18 +45,6 @@ Oskari.clazz.define("Oskari.mapframework.bundle.layerselector2.LayerSelectorBund
         }, {
             "type": "text/javascript",
             "src": "../../../../bundles/framework/layerselector2/view/PublishedLayersTab.js"
-        }, {
-            "type": "text/javascript",
-            "src": "../../../../bundles/framework/layerselector2/request/ShowFilteredLayerListRequest.js"
-        }, {
```
ZakarFin: Removing requests should be documented in the changelog and the corresponding md-documents should be removed. Also the possibility of showing a filtered list programmatically is removed in this PR? This is used at least in publisher to show only publishable layers for the user and analysis to show only wfslayers as those are the ones that can be used in analysis.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -329,10 +344,191 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
+
+        /**
+         * Update layer groups
+         * @method updateGroupsLayers
+         * @param  {String}          layerId      layerid
+         * @param  {Object}          newLayerConf new layer conf
+         * @param  {Boolean}         deleteLayer delete layer
+         * @param  {Boolean}         newLayer is new layer
+         */
+        updateGroupsLayers: function(layerId, newLayerConf, deleteLayer, newLayer) {
```
ZakarFin: There's a function updateLayerGroups and updateGroupsLayers. With comments "Updata layer groups" and "Update layer groups". Please check the comments so they are relevant and try to find a better name for updateGroupsLayers.

ZakarFin: Is updateGroupsLayers updating the layers inside groups? If so maybe it could be updateLayersInGroups?

ZakarFin: Why is the newLayerConf passed here and not the instance of AbstractLayer?

bundles/framework/layerselection2/Flyout.js
```diff
@@ -873,9 +785,9 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselection2.Flyout',
 
             for (s = 0; s < laytools.length; s += 1) {
                 laytool = laytools[s];
-                if (laytool) {
+                if (laytool && laytool.getName() !== 'show-layer-legend-tool') {
```
ZakarFin: Why does 'show-layer-legend-tool' need special handling?

ZakarFin:
There's still a call to setInspireName() for grouplayers causing problems:

    map.layer.js:1078 Uncaught TypeError: baseLayer.setInspireName is not a function
       at e._createGroupMapLayer (map.layer.js:1078)
       at e.createMapLayer (map.layer.js:1004)

It should be replaced with layer.setGroups(mapLayerJson.groups); as in "simple layers"?

PR #271, opened 2018-02-19
Layer admin: multi-projection support

data-ux:
Added to layer admin UI:

- Shows supported SRS
- Shows "Missing SRS": SRS that are part of default app views, but not supported by current layer
- Allows adding "Forced SRS": SRS that will be considered supported even if GetCapabilities does not return the SRS as supported.


ZakarFin:
Is there a reason for updating tools/npm-shrinkwrap.json? Otherwise looks good for merging.

data-ux:
Updating npm-shrinkwrap.json isn't strictly necessary here, but new version of NPM adds some new metadata to the shrinkwrap definition: `integrity` and `requires`.

We should transition to using package-lock.json anyway, as it is the default dependency freezing method nowadays in NPM:
https://github.com/npm/npm/blob/latest/doc/spec/package-lock.md

PR #272, opened 2018-02-21
Statsgrid - diagram & grid changes

nxholm:
- Fixed a bug where the indicator didn't show correctly on first open
- Added functionality to show negative values
- Added functionality to sort the chart by ascending value or alphabetical order
- Created a separate svg to show the x-axis values so they are visible when scrolling the chart
- Fixed errors printing to console


bundles/framework/divmanazer/component/Chart.js
```diff
@@ -134,29 +195,95 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function() {
...
+            .attr('x2', function ( d ) {
+                if ( d === null ) {
+                    return;
+                }
+                if ( negatives !== undefined ) {
+                    if ( negatives.indexOf( d ) !== -1 && negatives !== undefined ) {
+                        return 5;
+                    } else {
+                        return -5;
+                    }
```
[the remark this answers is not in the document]
nxholm: the negatives array is there because "this.y" domain only contains the names of the datapoints, not the actual values.
https://github.com/nxholm/oskari-frontend/blob/statsgrid/diagram-grid-changes/bundles/framework/divmanazer/component/Chart.js#L69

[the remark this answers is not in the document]
data-ux: Yes, I see. Maybe in this complex y-axis rendering  custom code for drawing the y-axis ticks & labels would be better? I mean creating a new SVG group and binding the actual data in `this.data` to some new elements in the same way the bars are rendered. This way we have access to the whole datapoint; name and value.

Another solution would be to change Chart.js to use the whole data array as the domain (instead of mapping out just the names). For this to work, we'd need to implement toString() on the datapoint objects, though:
https://github.com/d3/d3-scale#ordinal_domain

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -57,25 +56,85 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function() {
...
+                });
+            },
+            "name-descending": function () {
+                me.data.sort(function (a, b) {
+                    return d3.ascending(a.name , b.name);
+                });
+            },
+            "value-ascending": function () {
+                me.data.sort(function (a, b) {
+                    return d3.descending(a.value || 1000000 , b.value || 1000000);
```
data-ux: What if `a.value` or `b.value` is number zero? Some values might be over 1000000, too. 

We should do an early return from the comparator instead: 
```
if(a.value == null && b.value == null) return 0;
if(a.value == null) return -1;
if(b.value == null) return 1;
return d3.descending(a.value, b.value);
```

While at it we could refactor the logic into a `nullsLast()` comparator, that we could then call in other places like so:
```
var result = nullsLast(a.value, b.value);
if(result === 0) {
    return d3.descending(a.value, b.value);
}
return result;

```

data-ux:
Looks good!

bundles/framework/divmanazer/component/Grid.js
```diff
@@ -14,6 +14,7 @@ Oskari.clazz.define('Oskari.userinterface.component.Grid',
         this.model = null;
         this._defaultLocKey = 'Grid';
         this._loc = this._getLocalization('DivManazer');
+        this.interpolate = Oskari.getMsg.bind(null, 'DivManazer');
```
ZakarFin: Does this really need both getLocalization and getMsg? Refactor Grid to use only getMsg()

ZakarFin: Both fetch the localization which is not done twice and kept in two separate variables.

bundles/framework/divmanazer/component/Grid.js
```diff
@@ -971,7 +971,7 @@ Oskari.clazz.define('Oskari.userinterface.component.Grid',
                 });
 
                 exportButton.setPrimary(true);
-                exportButton.setTitle(me._loc('Grid.export')['export']);
+                exportButton.setTitle(me._loc('Grid.export.export.export'));
```
ZakarFin: There's one too many export in this :D

PR #273, opened 2018-02-23
Feature/thematic layer symbol scaling

jheijari:
Changed scaling factor for circle symbols and fixed label positioning on legend.

### DIFF
--- a/bundles/statistics/statsgrid2016/service/ClassificationService.js
+++ b/bundles/statistics/statsgrid2016/service/ClassificationService.js
@@ -200,10 +200,18 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationService',
             if(!ranges.max) {
                 ranges.max = 4;
             }
+            
             var x = d3.scaleLinear()
                 .domain([ranges.min, ranges.max])
                 .range([iconSize.min, iconSize.max]);
-            return x(index);
+
+            // Make size an even value for rendering
+            var size = Math.floor(x(index));
+            if (size % 2 !== 0) {
+                size += 1;
+            }
+
+            return size;
         },
         _getPointsLegend: function(ranges, opts, color, counter, statsOpts){
             var me = this;
@@ -284,10 +292,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationService',
             var legendValuesPosition = function(size, index) {
                 var step = (maxSize - minSize/2)/(ranges.length-1);
                 var y = (ranges.length - index-1) * step;
-                if(y == 0) {
-                    y=1;
-                }
-                y += fontSize;
+                y += fontSize + 2;
                 return {
                     x1: maxSize/2,
                     x2: maxSize + 10,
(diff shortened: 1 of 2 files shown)


bundles/statistics/statsgrid2016/service/ClassificationService.js
```diff
@@ -200,10 +200,10 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationService',
             if(!ranges.max) {
                 ranges.max = 4;
             }
-            var x = d3.scaleSqrt()
+            var x = d3.scaleLinear()
```
data-ux: We should continue using `scaleSqrt()` because the values should be mapped to the area of the circles. See: [Interactive Data Visualization for the Web](https://books.google.fi/books?id=NmwvDwAAQBAJ&pg=SA6-PA52&lpg=SA6-PA52&dq=visualization+area+of+circle+scalesqrt&source=bl&ots=D2WSYvvxHk&sig=Mu3RcVs4A9J6C6Gn22-UUqxKR-E&hl=en&sa=X&ved=0ahUKEwj2hMfhyMPZAhVSJlAKHeV9Ac8Q6AEIOTAC)

If we use `scaleLinear()` we are mapping the values to radius of the circle, which is perceptually wrong.

bundles/statistics/statsgrid2016/service/ClassificationService.js
```diff
@@ -358,7 +361,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationService',
 
                 textSvgEl.find('text').attr({
                     x: valPos.x2 + 4,
-                    y: valPos.y2 + fontSize/2
+                    y: valPos.y2 + classificationLabelCorrection(index)
```
data-ux: SVG `<text>` elements have `dy` attribute that we can use for adjusting the height of the text relative to some "normal" position.

jheijari: Elements are not nested here, so the "normal" is 0,0. Using dx & dy would be a better solution, but requires more refactoring so we are leaving it out of the scope by now.

bundles/statistics/statsgrid2016/service/ClassificationService.js
jheijari: The task was to implement scaling similar to the example 5.cg.5 in https://www.e-education.psu.edu/geog486/node/1869.  Mapping the values to area is generally a good idea, but here we are just presenting the classifications instead of visualizing the actual values on the map.

data-ux: Ah, was this the reason the circles were different sizes in the legend and on the map? If so, we should use `scaleSqrt()` in both map and legend.

bundles/statistics/statsgrid2016/service/ClassificationService.js
data-ux: Ok.

PR #285, opened 2018-03-05
Add view uuid into map link

data-ux:
Needed for multi-projection support.

Error popup if current view srs is not found in system default views.


bundles/mapping/toolbar/default-buttons.js
```diff
@@ -21,6 +21,26 @@ Oskari.clazz.category(
...
+         * 
+         * @return {String} current view uuid, if public view. Otherwise returns uuid of system default view with same srs as current view.
+         */
+        _getLinkUuid: function () {
+            if(Oskari.app.isPublic()) {
+                return Oskari.app.getUuid();
+            }
+            // not public -> get the a system appsetup with matching srs
+            var srs = Oskari.getSandbox().getMap().getSrsName();
+            var matchingSystemAppsetup = Oskari.app.getSystemDefaultViews().find(function(appsetup) {
```
ZakarFin: If I interpret this correctly IE doesn't support Array.find() https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

ZakarFin:
Check if we can use find. Otherwise good for merging.

PR #287, opened 2018-03-07
Projection change in my data view

nxholm:
- A dialog to inform that the views projection isn't matching the maps -> clicking ok loads target projection to map

### DIFF
--- a/bundles/framework/personaldata/PublishedMapsTab.js
+++ b/bundles/framework/personaldata/PublishedMapsTab.js
@@ -275,7 +275,52 @@ Oskari.clazz.define('Oskari.mapframework.bundle.personaldata.PublishedMapsTab',
             }
             return gridModel;
         },
+        _isCurrentProjectionSupportedForView: function (data) {
+            var currentProjection = Oskari.getSandbox().getMap().getSrsName();
+            var viewProjection = data.state.mapfull.state.srs;
+            if ( viewProjection !== currentProjection ) {
+                return false;
+            }
+            return false;
+        },
+        _loadProjectionUuid: function (srs) {
+            var views = Oskari.app.getSystemDefaultViews();
+            var uuid;
+            views.forEach( function (view) {
+                if ( view.srsName === srs ) {
+                    uuid = view.uuid;
+                    return;
+                }
+            });
+            var url = window.location.origin;
+            if (window.location.pathname && window.location.pathname.length) {
+                url += window.location.pathname;
+            }
+            url += "?uuid="+uuid;
 
+            window.location.href = url;
+        },
+        /**
+         * @method createProjectionChangeDialog
+         * Creates Oskari.userinterface.component.Popup for infoing about projection mismatch between map / published view
+         *
+         * @param function cb - callback to call when clicking ok button
+          */
+        createProjectionChangeDialog: function (cb) {
+            var me = this;
+            var dialog = Oskari.clazz.create('Oskari.userinterface.component.Popup');
+            var btn = dialog.createCloseButton("OK");
+            var cancel = dialog.createCloseButton("Cancel");
+            cancel.setHandler( function () {
+                dialog.close(true);
+            });
+            btn.addClass('primary');
+            btn.setHandler( function () {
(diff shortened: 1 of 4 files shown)


bundles/framework/personaldata/PublishedMapsTab.js
```diff
@@ -275,7 +275,55 @@ Oskari.clazz.define('Oskari.mapframework.bundle.personaldata.PublishedMapsTab',
...
             return gridModel;
         },
-
+        _isCurrentProjectionSupportedForView: function (data) {
+            var currentProjection = Oskari.getSandbox().getMap().getSrsName();
+            var viewProjection = data.state.mapfull.state.srs;
+            if ( viewProjection !== currentProjection ) {
+                return false;
+            }
+            return false;
```
ZakarFin: You could just return viewProjection === currentProjection and get rid of 3 lines of code :)

bundles/framework/personaldata/PublishedMapsTab.js
```diff
@@ -275,7 +275,55 @@ Oskari.clazz.define('Oskari.mapframework.bundle.personaldata.PublishedMapsTab',
...
+            });
+            var url = window.location.origin;
+            if (window.location.pathname && window.location.pathname.length) {
+                url += window.location.pathname;
+            }
+            url += "?uuid="+uuid;
+            return url;
+        },
+        _loadProjectionUuid: function (url) {
+            window.location.href = url;
```
ZakarFin: This is not used anywhere?

ZakarFin: Oh, it is. Maybe you should just wrap this to a function when used or at least move the implementation closer to where it's used.

bundles/framework/personaldata/PublishedMapsTab.js
```diff
@@ -353,8 +401,24 @@ Oskari.clazz.define('Oskari.mapframework.bundle.personaldata.PublishedMapsTab',
...
+                        window.open(
+                            url(),
+                            'Published',
+                            'location=1,status=1,scrollbars=yes,width=850,height=800'
+                        );
+                    }
+
+                    var supported = me._isCurrentProjectionSupportedForView( data );
+                    if (!supported) {
+                        me.createProjectionChangeDialog( cb.bind(me, me.constructUrlUuid.bind(this, srs) ));
```
ZakarFin: What does this do? Seems kinda complex solution with lots of binds where it could be done more simply.

ZakarFin: Could be written as:

    me.createProjectionChangeDialog( function() {
	    window.open(
            me.constructUrlUuid(srs),
            'Published',
            'location=1,status=1,scrollbars=yes,width=850,height=800'
         );
    });

Which I think is much more readable and easier to understand.

bundles/framework/personaldata/PublishedMapsTab.js
```diff
@@ -407,8 +471,16 @@ Oskari.clazz.define('Oskari.mapframework.bundle.personaldata.PublishedMapsTab',
             //sending a request to publisher for editing view
             var editRenderer = function (name, data) {
                 var link = me.templateLink.clone();
+                var srs = data.state.mapfull.state.srs;
                 link.text(name);
                 link.bind('click', function () {
+
+                    var supported = me._isCurrentProjectionSupportedForView( data );
+                    if (!supported) {
+                        me.createProjectionChangeDialog( me._loadProjectionUuid.bind( this, me.constructUrlUuid(srs) ) );
```
ZakarFin: This could be written as:

    me.createProjectionChangeDialog( function() {
        window.location.href = me.constructUrlUuid(srs);
     });

Which again would be easier to understand and you wouldn't need to browse the file to find the single-line implementation of _loadProjectionUuid()

bundles/framework/personaldata/PublishedMapsTab.js
ZakarFin: See the comment below. I would remove this function.

PR #289, opened 2018-03-08
Projection change - check is layer supported on layer add & removal

nxholm:
- Fixed a bug with opening the flyout after closing it
- Added eventlisteners to the card component to listen to layerevents

### DIFF
--- a/bundles/asdi/asdi-projection-change/component/card.js
+++ b/bundles/asdi/asdi-projection-change/component/card.js
@@ -6,7 +6,7 @@ Oskari.clazz.define('Oskari.projection.change.component.card',
  */
 function (view, callback) {
     this.card = _.template('<div class="projection-card" data-srs="${srs}">'+
-                        '<div class="card-image"></div> '+
+                        '<img class="card-image" src="/Oskari/bundles/asdi/asdi-projection-change/resources/images/${img}"></img> '+
                         '<div class="info-row">'+
                             '<p class="card-header"> ${projectionName} </p>'+
                             '<div class="projection-info icon-info"></div>'+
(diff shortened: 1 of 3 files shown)


bundles/asdi/asdi-projection-change/component/card.js
```diff
@@ -21,6 +21,24 @@ function (view, callback) {
     this.errorListing = Oskari.clazz.create('Oskari.projection.change.view.ErrorListing');
     this.create(view);
 }, {
+    eventHandlers: {
+        AfterMapLayerRemoveEvent: function (event) {
+            this.checkUnsupportedLayers();
```
ZakarFin: This results in a new click handler whenever a layer is added to map -> leak

bundles/asdi/asdi-projection-change/component/card.js
```diff
@@ -21,6 +21,24 @@ function (view, callback) {
...
+        },
+        AfterMapLayerAddEvent: function (event) {
+            this.checkUnsupportedLayers();
```
ZakarFin: This results in a new click handler whenever a layer is removed from map -> leak

bundles/asdi/asdi-projection-change/component/card.js
```diff
@@ -21,48 +21,79 @@ function (view, callback) {
...
-            //warningLink
-            warningElement.on('click', function ( event ) {
-                event.stopPropagation();
-                me.errorListing.show( jQuery(this), unsupportedLayers );
-            });
+        return unsupportedLayers;
+    },
+    toggleWarningElement: function () {
+        if ( this.getUnsupportedLayers().length !== 0 ) {
+            this.getElement().find('.projection-warning').removeClass('oskari-hidden');
```
ZakarFin: This doesn't toggle. Only removes the class which makes the warning visible. Nothing hides the warning once it's shown.

bundles/asdi/asdi-projection-change/component/card.js
```diff
@@ -58,9 +58,14 @@ function (view, callback) {
         return unsupportedLayers;
     },
     toggleWarningElement: function () {
+        var element = this.getElement().find('.projection-warning');
         if ( this.getUnsupportedLayers().length !== 0 ) {
-            this.getElement().find('.projection-warning').removeClass('oskari-hidden');
-        }     
+            element.removeClass('oskari-hidden');
+        } else {
+            if ( !element.hasClass('.oskari-hidden') ) {
```
ZakarFin: Add class does this internally. No need to check if class already exists.

PR #295, opened 2018-03-15
Fix metadata data quality parsing

okauppinen:
Metadata backend implementation was changed and frontend didn't work correctly with the changes. Now there is minor new changes in the backend implementation and frontend is fixed to work with new and previosly made changes. 

backend changes: https://github.com/oskariorg/oskari-server/pull/166


bundles/catalogue/metadataflyout/view/MetadataPanel.js
```diff
@@ -603,14 +619,14 @@ Oskari.clazz.define('Oskari.catalogue.bundle.metadataflyout.view.MetadataPanel',
                 }
             }
         },
-
+        //DO we need this??
```
ZakarFin: This isn't required anymore since you have removed the calls to preprocessModel functions and inject the UIlabels here:
https://github.com/okauppinen/oskari/blob/27b02f60beded3ed65c58e65014d9aa8d890b2c2/bundles/catalogue/metadataflyout/view/MetadataPage.js#L157-L159

So this can be removed.

PR #296, opened 2018-03-15
Statsgrid classification update

nxholm:
- Change classification element into a plugin
- Always visible when thematic maps are open
- Updated ui


ZakarFin:
Changing the map/browser window size results in an error:

    logger.js:22 Sandbox: Error notifying MainMapModuleClassificationToolPlugin about MapSizeChangedEvent e {_width: 1710, _height: 943} TypeError: Cannot read property 'find' of null
       at e._calculatePluginSize (ClassificationToolPlugin.js:191)
       at e.MapSizeChangedEvent (ClassificationToolPlugin.js:182)
       at e.onEvent (AbstractMapModulePlugin.js:389)
       at e.notifyAll (sandbox.js:431)
       at e.updateSize (AbstractMapModule.js:845)
       at e.updateSize (instance.js:620)

ZakarFin:
At least in a minified version the edit classification icon isn't shown at all. Meaning it only ever shows the default classification and the user can't edit this.

ZakarFin:
* Changing indicator while classification is open changes the "open classification" icon from "open" to "closed" state while the classification panel itself remains open. 
* Sometimes the "data couldn't be classified" message is shown under the legend graphic. Only one or the other should ever be shown.

bundles/statistics/statsgrid2016/components/Legend.js
```diff
@@ -58,63 +50,105 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function(sandbox, loca
...
+        var container = this._element;
         var accordion = this._accordion;
-        // cleanup previous UI
         // NOTE! detach classification before re-render to keep eventhandlers
         this.editClassification.getElement().detach();
         accordion.removeAllPanels();
-        container.empty();
-        if(el) {
+
+        if ( container.find('.legend-noactive') ) {
```
ZakarFin: Find will always return an object so this will always be true. If you want to check if something was actually hit by the selector you need to check length of the returned object. Then again calling remove without the if-statement since you are not storing the result makes the performance worse.

bundles/statistics/statsgrid2016/components/Legend.js
```diff
@@ -58,63 +50,105 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function(sandbox, loca
...
+            container.find('.legend-noactive').remove();
+        }
+
+        for (var i = 0; i < container.children().length; i++) {
+            jQuery( container.children()[i] ).empty();
+        }
```
ZakarFin: Is there some reason that container.empty() cannot be used and instead you opt to clean up by looping through everything that it holds?

bundles/statistics/statsgrid2016/components/Legend.js
```diff
@@ -1,21 +1,25 @@
-Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function(sandbox, locale) {
+Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function(sandbox) {
     this.sb = sandbox;
-    this.locale = locale;
+    this.locale = Oskari.getMsg.bind(null, 'StatsGrid');
```
ZakarFin: Passing the locale better than assuming that locale has been registered to a global from testing point of view at least.

PR #297, opened 2018-03-16
Fixed inspire tab showing

kuosman:
This PR fix admin-layerselector inspire tab showing.


ZakarFin:
Fixes an issue caused by #270

PR #304, opened 2018-03-20
Printout - Manage scale

nxholm:
- hide scale option if projectionunits != 'm'


ZakarFin:
The scaleline isn't implemented for non-metric projections on the server side so hiding the option from users on such cases.

PR #309, opened 2018-03-22
Myplaces formatting

ZakarFin:
Fixed formatting errors that ESLint reported. Also a couple of early returns added where I thought they were missing. Mostly all for(var i=0...) and for(var x in y) loops changed to forEach-loops which got some "closureMagic" functions removed.

Anyways, this is supposed to be just formatting, but with forEach() changes there might be some "this" scope issues hiding in code branches that I didn't encounter while testing.

### DIFF
--- a/bundles/framework/myplaces3/ButtonHandler.js
+++ b/bundles/framework/myplaces3/ButtonHandler.js
@@ -250,12 +250,11 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.ButtonHandler',
             if (!this.dialog || !this.drawMode) {
                 return;
             }
-            var areaDialogContent = this.loc('tools.' + this.drawMode + '.next');
             var content = this.dialog.getJqueryContent();
-            if (content.find('div.infoText') !== areaDialogContent) {
-                content.find('div.infoText').html(areaDialogContent);
-                this.dialog.moveTo('#toolbar div.toolrow[tbgroup=default-myplaces]', 'top');
-            }
+            // after the first geometry change the popup text to instruct the user that another geometry can be added to the same feature
+            content.find('div.infoText').html(this.loc('tools.' + this.drawMode + '.next'));
+            // as the popup size probably changes with text change -> move it so it's still pointing at the button
+            this.dialog.moveTo('#toolbar div.toolrow[tbgroup=default-myplaces]', 'top');
         },
 
         stop: function () {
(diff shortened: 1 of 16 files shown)


bundles/framework/myplaces3/ButtonHandler.js
```diff
@@ -255,13 +247,14 @@ Oskari.clazz.define("Oskari.mapframework.bundle.myplaces3.ButtonHandler",
...
-                if (content.find('div.infoText') !== areaDialogContent) {
-                    content.find('div.infoText').html(areaDialogContent);
-                    this.dialog.moveTo('#toolbar div.toolrow[tbgroup=default-myplaces]', 'top');
-                }
+            if (!this.dialog || !this.drawMode) {
+                return;
+            }
+            var areaDialogContent = this.loc('tools.' + this.drawMode + '.next');
+            var content = this.dialog.getJqueryContent();
+            if (content.find('div.infoText') !== areaDialogContent) {
```
ZakarFin: This if doesn't actually make sense since we are comparing text string to a jQuery object.

bundles/framework/myplaces3/ButtonHandler.js
```diff
@@ -307,45 +301,43 @@ Oskari.clazz.define("Oskari.mapframework.bundle.myplaces3.ButtonHandler",
...
+                    this.featureAdded(event.getData());
+                    return;
+                }
+                // handle Finished drawing -> main view
+                // set ignore so we don't cancel our drawing unintentionally
+                this.ignoreEvents = true;
+                // ask toolbar to select default tool
+                this.instance.sandbox.request(this, Oskari.requestBuilder('Toolbar.SelectToolButtonRequest')());
+                // disable ignore to act normally after ^request
+                this.ignoreEvents = false;
```
ZakarFin: This assumes requests are synchronous so it might easily break. I think sandbox.request() might be sync, but postRequestByName is async.

bundles/framework/myplaces3/CategoryHandler.js
```diff
@@ -86,61 +86,47 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.CategoryHandler',
...
-                            };
-                            // update maplayer name if category name has changed
-                            mapLayerService.updateLayer(mapLayers[i].getId(), layerConf);
-                        }
-                    }
-                }
-                if (!found) {
+            categories.forEach(function (cat) {
+                var layer = mapLayers.find(function (layer) {
+                    return this._getMapLayerId(cat.getId()) === layer.getId();
```
ZakarFin: This -> me

PR #311, opened 2018-03-26
jQuery live and die replacements

kuosman:
This PR replaces jQuery .live/.die methods to .on/.off.

After these changes jQuery can be updated 1.9 or newer.


ZakarFin:
Nice, thanks! This makes oskariorg/oskari-server#172 unnecessary

PR #317, opened 2018-03-27
ProjectionChange - MobileUi

nxholm:
- Created a mobile mode for the projectionchange ui
- UI stays the same for now


ZakarFin:
The current changes don't need the metatag anymore. Can you update the description to match the current changeset.

PR #321, opened 2018-03-28
Fixed issue when subgroup subgroup has layers but these not visible

kuosman:
This PR fixes issue when subgroup subgroups has layers but these are not visible to no admin users


ZakarFin:
kuosman told me privately NOT to merge this yet, so pending

PR #322, opened 2018-03-28
Openlayers popup plugin pan map fix

okauppinen:
_panMapToShowPopup is used to pan map if infobox isn't entirely in the viewport (screen). If map is moving or ShowInfoBoxRequest's given position (lonlat) isn't in the current viewport, pixels for pan might be calculated wrongly. Now map pan is done only if popups position (lonlat) is in the current viewport. If you want to show popup outside current viewport, you have to make mapmoverequest first.

Also pixels calculations for pan map is fixed. Now supported ol.OverlayPositionings are calculated correctly.  Before _panMapToShowPopup presumed that overlay's positioning is always 'top-left' (point's location) -> popup is rendered right and down from location. Now if ol.OverlayPositioning isn't defined ('no-position-info') -> 'top-left' positioning is used for calculations.

### DIFF
--- a/bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol3.js
+++ b/bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol3.js
@@ -666,7 +666,7 @@ Oskari.clazz.define(
          */
         _panMapToShowPopup: function (lonlatArray, positioning, margins) {
             var margins = margins || this._viewportMargins;
-            // don't try to pan the map if gfi popup position isn't in the viewport
+            // don't try to pan the map if gfi popup position isn't in the viewport (extended with isInViewport margin)
             if (!this.getMapModule().isLonLatInViewport(lonlatArray, margins.isInViewport)){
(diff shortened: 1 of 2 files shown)


bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol3.js
```diff
@@ -658,36 +664,71 @@ Oskari.clazz.define(
...
-                } else {
-                    panx = width - (pixels.x + infoboxWidth);
+                popupX = popup.width(),
+                popupY = popup.height();
+
+            //If supported ol.OverlayPositioning is used, use it instead of default values
+            Object.keys(posClasses).forEach(function (pos){
+                if (positioning === posClasses[pos]){
+                    positionX = positioning.split("-")[1];
+                    positionY = positioning.split("-")[0];
```
ZakarFin: Split once and take the results:

    var popupDirection = positioning.split("-");
    positionY = popupDirection[0];
    positionX = popupDirection[1];

PR #325, opened 2018-03-29
Timeseries - default speed

nxholm:
- Change the frameinterval from 1000 (fast) to 2000 (normal)


ZakarFin:
Does this change make the "normal" speed selected in the UI or does this just slow down each of the speed options?

ZakarFin:
Looks like it does 👍

PR #327, opened 2018-04-03
Occasionally adding a new feature would not work with content editor

heikkiy:
Fixed a bug where occasionally on at least IE clicking the add feature button in content editor would not work. In the browser console the complaint was 'Object doesn't support property or method startsWith'. Changed startsWith with indexOf.


ZakarFin:
Thanks, String.startsWith() isn't supported by IE so makes sense :)

PR #328, opened 2018-04-03
Layerselector & Projectionchange - Warning icon update

nxholm:
- Moved the warning icon from projectionchange bundle to oskari resources
- Layerselector projection warning uses the same icon now


ZakarFin:
Rename the icon files to icon-warning-[light|dark].png for consistency (like icon-info.png)

bundles/framework/layerselector2/view/Layer.js
```diff
@@ -20,7 +20,7 @@ Oskari.clazz.define("Oskari.mapframework.bundle.layerselector2.view.Layer",
     }, {
         __template: '<div class="layer"><input type="checkbox" /> ' +
                     '<div class="layer-tools">'+
-                    '   <div class="layer-not-supported backendstatus-maintenance-pending" title="" ></div>' +
```
ZakarFin: backendstatus-maintenance-pending is meant to signal that a service will be down for maintenance. I don't think it's used anywhere so it's ok to remove.

PR #330, opened 2018-04-04
Featuredata drawing event fix

okauppinen:
Published maps won't have map selection plugin at least always so drawing event sent by measuretool  caused error.

Added check that if map selection plugin doesn't exist, featuredata doesn't handle drawing events.


ZakarFin:
Nice catch! Published maps never have the featuredata selectiontool so I guess this has caused problems we haven't noticed for a while now. Basically whenever someone is doing measurements (or other draw operations) on a published map that include the featuredata table this has caused JS errors.

PR #331, opened 2018-04-05
add layer opacity to ol3/4 wmts layer

okauppinen:


ZakarFin:
Sets the opacity on the map to match the layer "state". Without this fix the opacity slider on selected layers (and other funtionalities like publishing a map or printing) showed the layer state correctly, but the layer on the map was always at 100% opacity.

PR #335, opened 2018-04-09
Resetting state after removing stats layer from map.

jheijari:


ZakarFin:
Removes all indicators and clears state from the statistical maps if the user removes the statistics layer from the map (or if it’s removed programmatically)

PR #337, opened 2018-04-10
Statsgrid - Classification transparency fix

nxholm:
- Now the whole classification ui is transparent if user has so selected in the publisher


ZakarFin:
Is !important really needed in styles for this?

PR #342, opened 2018-04-13
Statsgrid - Classification reverse colors fix

nxholm:
- Fix an issue with the reverse colors checkbox not working.
- Also fixes an minor issue with incrementally adding on click listeners to the checkbox on every setColorValue call


bundles/framework/divmanazer/component/ColorSelect.js
```diff
@@ -307,10 +307,6 @@ Oskari.clazz.define('Oskari.userinterface.component.ColorSelect',
 
             me._setHandlers();
 
-            jQuery(document).bind('click', function(){
```
ZakarFin: Should this still be part of the component, but just do it once instead of every time colors are set? I think the intent here is that the dropdown is closed if the user clicks outside the selection.

PR #343, opened 2018-04-16
Layerselector2 - Filtering changes

nxholm:
- Create only one default filter (newest)
- Other filters are checked when layers are loaded -> creates filters based on layers available (currently featuredata & timeseries)


bundles/framework/layerselector2/Flyout.js
```diff
@@ -157,7 +167,21 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
                 },
                 'featuredata');
         },
+        /**
+         * Add timerseries filter.
+         * @method  @public addNewestFilter
+         */
+        addTimeseriesFilter: function() {
```
ZakarFin: This should be in the timeseries bundle.

bundles/framework/layerselector2/Flyout.js
```diff
@@ -117,15 +122,20 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
...
-            // Add featuredata filter
-            me.addFeaturedataFilter();
+            this.addNewestFilter();
+        },
+        /**
+         * Adds extra filter buttons if layers exists
+         * @method  @private displayFilters
+         */
+        displayExtraFilters() {
+            if ( this.mapLayerService.getFilteredLayers('featuredata').length !== 0) {
```
ZakarFin: You shouldn't hardcode the filters here. Instead have a function in maplayerservice that returns the "active" filters (returns only the ones that actually hit layers when filtering). It should loop through all the registered filters instead of hardcoded names as filters can be added externally. Layerselector should call the function again and re-render the filters on MapLayerEvent with operation add/remove as they might change the active filters (more layers have been registered to the system or removed).

bundles/framework/layerselector2/Flyout.js
ZakarFin: A better place for the similar addFeaturedataFilter() should be on the mapwfs bundle. Maybe the modelbuilder class in it or the plugin.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -525,6 +529,7 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
                 this.getSandbox().notifyAll(Oskari.eventBuilder('MapLayerEvent')(null, 'add'));
 
                 if (typeof callbackSuccess === 'function') {
+                    me.trigger("Layers.Loaded", true);
```
ZakarFin: Why? It already sends out a MapLayerEvent when this happens. In any case the event shouldn't be tied on having the callback function sent or not.

bundles/framework/layerselector2/Flyout.js
```diff
@@ -26,7 +26,12 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
 
         this.mapLayerService = Oskari.getSandbox().getService('Oskari.mapframework.service.MapLayerService');
         this.layerlistService = Oskari.getSandbox().getService('Oskari.mapframework.service.LayerlistService');
-
+        
+        this.mapLayerService.on('Layers.Loaded', function (loaded) {
```
ZakarFin: Remove this event and use MapLayerEvent instead. Also this is only sent when the bulk loading is completed. Currently it doesn't react to runtime changes on the layers. Using MapLayerEvent will solve that as well (if layer id is null -> it's a massive update like this and when layer id is present only the layer referenced changed).

nxholm:
Closed due to fixes done in new branch.

PR #344, opened 2018-04-16
Map state - add layer - unsupported layer

nxholm:
- Create a check that if a layer is unsupported and has timeseries, don't add it to the map
- Move the warning message from the event to the function calling the event

### DIFF
--- a/bundles/mapping/mapmodule/AbstractMapModule.js
+++ b/bundles/mapping/mapmodule/AbstractMapModule.js
@@ -2250,10 +2250,6 @@ Oskari.clazz.define(
                 layerFunctions = [],
                 sandbox = this.getSandbox();
 
-            if (!layer.isSupported(sandbox.getMap().getSrsName())) {
-                this._mapLayerService.showUnsupportedPopup();
-            }
-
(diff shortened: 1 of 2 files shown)


ZakarFin:
This will lose the maplayer state when for example changing between projections. The original problem is that timeseries ui will become visible even if the layer is not visible on the map. This change has much wider effect than just fixing the problem and I don't think it should work like this. A better solution would be that timeseries bundle detects if the layer is visible to the user instead of assuming it is.

ZakarFin:
To be clear, in my opinion this shouldn't be merged. Can the original problem of showing the UI control when layer is not visible/supported be fixed by just changing the line:

    var srsName = me._sandbox.getMap().getSrsName();
    if (!layers[i].hasTimeseries() || !layer.isVisible() || !layer.isSupported( srsName )) { ... }

 in https://github.com/oskariorg/oskari-frontend/blob/1.46.0/bundles/framework/timeseries/service/TimeseriesLayerService.js#L125 ?

nxholm:
I'll close it since the impact is too heavy

PR #345, opened 2018-04-16
Timeseries - handle layer if it's visible & supported

nxholm:


nxholm:
wrong base branch for hotfix

PR #351, opened 2018-04-19
Hierarchical layerlist fixes

kuosman:
Fixes for hierarchical layerlist:
- not use json keys anymore (now uses functions)
- fixed not backend loaded layers visibility (myplaces)

Fixes for MapLayerService:
- fixed getFilteredLayerGroups function to check at group layers belongs to filtered layers


ZakarFin:
Thanks!

PR #352, opened 2018-04-19
Layerselector2 - Filter layers

nxholm:
- Add new filter - Timeseries
- Display only registered filters that have layers


nxholm:
Active icon is still missing, so we'll have to wait for that to be able to merge

bundles/framework/layerselector2/Flyout.js
```diff
@@ -19,18 +19,24 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
...
         this.mapLayerService = Oskari.getSandbox().getService('Oskari.mapframework.service.MapLayerService');
         this.layerlistService = Oskari.getSandbox().getService('Oskari.mapframework.service.LayerlistService');
-
+        this.filterComponent = Oskari.clazz.create("Oskari.layerselector2.view.FilterButtons");
         this.addedButtons = {};
 
         this.layerlistService.on('Layerlist.Filter.Button.Add', function(button) {
-            me.addFilterTool(button.properties.text, button.properties.tooltip, button.properties.cls.active, button.properties.cls.deactive, button.filterId);
+            var filters = me.mapLayerService.getActiveFilters();
+            if ( filters.includes(button.filterId) || button.filterId === "newest" ) {
```
ZakarFin: Array.includes() doesn't work on IE so you'll have to replace it or add a polyfill for it in the core. The newest filter shouldn't require any special handling here so you should remove the handling or describe why it's needed.

bundles/framework/layerselector2/view/FilterButtons.js
```diff
@@ -0,0 +1,175 @@
+Oskari.clazz.define("Oskari.layerselector2.view.FilterButtons",
+    function (layertabs) {
```
ZakarFin: layertabs is sent in the constructor but not saved anywhere?

bundles/framework/layerselector2/view/FilterButtons.js
```diff
@@ -0,0 +1,175 @@
...
+          /**
+         * Set filter button tooltip
+         * @method  @private _setFilterTooltip
+         * @param {String} filterName filter name
+         * @param {String} tooltip    tooltip
+         */
+        _setFilterTooltip: function(filterName, tooltip) {
+            var me = this;
+            me.layerTabs.forEach(function(tab) {
+                var filterContainer = tab.getTabPanel().getContainer().find('.layerselector2-layer-filter');
```
ZakarFin: You should pass the element for the buttons in constructor instead of this component needing very detailed information about the structure of the parent component.

bundles/framework/layerselector2/Flyout.js
```diff
@@ -104,28 +99,40 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
...
-            me.addFeaturedataFilter();
+                filterButton.on('FilterActivate', function (currentFilter) {
+                    me._currentFilter = currentFilter;
+                    me.populateLayers();
+                });
+                
+            });
         },
+        setActiveFilter: function () {
+            var me = this;
```
ZakarFin: Shouldn't this take me._currentFilter as parameter instead of having a "set" method without any parameters?

bundles/framework/layerselector2/Flyout.js
```diff
@@ -255,7 +244,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.layerselector2.Flyout',
             // Add filter tab change listener
             me.tabContainer.addTabChangeListener(function(previousTab, newTab) {
                 if (me._currentFilter) {
-                    me.activateFilter(me._currentFilter);
+                    me.setActiveFilter();
```
ZakarFin: me.activateFilter(me._currentFilter); was better than calling a set() without parameters :)

bundles/framework/layerselector2/view/FilterButtons.js
```diff
@@ -0,0 +1,180 @@
+Oskari.clazz.define("Oskari.layerselector2.view.FilterButtons",
+    function ( element ) {
+        var me = this;
+        this.mapLayerService = Oskari.getSandbox().getService('Oskari.mapframework.service.MapLayerService');
```
ZakarFin: MapLayerService isn't used in the file. Remove the reference.

bundles/framework/layerselector2/view/FilterButtons.js
```diff
@@ -0,0 +1,180 @@
...
+        this.layerlistService = Oskari.getSandbox().getService('Oskari.mapframework.service.LayerlistService');
```
ZakarFin: You should pass layerlistService as parameter so the component can function without fetching an external reference and it's more obvious that it's a requirement for using the component.

bundles/framework/layerselector2/view/FilterButtons.js
```diff
@@ -0,0 +1,180 @@
...
+            if (!notDeactivateThisFilter) {
+                me.activateFilter();
+            }
+            this.trigger('FilterActivate', this._currentFilter);
+        },
+        render: function() {
+            var me = this;
+            this.rootElement.empty();
+            this.filterButtons.forEach( function ( button ) {
+                if ( _.include(me.renderableButtons, button.name) ) {
```
ZakarFin: As this is the only time you need lodash/underscore for the component I would just replace this with:

   if(me.renderableButtons.findIndex(button.name) !== -1) { ... }

Or if renderableButtons is an object with Object.keys() to get rid of the extra dependency.

PR #355, opened 2018-04-23
Hierarchical layerlist search fix

kuosman:
Some fixes of hierarchical layerlist search:
- no result text showed now if no layers founded
- only shows founded layers groups (not empty groups)
- moved ugly timeout different place (tryied remove but tools not rendered correctly so keep it there a while)

### DIFF
--- a/bundles/framework/hierarchical-layerlist/view/LayerGroupTab.js
+++ b/bundles/framework/hierarchical-layerlist/view/LayerGroupTab.js
@@ -986,7 +986,7 @@ Oskari.clazz.define(
 
             me.tabPanel.getContainer().append(layerTree);
 
-            if(jQuery('.hierarchical-layerlist-search-noresults').length === 0) {
+            if(me.tabPanel.getContainer().find('.hierarchical-layerlist-search-noresults').length === 0) {
                 var noSearchResults = me.templates.noSearchResults.clone();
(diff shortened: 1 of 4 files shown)


bundles/framework/hierarchical-layerlist/view/LayerGroupTab.js
```diff
@@ -980,6 +985,14 @@ Oskari.clazz.define(
             };
 
             me.tabPanel.getContainer().append(layerTree);
+
+            if(jQuery('.hierarchical-layerlist-search-noresults').length === 0) {
```
ZakarFin: This searches from the whole document. Please change it to find under the flyout/tabpanel.

bundles/framework/hierarchical-layerlist/view/LayerGroupTab.js
```diff
@@ -1058,6 +1071,25 @@ Oskari.clazz.define(
...
+            });
+        },
+
+        /**
+         * Set no result message visible
+         * @method setNoResultMessageVisible
+         * @param  {Boolean}                  visible has message visible
+         */
+        setNoResultMessageVisible: function(visible) {
+            var noResultsElement = jQuery('.hierarchical-layerlist-search-noresults');
```
ZakarFin: Same here, don't search the whole document. Just from where you expect to find it.

PR #356, opened 2018-04-23
Single click handing fixes

kuosman:
Fixed group click handling so at groups click not checks groups (and their subgroups and layers). 
Click only toggle node open/close.


ZakarFin:
Fix for hierarchical layerselector so clicking on group name don't select all the layers in the group. Clicking on the checkbox of the group is used for that.

PR #362, opened 2018-04-25
Statsgrid -  select multiple indicators

nxholm:
- Indicator selector was changed to allow selecting multiple indicators.


bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -66,6 +74,21 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
       });
       return el;
     },
+    /**@method makeMultiChosen
+    *  applies jQuery chosen to specidied element
+    * @param {element} el
+     */
+    makeMultiChosen: function( el, options ) {
+      el.find( 'select' ).attr('multiple', true);
```
ZakarFin: Why not just move the options.multi logic to makeChosen() function? This is basically a copy pasted version of the method with one line difference and 1 to 5 lines on the caller side :)

ZakarFin: Remove this function, make all the callers use makeChosen() as they did before and add this at the beginning of makeChosen() :

    if(options.multi) {
        el.find( 'select' ).attr('multiple', true);
    }

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -25,7 +34,53 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
...
         this.container = null;
     },
+    _createUi: function ( datasrc, indId, selections, regionsets, values) {
+        var me = this;
+        var locale = me.instance.getLocalization();
+        var errorService = me.service.getErrorService();
+        var panelLoc = locale.panels.newSearch;
+
+        var cont = jQuery(this.__templates.main());
+        this._anchorEl.append(cont);
```
ZakarFin: So _anchorEl needs to be set (by indicatorSelected()?) before calling _createUi(), but nothing in this file calls indicatorSelected()? Seems a bit too fragile.

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -25,112 +34,86 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
...
-            });
+        me.trigger('indicator.changed', regionsets.length > 0);
+    },
+    /**
+     * @method  @public indicatorSelected  handle indicator selected
+     * @param  {Integer} datasrc indicator datasource
+     * @param  {String} indId    indicator id
+     * @param  {Object} elements elements
+     */
+    indicatorSelected: function ( datasrc, indId, elements ) {
```
ZakarFin: As indicatorSelected is the "go to" function in this component I would move it up/ move _createUi() down in the file. So basically indicatorSelected is called externally, which calls IndicatorParameterHandler. The IndicatorParameterHandler triggers an event that triggers a call to _createUi(). Right?

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -3,9 +3,18 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
     this.sb = sandbox;
```
ZakarFin: Looks like instance is only used for getLocalization() calls in this component? Maybe just send the localization as parameter instead of the instance to make it more testable.

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -0,0 +1,152 @@
...
+        var me = this;
+        var panelLoc = this.locale.panels.newSearch;
+        var allSelectors = [];
+        var regionsSets = [];
+        var deferredArray = [];
+
+        indicators.forEach( function (indicatorId) {
+            var deferred = new jQuery.Deferred();
+            me.service.getIndicatorMetadata(datasrc, indicatorId, function (err, indicator) {
+                indicator.selectors.forEach( function (selector) {
```
ZakarFin: This is missing any error handling like:

    if(err) {
        deferred.fail(err)?
        return;
    }

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -0,0 +1,152 @@
...
+                selectors: selections,
+                regionSet: indicator.regionsets,
+                values: combinedValues
+            }
+
+            me.trigger('Data.Loaded', data);
+        });
+    },
+    handleMultipleIndicators: function (datasrc, indicators, elements) {
+        indicators = indicators.filter( function (n) { return n != "" } );
```
ZakarFin: In my opinion this could be implemented in a more maintainable and simpler code if handleSingleIndicator() took a callback as parameter, handleMultipleIndicators() would call it in a loop and do the "extra stuff" after all the data has been fetched based on the values object for the single indicators.

ZakarFin: Please refactor towards this ^

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -0,0 +1,133 @@
...
+                } else {
+                    regionsets =  regionsets.concat(value.regionset);
+                }
+
+                Object.keys(value.values).forEach( function(key) {
+                    if ( !combinedValues[key] ) {
+                        combinedValues[key] = value.values[key];
+                    } else {
+                        value.values[key].forEach( function (val) {
+                            var inArray =  combinedValues[key].some( function (obj) {
```
ZakarFin: array.some() hasn't been polyfilled.

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -0,0 +1,133 @@
...
+
+            var data = {
+                datasrc: me.datasource,
+                indicators: me.indicators,
+                selectors: selections,
+                regionset: indicator.regionsets,
+                values: combinedValues
+            }
+            cb(data);
+            //me.trigger('Data.Loaded', data);
```
ZakarFin: With single indicator the trigger isn't sent at all -> the logic is broken. Either do a:

    if (typeof cb === 'function') {
        cb(data);
    } else {
        me.trigger('Data.Loaded', data);
    }

Or move the trigger('Data.Loaded') to the constructor calling this function.

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
ZakarFin: Nevermind, looks like it's included by all the browsers we support :)

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -0,0 +1,122 @@
...
+                cb(data);
+            } else {
+                me.trigger('Data.Loaded', data);
+            }
+        });
+    },
+    handleMultipleIndicators: function (datasrc, indicators, elements) {
+        indicators = indicators.filter( function (n) { return n != "" } );
+        var me = this;
+        var panelLoc = this.locale.panels.newSearch;
```
ZakarFin: Remove panelLoc as it isn't used in this method.

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -1,15 +1,28 @@
...
     this._selections = [];
+    this.parentElement = null;
     Oskari.makeObservable(this);
+    var me = this;
+    var errorService = this.service.getErrorService();
+
+    this.paramHandler.on('Data.Loaded', function ( data ) {
+        me.spinner.stop();
+        if ( Object.keys(data.selectors).length === 0 ) {
+            errorService.show(locale.erros.title, locale.errors.regionsetsIsEmpty);
```
ZakarFin: Based on the localization key the error messages says something about regionsets, but you are checking if the indicator had any selectors. Showing wrong error message here?

PR #363, opened 2018-04-25
Layer and group order fix

kuosman:
This PR also needs https://github.com/oskariorg/oskari-server/pull/202

# MaplayerGroup domain
- now uses children instead of layers. children array includes type, id ann order number objects

# MapLayerService
- updateLayer: Fixed save layer handling  (layer conf now includes layr groups id array) update also layer groups correct
- fixed to use also MaplayerGroup getChildren method

# Hierarchical layerlist
- now uses MaplayerGroup getChildren method to getted sorted lists (removed sortings from bundle)

### DIFF
--- a/bundles/mapping/mapmodule/service/map.layer.js
+++ b/bundles/mapping/mapmodule/service/map.layer.js
@@ -705,7 +705,7 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
                     layerGroups = group;
                 }
             }
-            return (id) ? layerGroups : layerGroups || this._layerGroups;
+            return (id) ? layerGroups : this._layerGroups;
         },
 
         /**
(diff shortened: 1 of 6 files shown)


bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -688,7 +705,7 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
                     layerGroups = group;
                 }
             }
-            return layerGroups || this._layerGroups;
+            return (id) ? layerGroups : layerGroups || this._layerGroups;
```
ZakarFin: Reading the method implementation if id is falsey layerGroups will always be null so you could just do 

    return (id) ? layerGroups : this._layerGroups;

PR #373, opened 2018-05-03
Stats - multi regionset

nxholm:
- A regionset select that filters available datasources & indicators

### DIFF
--- a/bundles/framework/divmanazer/component/SelectList.js
+++ b/bundles/framework/divmanazer/component/SelectList.js
@@ -43,13 +43,13 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
         var dataKey = data[i];
         var option = this._option.clone();
 
-        if ( !dataKey.id && !dataKey.title ) {
+        if ( !dataKey.id && !dataKey.title || !dataKey.name ) {
           option.val( dataKey ).text( dataKey );
         }
         if(dataKey.cls) {
           option.addClass(dataKey.cls);
         }
-        option.val(dataKey.id).text(dataKey.title);
+        option.val(dataKey.id).text(dataKey.title || dataKey.name);
         select.find('select').append(option);
 
       }
@@ -79,11 +79,27 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
     },
     resetToPlaceholder: function() {
       var chosen = this.element.find('select');
-      chosen.find('option:first-child').attr('selected', 'selected');
+      if ( chosen.attr('multiple') ) {
+        chosen.val('');
+      } else {
+        chosen.find('option:first-child').attr('selected', 'selected');
+      }
       this.update();
     },
     update: function() {
       this.element.find('select').trigger('chosen:updated');
+    },
+    getOptions: function () {
+      var chosen = this.element.find('select');
+      //filter away the placeholder
+      var options = chosen.find('option').filter( function (option) {
+        return this.value !== "";
+      });
+      var disabled = chosen.find('option:disabled');
+      return {
+        'options': options,
+        'disabled': disabled
+      }
     },
     /**  
      * @method _setEnabledImpl or disable select
@@ -122,6 +138,30 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
         });
         tobeRemoved.remove();
         chosen.trigger('chosen:updated');
+    },
+    /** @method disableOption disables an options where value mathces id
+     *   @param { Array } ids array if ids to find
+     */
+    disableOptions: function ( ids ) {
+      var me = this;
+      var chosen = this.element.find('select');
+
+      ids.forEach(function (id) {
+        var valueOption = me.element.find('select option[value=' + id + ']');
+        valueOption.attr('disabled', true);
+      });
+
+      chosen.trigger('chosen:updated');
+    },
+    reset: function () {
+      var chosen = this.element.find('select');
+
+      var state = this.getOptions();
+      for (var i = 0; i < state.disabled.length; i++) {
+        jQuery( state.disabled[i] ).attr('disabled', false)
+      }
+      this.resetToPlaceholder();
+      chosen.trigger('chosen:updated');
     },
     /** @method updateOptions
     *   updates an already defined chosen with new data
@@ -167,7 +207,6 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
       }
       return this.element.find( 'select' ).val();
     },
-
     /** @method adjustChosen
     *   adjusts the chosen direction according to the screen
     */
--- a/bundles/framework/divmanazer/component/SelectList.js
+++ b/bundles/framework/divmanazer/component/SelectList.js
@@ -3,138 +3,133 @@
  *
  * Simple select using jQuery chosen
  */
-Oskari.clazz.define('Oskari.userinterface.component.SelectList',
-
-  function( id ) {
+Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
     this.id = id;
     this._option = jQuery('<option></option>');
-    this._selectTemplate = jQuery('<div class="oskari-select">'+
-                                  '<select></select>'+
+    this._selectTemplate = jQuery('<div class="oskari-select">' +
+                                  '<select></select>' +
                                   '</div>');
     this.element = null;
-  },{
-    /**@method create
+}, {
+    /** @method create
     *  creates a select with data specified
     * @param {Object} data, needs to have the keys id and title to construct a list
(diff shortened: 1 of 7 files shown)


bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -51,7 +52,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
      * @param  {String} indId    indicator id
      * @param  {Object} elements elements
      */
-    indicatorSelected: function ( datasrc, indId, elements ) {
+    indicatorSelected: function ( datasrc, indId, regionsets, elements ) {
```
ZakarFin: As there's a bunch of code passing around regionsets this variable could be named with something like regionsetRestriction or something like that. As it's used to filter out available regionsets to just include these ones, right?

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -123,6 +135,26 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
         tobeRemoved.remove();
         chosen.trigger('chosen:updated');
     },
+    /** @method disableOption disables an options where value mathces id
+     *   @param { String } id string to find
+     */
+    disableOption: function ( id ) {
```
ZakarFin: Maybe rename this disableOptions and take an array as parameter. Then chosen doesn't need to trigger an update for each indicator (that could be hundreds of times in a row)

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -167,7 +199,9 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList',
       }
       return this.element.find( 'select' ).val();
     },
-
+    getAllOptions: function () {
```
ZakarFin: Where is this used? At least state() is not using it.

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -70,7 +71,11 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
          me.spinner.insertTo(this.parentElement.parent());
          me.spinner.start();
         //get the data to create ui with
-        me.paramHandler.getData( datasrc, indId, elements );
+        me.paramHandler.getData( datasrc, indId, regionsets, elements );
```
ZakarFin: regionsets -> regionsetRestriction here as well

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -70,7 +71,11 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
...
+    },
+    refresh: function ( datasrc, indId, regionsets ) {
```
ZakarFin: regionsets -> regionsetRestriction here as well

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -116,7 +124,19 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
 
         me.trigger('indicator.changed', regionsets.length > 0);
     },
-  
+    datasourceSelected: function (datasrc, regionsets) {
```
ZakarFin: Why is this needed? Isn't parameters affected by indicatorselection, not datasource selection?

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -116,7 +124,19 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
...
+        if (regionsets === null) {
+            return;
+        }
+        return this.paramHandler.testIndicatorsRegionsets(datasrc, regionsets );
+    },
+    regionsetSelected: function (regionsets) {
```
ZakarFin: Again, not sure why this would be needed? If it really is needed here it should probably be called something like restrictRegionsets()

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -64,7 +69,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameterHandler', fun
                 datasrc: me.datasource,
                 selectors: combinedValues,
                 indicators: me.indicators,
-                regionset: indicator.regionsets
+                regionset: me.regionsets ? me.regionsets : indicator.regionsets
```
ZakarFin: This should be indicator.regionsets that are included in me.regionsets?

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -70,7 +73,14 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
          me.spinner.insertTo(this.parentElement.parent());
          me.spinner.start();
         //get the data to create ui with
-        me.paramHandler.getData( datasrc, indId, elements );
+        me.paramHandler.getData( datasrc, indId, regionsetRestriction, elements );
+    },
+    refresh: function ( datasrc, indId, regionsetRestriction ) {
+        this.clean();
+        // only refresh data if component has been initialized
+        if ( this._initialized ) {
```
ZakarFin: RegionSelector is created on indicatorSelected() as well. Would it be enough to check if it has been created instead of using another variable just for this? Or why is this check needed? What's the difference between indicatorSelected() and refresh()?

ZakarFin: Maybe you can get rid of refresh() if you just skip the spinner lines on indicatorSelected() when the spinner is already shown on the UI?

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -8,11 +8,15 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameterHandler', fun
 
     Oskari.makeObservable(this);
 }, {
-    getData: function ( datasrc, indicators, elements ) {
+    getData: function ( datasrc, indicators, regionsets, elements ) {
```
ZakarFin: Use the same regionsetRestriction name here.

bundles/statistics/statsgrid2016/components/IndicatorParametersHandler.js
```diff
@@ -25,10 +29,11 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameterHandler', fun
         var errorService = me.service.getErrorService();
         var locale = this.locale;
         var panelLoc = locale.panels.newSearch;
+        var elements = me.elements || {};
 
         this.service.getIndicatorMetadata(this.datasource, indId, function (err, indicator) {
-            if (me.elements.dataLabelWithTooltips) {
-                me.elements.dataLabelWithTooltips.find('.tooltip').hide();
+            if (elements.dataLabelWithTooltips) {
+                elements.dataLabelWithTooltips.find('.tooltip').hide();
```
ZakarFin: What are these "elements.dataLabelWithTooltips"? Some elements that's being passed around between classes? Feels like they shouldn't be referenced in so many places.

bundles/statistics/statsgrid2016/components/IndicatorSelection.js
```diff
@@ -130,19 +130,37 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorSelection', function (
         indicatorSelector.append(indicDropdown);
         indicSelect.adjustChosen();
 
+        //Regionsets
```
ZakarFin: Rename variables to be more in line with this being a regionset restriction/filter.

bundles/statistics/statsgrid2016/components/IndicatorSelection.js
```diff
@@ -190,5 +242,54 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorSelection', function (
...
+    /**
+     * @method  @public  getUnsupportedDatasets 
+     * @description returns a list of unsupported datasources for the currently selected regionset(s)
+     * @param regionsets regionsets
+     */
+    getUnsupportedDatasetsList: function ( regionsets ) {
+        if (regionsets === null) {
+            return;
+        }
+        regionsets = regionsets.map( function (id)  { return Number(id) } );
```
ZakarFin: The same number casting is done several times. Do it just once and pass the array of numbers around.

PR #379, opened 2018-05-14
Content editor drawtools

heikkiy:
Made the content-editor bundle work with OL3. Requires turf and also EPSG:4326 definition into projectionDefs of the OL3 view so that calculating distance from clicked point into a multilinestring can be accomplished when editing a multilinestring feature.


bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -408,7 +419,6 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
             } else {
                 url = ajaxUrl + 'action_route=SaveFeature';
```
ZakarFin: Is this relying on a global variable? You should use Oskari.urls.getRoute('SaveFeature') instead.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -474,48 +477,49 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
-                    requestData.geometries.type = "multipolygon";
-                    for (var i = 0; i < geometries.components.length; i++) {
-                        var tmpPolygon = [];
-                        for (var j = 0; j < geometries.components[i].components.length; j++) {
-                            var tmpLinearString = [];
-                            for (var k = 0; k < geometries.components[i].components[j].components.length; k++) {
-                                tmpLinearString.push({x: geometries.components[i].components[j].components[k].x, y: geometries.components[i].components[j].components[k].y});
+            if (geometries != null) {
+                if(geometries.type === 'FeatureCollection') {
+                    jQuery.each(geometries.features || [], function(key, value){
```
ZakarFin: Use Array.forEach instead.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -474,48 +477,49 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
-                    }
-                } else if (geometries.id.indexOf("OpenLayers_Geometry_MultiPolygon_") == 0) {
-                    requestData.geometries.type = "multipolygon";
-                    for (var i = 0; i < geometries.components.length; i++) {
-                        var tmpPolygon = [];
-                        for (var j = 0; j < geometries.components[i].components.length; j++) {
-                            var tmpLinearString = [];
-                            for (var k = 0; k < geometries.components[i].components[j].components.length; k++) {
-                                tmpLinearString.push({x: geometries.components[i].components[j].components[k].x, y: geometries.components[i].components[j].components[k].y});
+            if (geometries != null) {
```
ZakarFin: What's going on in here? Please add some comment why this is done. A similar loop is implemented in _fillLayerGeometries() so maybe you could move this in another function like cloneGeometry() or similar that better describes what's being done.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -662,19 +668,19 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
                 colourScheme,
                 font;
 
-            $('.content-draw-tools').removeClass('hide');
+            jQuery('.content-draw-tools').removeClass('hide');
```
ZakarFin: Limit the element search to the container of this functionality. Something like me.mainPanel.find('.content-draw-tools'). Same comment for similar lines below.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -931,8 +937,8 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
         },
         _storeFormData: function () {
             var me = this;
-            $(".getinforesult_table tr input").each(function (index) {
-                me.currentData.features[0][index+1] = $(this).val();
+            jQuery(".getinforesult_table tr input").each(function (index) {
```
ZakarFin: Limit the search to inputs in this functionality with: me.mainPanel.find('.getinforesult_table tr input')

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1044,17 +1050,17 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
 
                 return wrapper;
-            }, $(me.templates.wrapper));
+            }, jQuery(me.templates.wrapper));
         },
         _getFeatureData: function () {
             var result = [];
-            $('.edit-feature .getinforesult_table').first().find('tr').each(function () {
-                var key = $(this).find('td').eq(0).html();
+            jQuery('.edit-feature .getinforesult_table').first().find('tr').each(function () {
```
ZakarFin: Same here. Don't search from the whole document.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1089,77 +1095,104 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
                     me.drawToolType = "edit";
                     me.clickedGeometryNumber = null;
-                    if (me.layerGeometries.geometry.id.indexOf("Multi") > -1) {
-                        if (me.layerGeometries.geometry.components != undefined) {
-                            for (var i = 0; i < me.layerGeometries.geometry.components.length; i++) {
-                                if (me.layerGeometries.geometry.components[i].atPoint({lon:me.clickCoords.x,lat:me.clickCoords.y}, 2, 2))
-                                {
-                                    me.clickedGeometryNumber = i;
-                                    break;
+                    var geometryType = me.layerGeometries.geometry.getType();
```
ZakarFin: Move this block in to a separate function.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1170,13 +1203,13 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
 
-            var toolContainer = $("<div />").addClass('toolrow');
+            var toolContainer = jQuery("<div />").addClass('toolrow');
             toolContainer.append(pointButton);
             toolContainer.append(lineButton);
             toolContainer.append(areaButton);
             toolContainer.append(geomEditButton);
             toolContainer.append(geomDeleteButton);
-            $('.content-draw-tools').append(toolContainer);
+            jQuery('.content-draw-tools').append(toolContainer);
```
ZakarFin: Limit the search to the main panel.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1300,21 +1333,21 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
                 var request = me.sandbox.getRequestBuilder('InfoBox.ShowInfoBoxRequest')("contentEditor", me.loc.deleteGeometryDialog.title, content, new OpenLayers.LonLat(featureGeometry.getCentroid().x, featureGeometry.getCentroid().y), options);
                 me.sandbox.request(me.getName(), request);
-                $("#delete-dialog").parent().parent().css('height', 'auto').css('padding-bottom', '30px');
+                jQuery("#delete-dialog").parent().parent().css('height', 'auto').css('padding-bottom', '30px');
             }
             me.operationMode = "edit";
         },
         _clearFeaturesList: function () {
-            $(".properties-container").empty();
+            jQuery(".properties-container").empty();
```
ZakarFin: Limit the search to the main panel.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1300,21 +1333,21 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
         },
         _formIsValid: function () {
-            if ($(".properties-container input.field-invalid").length > 0) {
+            if (jQuery(".properties-container input.field-invalid").length > 0) {
```
ZakarFin: Limit the search to the main panel.

ZakarFin:
Nice, thanks! Any suggestions how to improve the DrawTools requests?

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -171,21 +192,17 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
         _findGeometryByFidFromLayer: function (fid) {
             var layer = this.sandbox.findMapLayerFromSelectedMapLayers(this.selectedLayerId);
             var geometries = layer.getClickedGeometries();
-            var wkt = new OpenLayers.Format.WKT();
             for(var j = 0; j < geometries.length; j++) {
                 if(geometries[j][0] === fid) {
-                    var feature = wkt.read(geometries[j][1]);
-                    var geometry = feature.geometry;
-                    return {fid: fid, geometry: geometry};
+                    return {fid: fid, geometry: geometries[j][1]};
```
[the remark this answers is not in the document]
heikkiy: Yes it seems it is an array of arrays with 0 as the fid and 1 as the WKT geometry. The below is what this.sandbox.findMapLayerFromSelectedMapLayers(this.selectedLayerId).getClickedGeometries() returns for a single item in my test case for a multilinestring:
```0:"wfs_test_multiline.1"
1:"MULTILINESTRING ((..))```
And this is what this.allClickedFeatures returns in the same use case:
```fid:"wfs_test_multiline.1"
geometry:N ...```
Where the geometry is actually an OpenLayers geometry. So there was a bug actually in that the geometry should be read with ol.format.WKT().readGeometry(). AFAIK the _findGeometryByFidFromLayer shouldn't be called that much, but there is a chance still so better fix that bug.

**Comment by heikkiy** [comment]
Some rather minor mostly documentation related suggestions:

1. Should probably fix the syntax of the examples in http://www.oskari.org/documentation/requests/startdrawingrequest
2. Make it explicitly clear what the values for `allowMiltipleDrawing` (sic) are. Should also be `allowMultipleDrawing` in documentation. For example it took me quite a bit of effort to find out that you can actually draw multigeoms on the map by setting this to `multiGeom`. Should've perhaps first checked the actual implementation of DrawPlugin.ol3.js as it is clearly stated there though.
3. I couldn't find out how to check whether the DrawingEvent comes from StartDrawingRequest or StopDrawingRequest. Maybe there could be a way to check that? Boolean `stop` perhaps?
4. It isn't documented in http://www.oskari.org/documentation/requests/stopdrawingrequest that you can actually supress the event by giving a second boolean parameter to the request e.g. `Oskari.getSandBox().postRequestByName('DrawTools.StopDrawingRequest', ['asdf', true, true]);` Again i had to dig the source code to find that one out. Kind of useful for the previous point 3. as I used this feature to not trigger the event in this case.

File: bundles/tampere/content-editor/view/SideContentEditor.js [hunk 12]
```diff
@@ -1089,77 +1095,104 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
-                                    me.clickedGeometryNumber = i;
-                                    break;
+                    var geometryType = me.layerGeometries.geometry.getType();
+                    if (geometryType.indexOf("Multi") > -1) {
+                        if (me.layerGeometries.geometry.getCoordinates() != undefined) {
+                            var exists = false;
+                            for (var i = 0; i < me.layerGeometries.geometry.getCoordinates().length; i++) {
+                                if(geometryType === "MultiPoint") {
+                                    var closestPoint = me.layerGeometries.geometry.getClosestPoint([me.clickCoords.x, me.clickCoords.y]);
+                                    var distance = (
```
jampukka: This calculation doesn't work properly for geographic coordinates. However it might yield correct enough results for this purpose.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1089,77 +1095,104 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
+                                        var lineStringTransformed = [];
+                                        for(var k = 0; k < lineStringCoords.length; ++k) {
+                                            var lineStringPoint = lineStringCoords[k];
+                                            var lineStringPointWGS84 = ol.proj.transform(lineStringPoint, me.sandbox.getMap()._projectionCode, 'EPSG:4326');
+                                            lineStringTransformed.push(lineStringPointWGS84);
+                                        }
+                                        var line = turf.lineString(lineStringTransformed);
+                                        //Get default distance in kilometers
+                                        var distance = turf.pointToLineDistance(pt, line);
+                                        if(distance <= 0.05) {
```
jampukka: The hard coded 50 meter threshold doesn't account for map resolution.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1089,77 +1095,104 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
+                                        var lineStringCoords = lineStrings[j].getCoordinates();
+                                        var lineStringTransformed = [];
+                                        for(var k = 0; k < lineStringCoords.length; ++k) {
+                                            var lineStringPoint = lineStringCoords[k];
+                                            var lineStringPointWGS84 = ol.proj.transform(lineStringPoint, me.sandbox.getMap()._projectionCode, 'EPSG:4326');
+                                            lineStringTransformed.push(lineStringPointWGS84);
+                                        }
+                                        var line = turf.lineString(lineStringTransformed);
+                                        //Get default distance in kilometers
+                                        var distance = turf.pointToLineDistance(pt, line);
```
jampukka: For planar coordinate reference systems we could use `{'method': 'planar'}` as the third parameter (https://github.com/Turfjs/turf/tree/master/packages/turf-point-to-line-distance#pointtolinedistance). This way all the projection transformations could be avoided and the calculations pointToLineDistance makes would be more lightweight.

bundles/tampere/content-editor/view/SideContentEditor.js
```diff
@@ -1089,77 +1095,104 @@ Oskari.clazz.define('Oskari.tampere.bundle.content-editor.view.SideContentEditor
...
+                                        me.clickedGeometryNumber = i;
+                                        break;
+                                    }
+                                } else if(geometryType === "MultiPolygon") {
+                                    if(me.layerGeometries.geometry.intersectsCoordinate([me.clickCoords.x, me.clickCoords.y])) {
+                                        me.clickedGeometryNumber = i;
+                                        break;
+                                    }
+                                } else if(geometryType === "MultiLineString") {
+                                    if(exists) {
```
jampukka: Seems that this `exists` flag is only used for breaking the for loop. Why not just call `break;` in L1179 instead of calling `exists = true;`?

ZakarFin:
We should actually remove the documentation found in http://www.oskari.org/documentation/requests/startdrawingrequest as it has been replaced by http://oskari.org/api/requests#1.46.0/mapping/drawtools/request/startdrawingrequest.md. Do you think it has all the appropriate information? The http://oskari.org/api/requests#1.46.0/mapping/drawtools/request/stopdrawingrequest.md doesn't mention the event suppression for StopDrawingRequest. It's something that should be used very carefully and probably handled in some other way, but I guess it should be documented anyway. We have actually stumbled on the same problem with not knowing if the drawing is actually finished or just "valid" since the DrawingEvents isComplete is more about if a polygon has been closed etc. 

Thanks for the feedback @heikkiy!

PR #380, opened 2018-05-14
Guided tour - popup

nxholm:
- attr("checked") has been removed in jquery 1.9 -> use the .is(":checked") instead


ZakarFin:
Fixes an issue where checking the "Do not show this again" checkbox in guided tour didn't work/the tour was shown again even with the checkbox checked.

PR #381, opened 2018-05-16
Stats - Regionset filtering

nxholm:


bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -97,6 +100,15 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
             cont.append(tempSelect);
             me._selections.push(select);
         });
+        // map to numbers so we can filter
+        var numMap = me.regionsetRestrictions.map(function (iter) {
```
ZakarFin: Move this string -> number mapping to indicatorSelected() where me.regionsetRestrictions is set. Then you don't need the numMap variable and can use me.regionsetRestrictions instead.

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -139,6 +139,12 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
      */
     disableOptions: function (ids) {
         var chosen = this.element.find('select');
+        
+        var state = this.getOptions();
+        for (var i = 0; i < state.disabled.length; i++) {
+            jQuery(state.disabled[i]).attr('disabled', false);
+        }
```
ZakarFin: This could be a call to reset(). If you don't want to trigger the chosen:updated event you can send use some "suppressEvent" kind of optional boolean param on reset for this.

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -107,6 +119,15 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
         }
         regionSelect.container.addClass('margintop');
         cont.append(regionSelect.container);
+        var select = regionSelect.selectInstance;
+        select.disableOptions(optionsToDisable);
+        
+        var state = select.getOptions();
+        if (state.options.length - state.disabled.length === 1) {
```
ZakarFin: You could do this always. Preselect the first option that isn't disabled.

PR #383, opened 2018-05-17
Maprotator - change style fix

nxholm:
- Added scss file
- Implement the changeToolStyle method
- Eslint made lots of changes


ZakarFin:
About those formatting changes... 😱 I'll just blindly trust this does what it says. And the description doesn't really tell me anything. But from user perspective this change SHOULD implement different UI-styles available in the publisher for the maprotator bundle.

PR #385, opened 2018-05-21
Stats - multi regionset error check

nxholm:
- Previous error check was not correct since the value is an array


ZakarFin:
Removes an annoying error message when no indicator has been selected/all indicators have been removed from the on statistical maps

PR #386, opened 2018-05-22
Feature/open publisher with url parameter

jheijari:


bundles/framework/publisher2/instance.js
```diff
@@ -129,8 +123,140 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
...
+
             this._registerForGuidedTour();
         },
+        /**
+         * @private @method _getEditUuidUrlParameter
+         * Checks if "editPublished" url parameter exists.
+         * Parameter contains view uuid as value indicating that the view should be opened for editing.
+         * @return {String} uuid for the map view to edit or undefined if not found.
+         */
+        _getEditUuidUrlParameter: function () {
```
ZakarFin: Use Oskari.util.getRequestParam('editPublished') instead.

bundles/framework/publisher2/instance.js
```diff
@@ -129,8 +123,140 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
...
+            };
+        },
+        /**
+         * @private @method _sendEditRequest
+         * Sends Publisher.PublishMapEditorRequest with given uuid.
+         *
+         * @param uuid {String} Uuid for the view to edit.
+         */
+        _sendEditRequest: function (uuid) {
+            var publishMapEditorRequestBuilder = this.sandbox.getRequestBuilder(
```
ZakarFin: Use Oskari.requestBuilder('Publisher.PublishMapEditorRequest') instead and you can assume the builder exists as it's part of the same publisher bundle.

ZakarFin:
Improves the functionality where clicking "edit" on an embedded maps tab in "personal data" flyout AND the embedded map to be edited used another projection that the current map used. Previously clicking the link only reloaded the page with a new appsetup with the same projection as the embedded map and the user had to navigate back to clicking the edit link again. This PR makes the publisher open automatically after the page reload.

PR #388, opened 2018-05-22
Grid opacity fix

jheijari:
Looks like this broke down on commit [d9c26fa6a](https://github.com/oskariorg/oskari-frontend/blob/d9c26fa6aee38bca5a6cf4f20dac368a28e49f6e/bundles/framework/featuredata2/instance.js)


ZakarFin:
Fixes an issue where rows in featuredata for "manual refresh" layers opened up as "gray"/disabled looking. Now they should go "gray" after moving the map like intended to signify the features in the table might no longer be in the map viewport.

PR #391, opened 2018-05-25
Indicator server calls

ZakarFin:
- Added save, saveData and delete operations that actually call the server
- Moved lots of code dealing with StatisticsService cache out of the service and in to a new CacheHelper
- Disables indicator edit-button if there's more than one indicator selected
- Also fixes an issue where the indicator id for editing was an array
- Also fixes an issue where an allowed value for a an indicator parameter wasn't an object with id and name/UI-label, but only the value (like year)
- Some usability improvements like focusing on the fields the user is expected to fill after clicking a button

Note! This requires a matching oskari-server PR to function properly so don't merge before it's submitted (will link it next week).


ZakarFin:
Can be merged after oskariorg/oskari-server#216

PR #394, opened 2018-05-28
Feature/my indicators tab

jheijari:
New panel to user data tabbed panel "Indicators". Renders list of user created statistics data for stats layers.


bundles/statistics/statsgrid2016/instance.js
```diff
@@ -89,6 +89,27 @@ Oskari.clazz.define(
...
+                    }
+                });
+            }
+        },
+        _addIndicatorsTabToPersonalData: function (sandbox) {
+            var tab = Oskari.clazz.create('Oskari.statistics.statsgrid.MyIndicatorsTab', this);
+            tab.bindEvents();
+            var addAsFirstTab = false;
+            var reqBuilder = sandbox.getRequestBuilder('PersonalData.AddTabRequest');
+            var req = reqBuilder(tab.getTitle(), tab.getContent(), addAsFirstTab, tab.getId());
```
ZakarFin: Use Oskari.requestBuilder() instead and check if you got a function as return value. Statsgrid is used in embedded maps as well and those don't usually have personaldata bundle so the request won't be available.

bundles/statistics/statsgrid2016/MyIndicatorsTab.js
```diff
@@ -0,0 +1,316 @@
...
+         * Shows a confirmation dialog on deleting a indicator
+         *
+         * @param {Object} indicator data object for the indicator to delete
+         */
+        _confirmDelete: function (indicator) {
+            var me = this;
+            var dialog = Oskari.clazz.create('Oskari.userinterface.component.Popup');
+            var okBtn = Oskari.clazz.create('Oskari.userinterface.component.Button');
+
+            okBtn.setTitle(me.loc('tab.delete'));
```
ZakarFin: You can use Oskari.userinterface.component.buttons.DeleteButton instead so we don't need to duplicate the localization

ZakarFin: But I guess we need the localization outside the button as well so it's ok to have the localization here.

PR #397, opened 2018-05-31
FileInput component refactoring

okauppinen:
Now FileInput component is more general implementation.

Component makes drag & drop file input box if user's browser supports it. If browser doesn't support then makes basic file input element. Selected or dropped file(s) can be retrieved by getFiles() which now shows error message if there is no files or returns [files] (allow multiple) or file.

Options: allowMultipleFiles (default false), maxFileSize (default 10), allowedFileTypes (default all).

### DIFF
--- a/bundles/framework/divmanazer/component/FileInput.js
+++ b/bundles/framework/divmanazer/component/FileInput.js
@@ -36,8 +36,8 @@ Oskari.clazz.define('Oskari.userinterface.component.FileInput', function (option
                 '<input type="file" class="basic__file" accept="<%= allowedFiles %>" <%= allowMultiple %> />' +
             '</div>')
     }
-    this.isAdvancedUpload = this.canUseAdvancedUpload();
-    this.createUi();
+    this.isAdvancedUpload = this._canUseAdvancedUpload();
+    this._createUi();
 }, {
         getElement: function () {
             return this.element;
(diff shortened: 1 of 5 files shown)


bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -80,7 +112,90 @@ Oskari.clazz.define('Oskari.userinterface.component.FileInput', function () {
...
+            var fileNameElem = this.getElement().find('.box__uploaded');
+            var fileNames;
+            var files = this.files;
+            if (files.length === 0){
+                fileNameElem.text("");
+            } else {
+                fileNames = files[0].name;
+                for (var i = 1; i < files.length; i++){
+                    fileNames += ", " + files[i].name;
+                }
```
ZakarFin: You could also do this with:

	var fileNames = files.map(function (file) {
		return file.name;
	}).join(', ');

ZakarFin: I guess you wouldn't even need the if-statement if you did it like this.

bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -80,7 +112,90 @@ Oskari.clazz.define('Oskari.userinterface.component.FileInput', function () {
...
+                fileNameElem.text(fileNames);
+            }
+        },
+        validateFile: function (file){
+            var valid = true;
+            var opts = this.options;
+            var maxFileSi
```
ZakarFin: This looks like it was cut off?

bundles/framework/divmanazer/component/FileInput.js
```diff
@@ -99,27 +214,74 @@ Oskari.clazz.define('Oskari.userinterface.component.FileInput', function () {
...
+            if (opts.allowMultipleFiles === true ){
+                allowMultiple = "multiple";
+                fileUpload = this.loc('fileInput.fileUpload', {files: 2});
+            } else {
+                allowMultiple = "";
+                fileUpload = this.loc('fileInput.fileUpload', {files: 1});
+            }
+
+            if (this.isAdvancedUpload){
+                classes = "oskari-fileinput advanced-upload";
```
ZakarFin: I don't think classes need to be in a variable. You can just use the string when calling the template-function.

ZakarFin:
Nice! Can you prefix the function names that are for internal usage only with _ so it's easier to separate the "API" that's meant to be used with this component.

PR #403, opened 2018-06-04
Stats - handle multiple timeparameters in indicator specification

nxholm:
- Added a time parameter to the selector to indicate if selector describes time
- Updated select list to extend the default options variables and merge with values from implementing classes -> you only need to pass the variables you need to change now

### DIFF
--- a/bundles/framework/divmanazer/component/SelectList.js
+++ b/bundles/framework/divmanazer/component/SelectList.js
@@ -12,11 +12,12 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
     this.element = null;
 }, {
     defaultOptions: {
-      placeholder_text: '',
-      allow_single_deselect: true,
-      disable_search_threshold: 10,
-      no_results_text: '',
-      width: '100%'
+        placeholder_text: '',
+        allow_single_deselect: true,
+        disable_search_threshold: 10,
+        no_results_text: '',
+        multi: false,
+        width: '100%'
     },
     /**
      * @method create
@@ -26,8 +27,18 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
      * @param {Object} options
      * @return {jQuery Element} a list with chosen applied
      */
-    create: function (data, options) {
-        options = options || this.defaultOptions;
+    create: function (data, modifiedOptions) {
+        var me = this;
+        var getOptions = function (options) {
+            // extend default options to create new object where we concat passed values
+            var combinedOptions = jQuery.extend(true, {}, me.defaultOptions);
+            Object.keys(options).forEach(function (key) {
+                combinedOptions[key] = options[key];
+            });
+            return combinedOptions;
+        };
+        var options = getOptions(modifiedOptions);
+
         options.allowReset = options.allowReset === false ? options.allowReset : true;
         var select = this._selectTemplate.clone();
         this.element = select;
--- a/bundles/framework/divmanazer/component/SelectList.js
+++ b/bundles/framework/divmanazer/component/SelectList.js
@@ -29,11 +29,7 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
      */
     create: function (data, modifiedOptions) {
         var me = this;
-        var getOptions = function (options) {
-            // extend default options to create new object where we concat passed values
-            return jQuery.extend(true, {}, me.defaultOptions, options);
-        };
-        var options = getOptions(modifiedOptions);
+        var options = jQuery.extend(true, {}, me.defaultOptions, modifiedOptions);
 
         options.allowReset = options.allowReset === false ? options.allowReset : true;
         var select = this._selectTemplate.clone();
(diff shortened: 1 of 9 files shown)


bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -26,8 +27,18 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
      * @param {Object} options
      * @return {jQuery Element} a list with chosen applied
      */
-    create: function (data, options) {
-        options = options || this.defaultOptions;
+    create: function (data, modifiedOptions) {
+        var me = this;
+        var getOptions = function (options) {
+            // extend default options to create new object where we concat passed values
+            var combinedOptions = jQuery.extend(true, {}, me.defaultOptions);
```
ZakarFin: You can pass options as fourth parameter so you won't have to do the manual copy loop.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -85,12 +85,21 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
             var newActiveIndicator = false;
-
+            
             selectedIndicators.forEach(function (indicator) {
+                var added;
                 if (indicator === '') {
                     return;
                 }
-                var added = me.service.getStateService().addIndicator(values.datasource, indicator, values.selections);
+                if (Array.isArray(values.selections.year)) {
```
ZakarFin: Don't reference "year" here.

bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -26,8 +27,14 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
...
      */
-    create: function (data, options) {
-        options = options || this.defaultOptions;
+    create: function (data, modifiedOptions) {
+        var me = this;
+        var getOptions = function (options) {
+            // extend default options to create new object where we concat passed values
+            return jQuery.extend(true, {}, me.defaultOptions, options);
+        };
+        var options = getOptions(modifiedOptions);
```
ZakarFin: I can't see why getOptions should be a function that does nothing, but calls another function.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -85,12 +85,21 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
             var newActiveIndicator = false;
-
+            
             selectedIndicators.forEach(function (indicator) {
+                var added;
                 if (indicator === '') {
                     return;
                 }
-                var added = me.service.getStateService().addIndicator(values.datasource, indicator, values.selections);
+                if (Array.isArray(values.selections.time)) {
```
ZakarFin: Would it be a better solution to just check if any of the selections have an array as value? Then you wouldn't need to care if it's time.

bundles/statistics/statsgrid2016/components/IndicatorParameters.js
```diff
@@ -87,14 +86,24 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorParameters', function
...
+            if (selections[selected].time) {
+                options = {
+                    placeholder_text: placeholderText,
+                    multi: true
+                };
+                // assign the time describing property as "time" in the object
+                var old = selected;
+                selected = 'time';
+                Object.defineProperty(selections, selected, Object.getOwnPropertyDescriptor(selections, old));
+                delete selections[old];
```
ZakarFin: So you are renaming the variable to "time"? Doesn't this change what the getValues() return? If it does it will break the actual data fetching as it will try to ask for something like "time=2018"  from the statistical API which expects for example in SotkaNet to get "year=2018"

ZakarFin: In addition this causes problems if there's an existing variable named "time" for the indicator.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
nxholm: We still need to know the time values when they are in an array because then we need to call addIndicator in a loop for each of those values.

ZakarFin: Can't we just loop when we encounter a value that's array? addIndicator() doesn't care if there's a time variable.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -85,12 +85,25 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
+            
             selectedIndicators.forEach(function (indicator) {
+                var added;
                 if (indicator === '') {
                     return;
                 }
-                var added = me.service.getStateService().addIndicator(values.datasource, indicator, values.selections);
+                Object.keys(values.selections).forEach(function (key) {
+                    var selection = values.selections[key];
+                    if (Array.isArray(selection) && selection.length > 1) {
```
ZakarFin: Take out the selection.length > 1 as we will always want to reduce the array to single value.

PR #406, opened 2018-06-05
Force update timeseries control on app start

jheijari:


ZakarFin:
What does this fix?

jheijari:
It fixes an timing issue on app start. With some app setups the "changeActive"-event gets triggered before there's a handler for it.

ZakarFin:
Having a timeseries layer as a selected layer the user interface wasn't shown on the application start. Instead it expected timeseries layers to be added on runtime. This fixes the issue.

PR #408, opened 2018-06-07
Fixes for small mapsize in publisher

nxholm:
- Changed coordinatetoolplugin to use Oskar.getMsg
- Fixed MapLegend plugin not able to close in publisher
- Fixed layerselectionplugin not able to close in mobile
- Fixed toggling of tools in small map mode

### DIFF
--- a/bundles/framework/coordinatetool/instance.js
+++ b/bundles/framework/coordinatetool/instance.js
@@ -91,7 +91,7 @@ Oskari.clazz.define("Oskari.mapframework.bundle.coordinatetool.CoordinateToolBun
             me.setSandbox(sandbox);
             this.coordinateToolService = this.createService(sandbox, conf);
             var mapModule = sandbox.findRegisteredModuleInstance('MainMapModule');
-            var locale = this.getLocalization('display');
+            var locale = Oskari.getMsg.bind('display', 'coordinatetool');
             var plugin = Oskari.clazz.create('Oskari.mapframework.bundle.coordinatetool.plugin.CoordinateToolPlugin', this, conf, locale, mapModule, sandbox);
             mapModule.registerPlugin(plugin);
             mapModule.startPlugin(plugin);
--- a/bundles/framework/coordinatetool/plugin/CoordinateToolPlugin.js
+++ b/bundles/framework/coordinatetool/plugin/CoordinateToolPlugin.js
@@ -11,9 +11,9 @@ Oskari.clazz.define('Oskari.mapframework.bundle.coordinatetool.plugin.Coordinate
      */
     function (instance, config, locale, mapmodule, sandbox) {
         var me = this;
-        me._locale = locale;
+        me._locale = locale || Oskari.getMsg.bind('display', 'coordinatetool');
         me._config = config || {};
-        me._mapmodule = mapmodule;
+        me._mapmodule = mapmodule || Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
         me._sandbox = sandbox;
         me._instance = instance;
         me._messageDialog = null;
@@ -171,9 +171,9 @@ Oskari.clazz.define('Oskari.mapframework.bundle.coordinatetool.plugin.Coordinate
                 var dialog = Oskari.clazz.create('Oskari.userinterface.component.Popup');
                 me._dialog = dialog;
             }
-            var btn = me._dialog.createCloseButton(loc.checkValuesDialog.button);
+            var btn = me._dialog.createCloseButton(loc('display.checkValuesDialog.button'));
             btn.addClass('primary');
-            me._dialog.show(loc.checkValuesDialog.title, loc.checkValuesDialog.message, [btn]);
+            me._dialog.show(loc('display.checkValuesDialog.title'), loc('display.checkValuesDialog.message'), [btn]);
         },
         /**
          * Show popup.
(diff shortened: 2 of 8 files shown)


bundles/framework/coordinatetool/instance.js
```diff
@@ -91,7 +91,7 @@ Oskari.clazz.define("Oskari.mapframework.bundle.coordinatetool.CoordinateToolBun
             me.setSandbox(sandbox);
             this.coordinateToolService = this.createService(sandbox, conf);
             var mapModule = sandbox.findRegisteredModuleInstance('MainMapModule');
-            var locale = this.getLocalization('display');
+            var locale = Oskari.getMsg.bind('display', 'coordinatetool');
```
ZakarFin: Does 'display' in here do anything? Looks like every other piece of code using Oskari.getMsg.bind passes null as the first parameter. And later on you are still prefixing everything with "display".

ZakarFin: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind the first parameter assigns "this" for the function which makes "display" string a weird choice for that.

bundles/framework/coordinatetool/plugin/CoordinateToolPlugin.js
```diff
@@ -182,12 +182,12 @@ Oskari.clazz.define('Oskari.mapframework.bundle.coordinatetool.plugin.Coordinate
...
                 loc = me._locale,
-                popupTitle = loc.popup.title,
+                popupTitle = loc('display.popup.title'),
                 popupContent = me._templates.popupContent.clone(),
                 crs = me.getMapModule().getProjection(),
-                crsDefaultText = loc.crs.default,
+                crsDefaultText = loc('display.crs.default'),
                 popupName = 'xytoolpopup',
-                crsText = loc.crs[crs] || crsDefaultText.replace('{crs}', crs),
+                crsText = loc('display.crs.'+crs) || crsDefaultText.replace('{crs}', crs),
```
ZakarFin: getMsg() allows built-in templating/interpolation so the replace call could be removed with something like this:

      crsDefaultText = loc('display.crs.default', {crs: crs});

bundles/framework/coordinatetool/plugin/CoordinateToolPlugin.js
```diff
@@ -1203,7 +1201,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.coordinatetool.plugin.Coordinate
          */
         _allowDegrees: function(checkedProjection){
             var me = this;
-
+            var mapmodule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule')
```
ZakarFin: Missing ;

bundles/framework/coordinatetool/plugin/CoordinateTransformationExtension.js
```diff
@@ -85,8 +85,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.coordinatetool.plugin.Coordinate
             _.each(projections, function(key) {
```
ZakarFin: Use Array.forEach() instead of _.each()

bundles/framework/publisher2/tools/AbstractPluginTool.js
```diff
@@ -104,6 +104,7 @@ function(sandbox, mapmodule, localization, instance, handlers) {
             me.__started = true;
         } else {
             if(me.__started === true) {
+                me.__plugin.teardownUI();
                 me.__plugin.stopPlugin(me.__sandbox);
```
ZakarFin: What if there is not teardownUI-function in the plugin? Shouldn't it be the plugins responsibility to remove it's UI when it's stopped?

bundles/framework/publisher2/tools/LayerSelectionTool.js
```diff
@@ -88,14 +88,14 @@ function() {
         if(enabled === true) {
             me.__plugin.startPlugin(me.__sandbox);
             me.__started = true;
-            setTimeout(function(){
+            setTimeout(function() {
                 me._checkLayerSelections();
             }, 300);
         } else if(me.__started === true) {
+            me.__plugin.teardownUI();
```
ZakarFin: Shouldn't it be the plugins responsibility to remove it's UI when it's stopped instead of the publisher functionality?

bundles/framework/publisher2/tools/LayerSelectionTool.js
```diff
@@ -350,6 +350,7 @@ function() {
             }
         }
         if(me.__plugin) {
+            me.__plugin.teardownUI();
```
ZakarFin: Shouldn't it be the plugins responsibility to remove it's UI when it's stopped instead of the publisher functionality?

bundles/mapping/mapmodule/plugin/layers/LayerSelectionPlugin.js
```diff
@@ -19,32 +19,44 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.LayerSelectionP
...
         };
     }, {
+        _toggleToolState: function () {
+            var el = this.getElement();
+
+            if (this._toolOpen) {
+                if (el) {
+                    el.removeClass('active');
+                }
+                this._toolOpen = false;
```
ZakarFin: Isn't the popup.isVisible() good enough indication about this? Do we really want another variable for tracking the state?

bundles/framework/coordinatetool/plugin/CoordinateToolPlugin.js
```diff
@@ -1203,7 +1200,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.coordinatetool.plugin.Coordinate
          */
         _allowDegrees: function(checkedProjection){
             var me = this;
-
+            var mapmodule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
```
ZakarFin: The variable mapmodule isn't used here.

bundles/framework/coordinatetool/plugin/CoordinateTransformationExtension.js
```diff
@@ -82,11 +82,11 @@ Oskari.clazz.define('Oskari.mapframework.bundle.coordinatetool.plugin.Coordinate
          _populateCoordinatesTransformSelect: function(select) {
             var me = this,
                 projections = me._config.supportedProjections;
```
ZakarFin: Use "me._config.supportedProjections || []" to make sure the forEach() call doesn't fail.

PR #417, opened 2018-06-14
Tampere search channel changes for ol2 to ol4

ZenJappanen:
Ported ol2 functions to ol4 but this is only partial solution. There is a problem with the MapMoveRequest. Old implementation provided bounding box to the request and it seemed to work. The API has changed since and it requires zoomLevel (0-12) as attribute and not bounding box. 

This commit includes the request with bounding box and it does not zoom correctly. Is it possible to have old request functionality with bbox back?

### DIFF
--- a/bundles/tampere/search-from-channels/instance.js
+++ b/bundles/tampere/search-from-channels/instance.js
@@ -812,7 +812,7 @@ Oskari.clazz.define(
                 },
                 stroke: {
                   color: '#FF0000',
-                  width: 10
+                  width: 1
                 },
                 text : {
                   scale : 1.3,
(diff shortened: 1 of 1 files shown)


ZenJappanen:
New version with fixed zoom.

ZenJappanen:
Reopened.

bundles/tampere/search-from-channels/instance.js
```diff
@@ -806,16 +806,27 @@ Oskari.clazz.define(
...
+                text : {
+                  scale : 1.3,
+                  fill : {
+                    color : 'rgba(0,0,0,1)'
+                  },
+                  stroke : {
+                    color : 'rgba(255,255,255,0.8)',
+                    width : 2
+                  },
+                  labelProperty: 'test_property'
```
ZakarFin: Seems you don’t have an attribute named ’test_property’ for the features so this styling can be removed.

ZakarFin:
Check the style if everything is needed there. At least labeling seems a bit off. Other than that this looks ready for merging 👍

PR #426, opened 2018-06-25
Publisher2 mobile tools fix

nxholm:
- Don't call Publisher2.ToolEnabledChangedEvent in mobile since it calls set location to plugins -> plugins end up on map instead of mobile toolbar

### DIFF
--- a/bundles/framework/publisher2/tools/AbstractPluginTool.js
+++ b/bundles/framework/publisher2/tools/AbstractPluginTool.js
@@ -107,8 +107,10 @@ function(sandbox, mapmodule, localization, instance, handlers) {
                 me.stop();
             }
         }
-        var event = Oskari.eventBuilder('Publisher2.ToolEnabledChangedEvent')(me);
-        sandbox.notifyAll(event);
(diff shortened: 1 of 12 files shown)


bundles/framework/publisher2/tools/AbstractPluginTool.js
```diff
@@ -107,8 +107,10 @@ function(sandbox, mapmodule, localization, instance, handlers) {
                 me.stop();
             }
         }
-        var event = Oskari.eventBuilder('Publisher2.ToolEnabledChangedEvent')(me);
-        sandbox.notifyAll(event);
+        if (!Oskari.util.isMobile()) {
```
ZakarFin: This is the wrong place to solve the issue. Things start to go really weird really quickly at some point if "tool was enabled/disabled" events are sent or NOT sent depending on the map size.

ZakarFin:
Fixes an issue with "Map movement by dragging allowed" control didn't do anything in the publisher preview. Also touches featuredata visibility and a bunch of other things in publisher... that's to say this PR includes too many changes.

PR #431, opened 2018-06-27
Load sample 3D dataset

jheijari:


ZakarFin:
Assumes sample dataset is in the same server where the frontend is loaded from and in directory /3dtiles/helsinki/. This is part of the 3d-poc for paikkatietoikkuna.fi

PR #433, opened 2018-06-28
Publisher2 tool stop

nxholm:


ZakarFin:
Fixes an issue where tools don’t stay disabled in publisher preview when changing the map size

PR #437, opened 2018-07-10
Printout scale text

kuosman:
This PR allows PDF print show scale text.
Scale text is showed when user has selected map scale on UI, scales showed like 1:500, 1:1000 and so on dropdown list. 
When user selected map scale on selection, map is zoomed defined scale and also zoom by mouse/keyboard are disabled. When used map selection scale zoom by mouse/keyboard is enabled back.

Scale selection is visible when printout bundle config has following config:
```javascript
{
    "scaleSelection": true
}
```

UI uses by default map scales on selection box when it is visible, but you can also override default options following config:
```javascript
{
    "scales": [500,1000,10000,25000,50000,100000]
}
```

This PR also unfortunately also includes small modifications of mapmodule and requests.

Mapmodule changes:
- New `setResolution`-function: Sets the map resolution to given resolution
- New `getExactResolution`-funtion: Gets exact resolution for given scale, if imp function not implemented then use `getResolutionForScale` -function
- Changed `zoomToScale`-function. When closed param is true, used `getExactResolution` -function. 

Requests changes:
- EnableMapKeyboardMovementRequest, EnableMapMouseMovementRequest: can enable all interactions or defined interaction (`pan` or `zoom`)
- DisableMapKeyboardMovementRequest, DisableMapMouseMovementRequest: can disable all interactions or defined interaction (`pan` or `zoom`)

Server side code also need merged at text is showed (https://github.com/oskariorg/oskari-server/pull/235).


bundles/mapping/mapmodule/AbstractMapModule.js
```diff
@@ -716,11 +736,33 @@ Oskari.clazz.define(
...
+         * @method getExactResolution
+         * @param  {Float}           scale the new scale
+         * @return {Float}           exact resolution
+         */
+        getExactResolution: function(scale) {
+            if(typeof this._getExactResolutionImpl === 'function') {
+                return this._getExactResolutionImpl(scale);
+            }
+
+            return this.getResolutionForScale(scale);
```
ZakarFin: This should throw an exception if the exact resolution is not available.

bundles/mapping/mapmodule/mapmodule.ol3.js
```diff
@@ -593,6 +621,16 @@ Oskari.clazz.define('Oskari.mapframework.ui.module.common.MapModule',
         _setZoomLevelImpl: function (newZoomLevel) {
             this.getMap().getView().setZoom(newZoomLevel);
         },
+        _setResolutionImpl: function (newResolution) {
+            this.getMap().getView().setResolution(newResolution);
+        },
+        _getExactResolutionImpl: function(scale){
+            var units = this.getMap().getView().getProjection().getUnits();
+            var dpi = 25.4 / 0.28;
```
ZakarFin: Doesn't this depend on the screen that's been used?

ZakarFin:
I'm not sure this actually does what it's supposed to do with the images that are used on the printed out output. It relies completely for the frontend for scaling the images on screen and assumes they are scaled for printing as well. Having two parameters to trigger the "scaletext" mode on the server is a bit much and there's no warning that some of the layers (wmts) won't be included in the printout when printing in scale.

PR #440, opened 2018-07-10
Removed deprecated core parts

data-ux:
Deprecated parts of Oskari core `module_spec.js` and `builder_api.js` were removed and code depending on them was refactored to use remaining core APIs. After this change no code in the oskari-fontend repo depends on the removed core APIs, but from a public API perspective this is a breaking change.

Rationale for this change was to limit the API usage of the class system from other parts of Oskari core (in anticipation to changes induced to the class system by ES6 class).


ZakarFin:
Looks good - just need to rebuild bundles/bundle.js to get the conflicts resolved since #439 changed it as well

data-ux:
Fixed conflict.

PR #441, opened 2018-07-11
Refactored class system

data-ux:
Oskari class system rewritten to internally use classes defined by EcmaScript constructor functions and prototypal inheritance. This is the normal, idiomatic way of defining classes in JavaScript and ES6 classes are built on this mechanism. After this change Oskari classes are compatible with ES6 classes.

The main semantics of the class system API remain unchanged. Classes/categories can be defined in any order. The chain of super classes just needs to be defined before instantiating as before. Only changes to the API are:
- `getMetadata(...)` now returns the same metadata given at `.define(...)`
- `define(...)`, `category(...)` etc. do not return `classInfo` anymore as it is regarded as class system internal state. Other parts of Oskari core can request classInfo with `_getClassInfo(...)`.
- Only single inheritance is now suported, but there were no classes using multiple inheritance in oskari-frontend.
-  removed public methods `purge`, `lookup`, `extend`, `createWithClassInfo `, `getBuilderFromClassInfo` as they are no longer used anywhere.
- New public methods `get(...)` and `defineES(...)` added. See below for more.

**ES6 integration**
Native ES6 classes can be defined as Oskari classes with `Oskari.clazz.defineES(...)`. When browsers catch up and ES5 transpilation can be omitted, no changes will be needed. ES6 classes transpiled to ES5 constructor functions work, too.

A reference to the Oskari class' ES constructor can be obtained with `Oskari.clazz.get(...)`. It can be used directly to create an instance with `new` keyword or it can be used with the `extends` keyword to define a subclass inheriting from the class.

```javascript
const DefaultExtension = Oskari.clazz.get('Oskari.userinterface.extension.DefaultExtension');
Oskari.clazz.defineES('Oskari.MyNameSpace.MyBundleInstance',
    class MyBundleInstance extends DefaultExtension {
        ...
    },
    {
        protocol: [
            'Oskari.bundle.BundleInstance',
            'Oskari.mapframework.module.Module',
            'Oskari.userinterface.Extension'
        ]
    }
);
```

**Limitations**
- ES6 classes can only extend existing Oskari classes by obtaining a constructor reference via `.get(...)` and using the `extends` keyword. Extending by supplying metadata is not possible. The class that is being extended must be registered in the class system before the call to `.get(...)`.
- Classes defined with the traditional `.create(...)` cannot extend ES6 classes in any way.

**Testing**
This change has been extensively tested manually and all Robot tests pass on demo-kartta.paikkatietoikkuna.fi.


src/class_system.js
```diff
@@ -0,0 +1,273 @@
...
+    function processMetadata(classInfo, metadata) {
+        if (!metadata) {
+            return true;
+        }
+        classInfo.metadata = metadata;
+        processProtocols(classInfo, metadata.protocol);
+        if (!metadata.extend) {
+            return true;
+        }
+        if (!Array.isArray(metadata.extend) || metadata.extend.length > 1) {
```
ZakarFin: Doesn't really make sense to require an array with only one value. Looking at any existing code will make you think that adding classes to the array will work. But as they are currently arrays in the existing code it's a minor grievance since we are moving towards the ES6 class syntax anyway. Nice that it throws an error with a clear message.

src/class_system.js
```diff
@@ -0,0 +1,273 @@
...
+        /**
+         * @public @method protocol
+         * Get list of classNames that implement given protocol
+         *
+         * @param  {string} protocolName Protocol name
+         *
+         * @return {string[]}              ClassNames implementing protocol
+         */
+        protocol: function(protocolName) {
+            return Object.keys(protocols[protocolName] || {});
```
ZakarFin: The fallback should be an empty array []?

src/class_system.js
data-ux: Yes, now that only single inheritance is supported, the array is no longer needed other than for backward compatibility with existing code.

PR #443, opened 2018-07-31
Statsgrid - empty indicator handling

nxholm:
- updated icons
- added checks for the data actually containing values


ZakarFin:
More user-friendly UI when an indicator without data is shown.

PR #446, opened 2018-08-06
Statsgrid - Button to reset user selections during search

nxholm:
- IndicatorSelection.js stores the SelectList instances in an array now.
- Added boolean resetInProgress, indicates when reset has been clicked. Used to control what functions get called in the bubbling .on('change') events.


ZakarFin:
Did you try reseting the datasource select and the regionset filter? Clearing those two should reset everything else without additional tracking variables.

PR #453, opened 2018-08-14
Multiselect

nxholm:
This pr contains a lot of library changes
- Update jQuery to version 1.12.4
- Update jQuery-Ui to version 1.12.1
- Add library semantic ui - for now it's the whole library - still have to figure out how to properly include only the dropdown component and it's dependency components (individual components are found in libraries/semantic-ui-dropdown/components)
- add `<script type="text/javascript" src="/Oskari/libraries/jquery/jquery-1.12.4.min.js"></script>` to index.jsp

- multiselect component in itself has not yet been implemented anywhere


ZakarFin:
Like you said it's a bit much for just having another kind of UI for a multiselect. Let's postpone merging until we'll see if this is a good idea and if the lib could be used for other components as well or until we can do a smaller customized build for it.

ZakarFin:
Decided not to use semantic-ui for this. Other changes have been done in other pull requests already. Thanks!

PR #455, opened 2018-08-15
Eventhandler fix

jheijari:
AfterRearrangeSelectedMapLayerEvent doesn't have getMapLayer function.


jheijari:
Removed as a duplicate fix.

PR #486, opened 2018-08-30
Locale scripts

data-ux:
Added `extract-locale` and `inject-locale` scripts for pulling localizations strings into JSON and pushing them back. Both scripts take only one parameter, the locale identifier, for example "fi" or "sv".

Extract produces JSON files under `locale/`, one for each bundle.

Inject reads the JSON files from `locale/` and creates bundle locale files under `resources/locale/` for each bundle that a JSON file was found for. The operation overwrites any existing locale file for a specific locale.

Note: locale files that have multiple calls to `Oskari.registerLocalization(...)` will be skipped while extracting.

### DIFF
--- a/tools/extract-locale.js
+++ b/tools/extract-locale.js
@@ -69,3 +69,5 @@ Object.keys(collection)
         ensureDirectoryExists(outputPath);
         fs.writeFileSync(outputPath, JSON.stringify(output, null, 2));
     });
+
+console.log('Extracted files succesfully to "locale" directory in repo root');
--- a/tools/extract-locale.js
+++ b/tools/extract-locale.js
@@ -1,0 +1,64 @@
+var path = require('path');
+var fs = require('fs');
+var glob = require('glob');
+var merge = require('merge');
+
+function deconstructPath(filePath, chain) {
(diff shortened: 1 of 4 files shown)


ZakarFin:
Looks good. Add some usage instructions to tools/README.md

tools/extract-locale.js
```diff
@@ -0,0 +1,73 @@
...
+    return deconstructPath(path.dirname(filePath), chain);
+}
+
+function ensureDirectoryExists(filePath) {
+    var dirname = path.dirname(filePath);
+    if (fs.existsSync(dirname)) {
+      return true;
+    }
+    ensureDirectoryExists(dirname);
+    fs.mkdirSync(dirname);
```
ZakarFin: Shouldn't mkdir and recursion be switched?

data-ux: We are recursing from the end of the path down (right to left). By calling the self-recursion first, we first make sure all parent directories exist. After that's done, we can create the `dirname` on the "current" level.

If we were recursing from start to end (left to right), then we would first do `mkdirSync`.

PR #487, opened 2018-08-31
Stats - Disable unsupported regionsets for a selected datasource

nxholm:
- Disable unsupported regionsets in the dropdown menu when selecting datasource


bundles/framework/divmanazer/component/SelectList.js
```diff
@@ -182,6 +182,8 @@ Oskari.clazz.define('Oskari.userinterface.component.SelectList', function (id) {
         }
         if (!supressEvent) {
             this.resetToPlaceholder();
+        } else {
+            this.element.find('select').trigger('chosen:updated');
```
ZakarFin: Should suppressEvent be renamed to suppressChangeEvent? That's the event that is not triggered if the reset is called with parameter true?

PR #502, opened 2018-09-10
Layersplugin visibility event

okauppinen:
Handle ol maplayers visibility changes in layers plugin. Also removed ol dependency from MapLayerVisibilitiyRequestHandler.

Also reduced wait time from 1500ms to 750ms for scheduled visibility check. OL4 sends map move start event which resets the timer and user have reported too long latency with scale-limited layers.


bundles/mapping/mapmodule/plugin/layers/LayersPlugin.ol3.js
```diff
@@ -239,25 +240,30 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.LayersPlugin',
...
+                mapLayers.forEach(function (ol) {
+                    ol.setVisible(false);
+                });
+                this.notifyLayerVisibilityChanged(layer, scaleOk, geometryMatch);
             }
         }
-        var event = this._sandbox.getEventBuilder('MapLayerVisibilityChangedEvent')(layer, scaleOk, geometryMatch);
+    },
+    notifyLayerVisibilityChanged: function (layer, inScale, geometryMatch){
+        var event = this._sandbox.getEventBuilder('MapLayerVisibilityChangedEvent')(layer, inScale, geometryMatch);
```
ZakarFin: Change this._sandbox.getEventBuilder() -> Oskari.eventBuilder() to get rid of the deprecated warning.

PR #507, opened 2018-09-11
Migrate to OpenLayers 5.x

data-ux:
OpenLayers dependency as ES6 modules from NPM. Will only work with a bundler (Webpack).

OL modules needed in each file are imported to the local scope. The local name (variable name) of the imported class/function mirrors the path to the module. For example 'ol/source/Vector' is imported as `olSourceVector`. This way the naming across files stays consistent. The geometry classes are imported as one module with `import * as olGeom from 'ol/geom';` to avoid having to import all classes separately. Referencing the classes happens with `new olGeom.Point(...)`, `new olGeom.Polygon(...)` etc.

JSTS library was also migrated to ES56 modules & NPM, because it has OpenLayers as dependency.


ZakarFin:
Good job, will merge after we get the release sorted.

PR #511, opened 2018-09-11
Fix initializing vectorlayer

jheijari:


ZakarFin:
Previously statistics functionality added the regionset layer on startup to selected layers. Now it adds it after statistics indicator is added to the map.

PR #512, opened 2018-09-12
Fix tool positions

jheijari:
Set legend and series control locations absolute on top right plugin slot.


jheijari:
This is not complete yet.

PR #542, opened 2018-09-19
Restore analysis tab on personal data

ZakarFin:
Analyse now checks if it's loaded before personaldata and tries to add the tab only when personaldata is available.

### DIFF
--- a/bundles/analysis/analyse/instance.js
+++ b/bundles/analysis/analyse/instance.js
@@ -112,7 +112,7 @@ Oskari.clazz.define(
                 'Oskari.analysis.bundle.analyse.request.AnalyseRequestHandler',
                 me
             );
-            sandbox.addRequestHandler(
+            sandbox.requestHandler(
                 'analyse.AnalyseRequest',
                 me.analyseHandler
             );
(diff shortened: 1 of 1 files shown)


ZakarFin:
Without this fix yes. This was already implemented since _addTab is called both when the bundle starts (handles the case where personaldata has been started before analysis) and on an eventhandler for 'Personaldata.PersonaldataLoadedEvent' (https://github.com/oskariorg/oskari-frontend/blob/1.47.1/bundles/analysis/analyse/instance.js#L221-L223), but broken in #541 with a seemingly innocent replacement of the deprecated sandbox.getRequestBuilder() with Oskari.requestBuilder(). 

Oskari.requestBuilder() returns a function if the request is present. With minified code this happens if the request is provided in a bundle that's part of the appsetup. With non-minified code this happens if the bundle has been _loaded_ prior to calling Oskari.requestBuilder(). Where as sandbox.getRequestBuilder() returns a function only if both the request is available AND something is actually handling it already. 

So in this PR the actual change is that instead of checking if Oskari.requestBuilder() returns a truthy value (it will be a function if the request is available) to checking if something is actually going to handle the request with sandbox.hasHandler('PersonalData.AddTabRequest').

ZakarFin:
To be exact this problem is only reproducible with minified code since that's the only case where requestBuilder will be available before the request has a handler.

PR #554, opened 2018-09-25
Save vector tile layer

jheijari:
Saving vector tile layer as an admin user.
- new field "options"
- fixed an issue where newly added layer would not appear in the layer selected before page reload.

### DIFF
--- a/bundles/integration/admin-layerselector/templates/adminLayerSettingsTemplate.html
+++ b/bundles/integration/admin-layerselector/templates/adminLayerSettingsTemplate.html
@@ -252,7 +252,7 @@
         </div>
     </div>
 
-    <!-- % if(model.getStyles().length > 1) { % -->
+    <% if (layerType !== "vectortile" && layerType !== "tiles3d") { %>
     <div class="add-layer-record">
         <div class="add-layer-label-area">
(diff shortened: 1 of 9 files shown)


bundles/integration/admin-layerselector/templates/adminLayerSettingsTemplate.html
```diff
@@ -252,7 +252,7 @@
         </div>
     </div>
 
-    <!-- % if(model.getStyles().length > 1) { % -->
+    <% if (layerType !== "vectortile" && layerType !== "tiles3d") { %>
```
ZakarFin: Why is default style not selectable for these? I guess it doesn't make much sense to have the select, but it kinda bothers me to see layertype-specific stuff on the "generic" template.

bundles/integration/admin-layerselector/templates/adminLayerSettingsTemplate.html
```diff
@@ -375,6 +375,7 @@
         </div>
     </div>
     
+    <% if (layerType === "wms" || layerType === "wmts") { %>
```
ZakarFin: This should be moved to wms and wmts wmslayerSettingsTemplateFooter.html and wmtslayerSettingsTemplateFooter.html.

PR #558, opened 2018-09-27
Webpack config for multiple appsetups

jheijari:
Requires changes to application build parameters.
env.appdef=applications/application/guest/minifierAppSetup.json -> env.appdef=applications/application


data-ux:
Looks good!

PR #570, opened 2018-10-03
Hover for vector tile layers

jheijari:
Adds hover and feature click functionalities for vector tile layers.
Hover settings can me adjusted using VectorLayerRequest.
Synchronized usage of map hover, click and VectorLayerRequest between vectorlayers and vectortilelayers.


bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileLayerPlugin.js
```diff
@@ -89,16 +99,74 @@ class VectorTileLayerPlugin extends AbstractMapLayerPlugin {
...
+        // Properties id, type and hover are being used in VectorFeatureService.
         const vectorTileLayer = new olLayerVectorTile({
             opacity: layer.getOpacity() / 100,
             renderMode: 'hybrid',
             source: new olSourceVectorTile(sourceOpts),
-            style: this._getLayerCurrentStyleFunction(layer)
+            style: this._getLayerCurrentStyleFunction(layer),
+            [LAYER_ID]: layer.getId(),
+            [LAYER_TYPE]: layertype,
+            [LAYER_HOVER]: layer.getHoverOptions()
```
data-ux: The ol/layer/VectorTile constructor options has a specific type signature:
http://openlayers.org/en/v5.2.0/apidoc/module-ol_layer_VectorTile-VectorTileLayer.html#VectorTileLayer

We shouldn't use the OL options-object as storing place for "extra" layer state. The options is meant for giving options for OL internal use. There is no guarantee these "extra" options will be stored in future versions of OL. In any case the options have a specific type and we should stick to that.

jheijari: The documentation states: "Note that any property set in the options is set as a module:ol/Object~BaseObject property on the layer object; for example, setting title: 'My Title' in the options means that title is observable, and has get/set accessors."

data-ux: The above excerpt just means that each valid options key has corresponding setter/getter on the layer instance. For example key `style` -> `getStyle()` on the layer. It's not meant as an endorsement for adding keys not listed in the type signature.

bundles/mapping/mapmodule/plugin/vectorlayer/VectorLayerPlugin.ol3.js
```diff
@@ -439,10 +411,14 @@ Oskari.clazz.define(
...
-                    name: me._olLayerPrefix + layer.getId(),
-                    id: layer.getId(),
                     source: new olSourceVector()
                 });
+                // Set oskari properties
+                const silent = true;
+                olLayer.set(LAYER_ID, layer.getId(), silent);
+                olLayer.set(LAYER_TYPE, layer.getLayerType(), silent);
+                olLayer.set(LAYER_HOVER, layer.getHoverOptions(), silent);
+                olLayer.set(LAYER_NAME, me._olLayerPrefix + layer.getId(), silent);
```
data-ux: Moving the custom properties from `options` into layer object properties is actually even more problematic. Here we are for example setting the the properties "id" and "name" of the OL layer object. I think there is a high probability OL might use these properties for something in the future resulting in a conflict. 

This is a potential WTF for any future developers. The solution is not to attach these custom data to the OL layer object in any way.

jheijari: The newly added properties are prefixed with "oskari" to reduce propability of Openlayers ever using the same property names. Properties "id" and "name" are part of the old implementation and the task at hand is not to refactor existing implementation. Extra properties have been moved away from the constructor to clarify, that these are not native OL properties.

jheijari: [OL BaseObject.set/get](https://openlayers.org/en/latest/apidoc/module-ol_Object-BaseObject.html) is widely used in the codebase already.

bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileLayerPlugin.js
```diff
@@ -8,6 +8,12 @@ import VectorTileModelBuilder from './VectorTileModelBuilder';
 import VectorTileLayer from './VectorTileLayer';
 import styleGenerator from './styleGenerator';
 
+const LAYER_ID = 'oskariId';
```
ZakarFin: These could be moved to some "constants" file and reused instead of declaring on multiple places.

bundles/mapping/mapmodule/plugin/vectorlayer/VectorLayerPlugin.ol3.js
ZakarFin: Jan has very good points for conflicting names but in my opinion it's ok to use the setters now that they have the oskari-prefixed names. Perhaps it's a bit overkill to have them as observables, but looks like OL docs agree with using the setter like this ("You can add your own observable properties with object.set('prop', 'value')") and it reduces complexity for having the layer-related props on the layer itself. 

We could of course get these from the OskariLayer as well where needed. Not sure if there would be some difference in performance on hover. Perhaps not if the OskariLayer-instance is stored on the hover handler and not fetched again and again for each feature from the service.

PR #577, opened 2018-10-03
Enforce single quote string literals

data-ux:
Generated with `eslint --fix` with rule:
`"quotes": ["error", "single", {"allowTemplateLiterals": true, "avoidEscape": true}]`

meaning double quotes are allowed if string literal contains single quotes (that would need to be escaped)


data-ux:
localization files not included to keep compatibility with JSON (requires double quotes)

ZakarFin:
Sorry, merged the mapmodule es6-module stuff before this so there's some conflicts to be solved. Can you add the eslint rule in 'oskari-frontend/.eslintrc.js'. Looks like it's not changed in this PR.

PR #602, opened 2018-10-12
Use Oskari.urls.getRoute() instead of global variable

ZakarFin:
Builds on #600 and replaces global ajaxUrl variable reference with Oskari.urls.getRoute() call.

### DIFF
--- a/bundles/framework/admin-layerrights/Flyout.js
+++ b/bundles/framework/admin-layerrights/Flyout.js
@@ -448,8 +448,7 @@ Oskari.clazz.define('Oskari.framework.bundle.admin-layerrights.Flyout',
             var me = this;
             me.progressSpinner.start();
 
-            jQuery.getJSON(ajaxUrl, {
-                action_route: 'GetPermissionsLayerHandlers',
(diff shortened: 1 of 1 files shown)


bundles/framework/admin-layerrights/Flyout.js
```diff
@@ -527,29 +525,28 @@ Oskari.clazz.define('Oskari.framework.bundle.admin-layerrights.Flyout',
...
 
             externalIdSelect.html('');
-            if (externalType !== '0') {
+            if (externalType === '0') {
+                return;
+            }
+            var optionEl = document.createElement('option');
+            optionEl.value = '0';
+            optionEl.textContent = '-- ' + rightsLoc.selectValue + ' --';
+            if ('' + selectedId === '0') {
```
data-ux: It's safe to make the comparison with the number `0`. The method is currently only called with number argument as `selectedId` (see line 259)

ZakarFin: Good catch. Removed selectedId and externalType from params as they are basically constants.

PR #605, opened 2018-10-12
Added instructions for frontend developers

ZakarFin:
Added instructions for using the frontend builds with FAQ section for memory issues: https://github.com/ZakarFin/oskari-frontend/blob/build-instructions/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 Documentation available at [http://www.oskari.org].
 
-This repository holds the front-end code for Oskari. Developing the front-end requires an oskari-server to be running that responds to the XHR requests made by the front-end. You can download a pre-compiled copy of the server from  [http://www.oskari.org/download] or checkout the source code in the oskari-server repository. Any customized application should use the oskari-server-extension template as base for customized server. The front-end code is build using Webpack.
+This repository holds the front-end code for Oskari. Developing the front-end requires an oskari-server to be running that responds to the XHR requests made by the front-end. You can download a pre-compiled copy of the server from  [http://www.oskari.org/download] or checkout the source code in the oskari-server repository. Any customized application should use the oskari-server-extension template as base for customized server. The front-end code is built using Webpack.
 
 ## Preparations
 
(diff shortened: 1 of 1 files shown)


ZakarFin:
A local oskari-server is actually not strictly needed when changing the proxy target to a matching external server and giving a version for npm start command. Not sure if we want to advertise it though. There's also an issue with not all apps under applications/sample having a minifierAppsetup.json file which breaks the sample commands,

README.md
```diff
@@ -4,6 +4,66 @@
...
+To build minifed JS and assets, run:
+`npm run build -- --env.appdef=1.49.0:applications/sample/`
+
+The number before the colon sets the directory name, here producing files under dist/1.49.0/servlet/
+
+Note: The version number given for the build command needs to match the client version (`oskari.client.version`) in Oskari-server `oskari-ext.properties`.
+
+### Customized application icons
+
+For customizing icons include a folder named `icons` under the application folder (for example in sample app `applications/sample/servlet/icons`) with files named similarly (and with same pixel size for maximum compatibility) as in `oskari-frontend/resources/icons`. You can do a partial customization by for example just having icon-arrow-down.png in the application icons folder and the build will create an icon-set with default icons only overriding the icon-arrow-down.png.
```
data-ux: Maybe we should start this section with emphasising that this step is only needed if the application actually has custom icons?

Also the last sentence is somewhat long and hard to parse. It could be misunderstood that the default icons are overriding, not the other way around.

README.md
```diff
@@ -4,6 +4,66 @@
 
 Documentation available at [http://www.oskari.org].
 
+This repository holds the front-end code for Oskari. Developing the front-end requires an oskari-server to be running that responds to the XHR requests made by the front-end. You can download a pre-compiled copy of the server from  [http://www.oskari.org/download] or checkout the source code in the oskari-server repository. Any customized application should use the oskari-server-extension template as base for customized server. The front-end code is build using Webpack.
```
data-ux: is *built* using

PR #615, opened 2018-10-18
Feature/visualization form

atlain:
Visualization form ui changes

### DIFF
--- a/bundles/framework/divmanazer/resources/scss/visualizationform.scss
+++ b/bundles/framework/divmanazer/resources/scss/visualizationform.scss
@@ -303,11 +303,6 @@ div.lineform {
     }
 }
 
-div.areaform div.column1,
-div.areaform div.column21 {
-    width: 115px !important;
-}
-
 div.lineform div.column1 {
     width:130px;
 }
@@ -332,6 +327,11 @@ div.areaform div.column221 {
     padding-right: 5px;
 }
 
+div.areaform div.column1,
+div.areaform div.column21 {
+    width: 115px;
+}
+
 div.pointform div.column22,
 div.lineform div.column22,
 div.areaform div.column222 {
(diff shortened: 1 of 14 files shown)


bundles/framework/divmanazer/resources/scss/visualizationform.scss
```diff
@@ -283,26 +290,44 @@ div.areaform, div.lineform {
...
 
+div.lineform {
+    div.column1 {
+        width: 130px;
+    }
+}
+
+div.areaform div.column1,
+div.areaform div.column21 {
+    width: 115px !important;
```
data-ux: We should avoid using `!important`, because it is hard to override. Increase selector specificity instead:
https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

bundles/framework/divmanazer/resources/scss/visualizationform.scss
```diff
@@ -311,20 +336,24 @@ div.pointform div.column22,
...
     width: 530px;
 }
 
 div.areavisualization {
-    width: 730px;
+    width: 430px;
 }
+
+label.oskari-formcomponent.oskari-colorpickerinput input {
+    width: 170px !important;
```
data-ux: `!important` here, too.

resources/icons.css
```diff
@@ -398,6 +398,12 @@
     background-position: -985px 0px !important;
 }
 
+.icon-line-solid {
+    width: 28px;
+    height: 28px;
+    background-color: black !important;
+}
+
```
data-ux: resources/icons.css is generated with a script. Any manual changes here will be overwritten next time it's run. We should put these styles elsewhere and give them some other name so they are not confused with regular image icons.

atlain:
Do not merge yet. Waiting for a solution to handle new fill style options.

bundles/framework/myplacesimport/StyleForm.js
```diff
@@ -58,10 +58,8 @@ Oskari.clazz.define(
          * @param {Object} style
          */
         setStyleValues: function (style) {
-            style.dot.color = this._formatColorFromServer(style.dot.color);
-            style.line.color = this._formatColorFromServer(style.line.color);
-            style.area.lineColor = (typeof style.area.lineColor === 'string' ? this._formatColorFromServer(style.area.lineColor) : null);
-            style.area.fillColor = (typeof style.area.fillColor === 'string' ? this._formatColorFromServer(style.area.fillColor) : null);
+            style.area.lineColor = (typeof style.area.lineColor === 'string' ? style.area.lineColor : null);
```
ZakarFin: Can _prefixColorForServer() and _formatColorFromServer() be removed now that the calls have been removed?

atlain: Yep. Added new commit.

PR #617, opened 2018-10-22
Reduced shininess of 3d tiles material

data-ux:
The tileset material produced by FME has a metallic component that causes unnecessary shininess in the 3D visualisation. This material **should** be corrected in the production side of the dataset, but in the mean time we override it in the client.

Run-time modification of the Cesium `Model` class prototype was unfortunately the only way to achieve the result.

3D view time is now configured to be always daytime (used client current time before).


data-ux:
Travis check has actually passed, but the status isn't reflected in the GitHub UI.

PR #620, opened 2018-10-24
Metadata fixes and error handling

atlain:
Fixes problem where metadata accordion is not open when the popup is opened.
Added error handling for request problems and empty metadatas.

### DIFF
--- a/bundles/catalogue/metadataflyout/view/MetadataPage.js
+++ b/bundles/catalogue/metadataflyout/view/MetadataPage.js
@@ -49,22 +49,28 @@ Oskari.clazz.define('Oskari.catalogue.bundle.metadataflyout.view.MetadataPage',
 
             template = _.extend({}, data);
             delete template.identifications;
-            // Create a panel for each identification
-            for (i = 0; i < data.identifications.length; i += 1) {
-                model = _.extend({}, template);
-                model.identification = data.identifications[i];
-                panel = Oskari.clazz.create(
-                    'Oskari.catalogue.bundle.metadataflyout.view.MetadataPanel',
-                    me.instance,
-                    me.locale,
-                    model
-                );
-                if (me.asyncTabs && !jQuery.isEmptyObject(me.asyncTabs)) {
-                    panel.addTabs(me.asyncTabs);
-                }
 
-                me.addPanel(panel);
-                panel.init(i === 0);
+            if (data.identifications.length === 0) {
+                //  No identifications, show metadata not found message
+                me._showMetadataNotFoundMessage();
+            } else {
+                // Create a panel for each identification
+                for (i = 0; i < data.identifications.length; i += 1) {
+                    model = _.extend({}, template);
+                    model.identification = data.identifications[i];
+                    panel = Oskari.clazz.create(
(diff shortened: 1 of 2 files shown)


bundles/catalogue/metadataflyout/view/MetadataPanel.js
```diff
@@ -176,7 +176,7 @@ Oskari.clazz.define('Oskari.catalogue.bundle.metadataflyout.view.MetadataPanel',
                     '        </ul>' +
                     '    <% } %>' +
 
-                    '    <% if (identification.citation.date.date.length) { %>' +
+                    '    <% if (identification.citation.date.date != null && identification.citation.date.date.length) { %>' +
```
data-ux: If we can assume the type of `identification.citation.date.date` is always string or null/undefined, we could just use it directly as the condition in `if`:
```
<% if (identification.citation.date.date) { %>'
```
This will take care of both the null check and possible empty string, because an empty string is evaluated as "falsy".

bundles/catalogue/metadataflyout/view/MetadataPanel.js
```diff
@@ -342,7 +342,7 @@ Oskari.clazz.define('Oskari.catalogue.bundle.metadataflyout.view.MetadataPanel',
...
```
data-ux: Here same as above.

bundles/catalogue/metadataflyout/view/MetadataPage.js
```diff
@@ -326,7 +335,16 @@ Oskari.clazz.define('Oskari.catalogue.bundle.metadataflyout.view.MetadataPage',
...
 
             return replacedText;
+        },
+
+        /**
+         * @method showErrorMessage
+         * Render 'metadata not found' message to ui
+         */
+        _showMetadataNotFoundMessage: function () {
+            jQuery(this.ui[0]).text(this.locale.notFound);
```
data-ux: `this.ui` is already a jQuery object. Could we just call `.text()` on that directly without making a new jQuery object?

bundles/catalogue/metadataflyout/view/MetadataPage.js
```diff
@@ -49,22 +49,28 @@ Oskari.clazz.define('Oskari.catalogue.bundle.metadataflyout.view.MetadataPage',
...
-                }
 
-                me.addPanel(panel);
-                panel.init(i === 0);
+            if (data.identifications.length === 0) {
+                //  No identifications, show metadata not found message
+                me._showMetadataNotFoundMessage();
+            } else {
+                // Create a panel for each identification
+                for (i = 0; i < data.identifications.length; i += 1) {
```
data-ux: Using `.forEach` for iterating an Array would be less error prone, because you don't need to keep track of the index. The function given to ´.forEach` is called with the index as the second argument, if the index is needed in the logic.

bundles/catalogue/metadataflyout/view/MetadataPanel.js
atlain: Yep, that is better.

PR #621, opened 2018-10-24
Added option for Cesium styling language

jheijari:
Now it is possible to style 3DTiles layer with Cesium 3DTiles style https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification/Styling

### DIFF
--- a/bundles/paikkatietoikkuna/mapmodule/mapmodule.olcs.js
+++ b/bundles/paikkatietoikkuna/mapmodule/mapmodule.olcs.js
@@ -329,19 +329,44 @@ class MapModuleOlCesium extends MapModuleOl {
     }
 
     /**
+     * Returns color expressions modified with layer opacity
+     * @param {String | Object} colorDef Cesium style expression or object containing conditions array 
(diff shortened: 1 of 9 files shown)


bundles/paikkatietoikkuna/mapmodule/mapmodule.olcs.js
```diff
@@ -328,20 +328,45 @@ class MapModuleOlCesium extends MapModuleOl {
         return params;
     }
 
+    /**
+     * Returns color expressions modified with layer opacity
+     * @param {String | Object} colorDef Cesium style expression or object containing conditions array 
+     * @param {Number} opacity Layer opacity
+     */
+    _getColorExpressionsWithOpacity (colorDef, opacity) {
+        if (!colorDef || isNaN(opacity)) {
```
data-ux: If `colorDef` is falsy, we return it anyway?

jheijari: Yes. If we can't modify the expression, we return it as is.

PR #623, opened 2018-10-24
Restarting timeseries

jheijari:
When pressing play on thematic maps timeseries, restart animation from the beginning, if we are at the end of the timeline.


bundles/statistics/statsgrid2016/service/SeriesService.js
```diff
@@ -48,7 +48,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.SeriesService',
         getSelectedIndex: function () {
             return this.values.indexOf(this.selectedValue);
         },
-        getValue: function () {
+        getValue: function (index) {
```
ZakarFin: The parameter isn't used?

PR #628, opened 2018-10-25
Coordtrans localization and ui changes

okauppinen:
Updated localization files. 

Changed ui order. Now coordinate system selections are made before source selections. Flyout's max height is calculated based on the map's height. Decimal count changed to metric precision and precision table added to info popup. Hide pointless export options if transform isn't file to file. 

If file is selected before crs settings and contains z-values input table didn't show z-values when crs settings are selected. Now input dimension is forced to 3 if crs settings aren't selected. This affects only for reading file to input table (not for transformation).

Added reversed epsg-codes (same values but reversed axis order). Also added replaced GK-codes. These are not listed in the dropdown but can be selected with find/filter with EPSG-code. Frontend uses epsg code listed in dropdown and replaces it to transformation params.

### DIFF
--- a/bundles/paikkatietoikkuna/coordinatetransformation/resources/locale/fi.js
+++ b/bundles/paikkatietoikkuna/coordinatetransformation/resources/locale/fi.js
@@ -5,7 +5,7 @@ Oskari.registerLocalization(
     "value": {
         "title": "Koordinaattimuunnos",
         "tile": {
-            "title": "Koordinaattimuunnos"
+            "title": "Koordinaatti- muunnos"
         },
         "flyout": {
             "title":"Koordinaattimuunnos",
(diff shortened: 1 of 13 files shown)


bundles/paikkatietoikkuna/coordinatetransformation/resources/locale/fi.js
```diff
@@ -5,7 +5,7 @@ Oskari.registerLocalization(
     "value": {
         "title": "Koordinaattimuunnos",
         "tile": {
-            "title": "Koordinaattimuunnos"
+            "title": "Koordinaatti- muunnos"
```
ZakarFin: Is this correct? Feels like it could be some erronous copypaste change.

PR #634, opened 2018-10-26
Configurable public path prefix for Webpack build

data-ux:
Enable absolute path with `--env.absolutePublicPath=true`


ZakarFin:
Nice, add a mention about this in https://github.com/oskariorg/oskari-frontend/blob/develop/README.md so it's documented.

PR #636, opened 2018-10-29
Cleaning repository

jheijari:
Remove unmaintained bundles and unsupported Openlayers 2 from the codebase.


ZakarFin:
We need to first setup a community-repository where we can host these. And instructions how to setup an app that uses code from the community-repository.

data-ux:
We should still keep framework/download-basket
Otherwise deleted (but still valuable) bundles have now been moved to contrib-repo

ZakarFin:
- bundle.js for download-basket under packages is still marked for deletion (shouldn't be).
- We also need to keep framework/selected-featuredata (or maybe move to oskari-frontend-contrib under tampere). 
- Not sure about libraries/ol3/oskari - are these handled in some other way now? They were required for earlier ol3+ builds. 
- The mapstats has been removed so packages/mapping/ol3/deprecated/mapstats/bundle.js can be removed as well (and any files it references)

data-ux:
framework/selected-featuredata moved to contrib under tampere

PR #640, opened 2018-10-29
Don't use ES6 in dynamically loaded js

jheijari:


ZakarFin:
Some of the code is not run through babel yet (namely admin-bundles) so we need to use the methods available in browsers (IE).

PR #641, opened 2018-10-29
Webpack copy appresources

jheijari:
Copy everything under application/ to build


ZakarFin:
Application folder can have png's etc (sample app has logo.png) that can be referenced in the app, but were not copied to the dist-folder.

PR #646, opened 2018-10-30
Dynamic import

data-ux:
Replace requirejs bundle loader with dynamic import based.

Use `"lazy": true` in minifierAppSetup.json to configure bundle to be lazy loaded.

After this change, all bundles used in the app must be referenced in minifierAppSetup.json. Bundles cannot be loaded based only on DB defined path or path given in `Oskari.app.playBundle(...)`.

Also cleaned up comments referencing to `Import-Bundle`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ This repository holds the front-end code for Oskari. Developing the front-end re
 Make sure you have at least Node 8.x / NPM 5.x. Run `npm install` in the front-end repository root.
 Make sure you have oskari-server running on localhost port 8080 (can be customized on webpack.config.js).
 
+### App composition
+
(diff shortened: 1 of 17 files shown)


ZakarFin:
Nice, looks good, but I think at least loader.js should mention that the bundles are not loaded in any way anymore (or that it has been offloaded to webpack) and README should clearly state that all the bundles the application NEEDS to load at ANY point should be referenced in the minifierAppSetup,json. Also a note about having optional bundles that most people don't see (like admin bundles) to have the lazy: true flag enabled.

ZakarFin:
After this only admin-layerselector uses require.js and once that's dealt with we can remove require.js from the core as well?

PR #648, opened 2018-11-02
Statsgrid indicator form fixes

okauppinen:
Prevent KeyboardPan and KeyboardZoom and fix replace commas on table cells input. 
Assign space for empty contenteditable (br-tag and min-height). Fixes the issue where user can't activate contenteditable on click. Check that own indicator's year is number. Add only valid rows.

UI changes:
- more space for region column (input indicator data)
- disable textarea resize
- remove diagram flyout's scroll bar
- fix own indicator flex container in IE
- set max-width for indicator select 
- allow dropdown to flow over flyout

### DIFF
--- a/bundles/statistics/statsgrid2016/components/IndicatorDataForm.js
+++ b/bundles/statistics/statsgrid2016/components/IndicatorDataForm.js
@@ -19,7 +19,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorDataForm', function (l
                             '</div>'),
         row: _.template('<tr data-id="${regionId}">' +
                             '<td class="region" style=" border: 1px solid black ;">${regionName}</td>' +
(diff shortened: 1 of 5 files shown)


bundles/statistics/statsgrid2016/components/IndicatorDataForm.js
```diff
@@ -163,13 +177,14 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorDataForm', function (l
             // separator can be tabulator, comma or colon
             var matches = line.match(/([^\t;,]+) *[\t;,]+ *(.*)/);
             if (matches && matches.length === 3) {
-                area = matches[1];
+                area = matches[1].trim();
                 value = (matches[2] || '').replace(',', '.').replace(/\s/g, '');
+
+                validRows.push({
+                    'name': area,
```
data-ux: No need for quotes around key names.

PR #653, opened 2018-11-05
Fix duplicate tools in publisher

jheijari:
Fixes issues where user modified published map view containing a stats layer.
- Diagram tool would become enabled even if it wasn't enabled in the publication.
- Diagram and table tools had duplicate buttons.


data-ux:
Looks good.

PR #657, opened 2018-11-06
Remove selected-featuredata from sample app

data-ux:


ZakarFin:
Moved to https://github.com/oskariorg/oskari-frontend-contrib/tree/master/bundles/tampere/selected-featuredata

PR #658, opened 2018-11-07
Divmanazer chart long labels and negative values fix

okauppinen:
Cut long labels. Use different text anchors for negative values. Use smaller margins if data contains both negative and positive values because then labels are rendered inside chart. Render no value string to x origin not to 0. Use biggest/widest value instead of max value to calculate x-labels' widths because with all negative values x-axis labels didn't render.

### DIFF
--- a/bundles/framework/divmanazer/component/Chart.js
+++ b/bundles/framework/divmanazer/component/Chart.js
@@ -32,7 +32,17 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function () {
     },
     chartDimensions: function (sideMargin) {
         var me = this;
-        var margin = sideMargin ? Math.min(sideMargin, 140) : 80;
+        var margin = isNaN(sideMargin) ? 80 : Math.min(sideMargin, 140);
+        var dataset = this.getDatasetMinMax();
+        // dataset has both negative and positive values, labels are inside chart -> use smaller margin
+        if (dataset.min < 0 && dataset.max > 0) {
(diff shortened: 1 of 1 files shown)


bundles/framework/divmanazer/component/Chart.js
```diff
@@ -254,7 +268,7 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function () {
      * @param { Object } options keys: colors -> color scale, valueRenderer -> function for rendering bar values
      */
     createBarChart: function (data, options) {
-        if (data != undefined && this.svg === null) {
+        if (data !== undefined && this.svg === null) {
```
data-ux: Changing the equals check to strict, will allow `null` value to pass to `handleData()`. Maybe this is now what we want?

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -380,7 +412,7 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function () {
      * @method createLineChart
      */
     createLineChart: function (data, options) {
-        if (data != undefined) {
+        if (data !== undefined) {
```
data-ux: Same thing with the strict equals as above.

bundles/framework/divmanazer/component/Chart.js
```diff
@@ -436,7 +478,7 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function () {
      */
     redraw: function (data, options) {
         var chart;
-        if (data != undefined) {
+        if (data !== undefined) {
```
data-ux: And here.

PR #659, opened 2018-11-07
Download basket ol4

kuosman:
This PR fix https://github.com/oskariorg/oskari-docs/issues/75 issue.
All ol2 specific code is not used anymore.

Includes also improvements (used Finnish transport agency):
* added optional configuration so licence url and name can also be a layer specific
* added optional configuration to add privacy policy url


ZakarFin:
Thanks!

PR #660, opened 2018-11-07
Drawtools fix

atlain:
Fixes an issue where drawing a box, square or lineString unnecessarily removes a coordinate point from the drawing when modifyControl is set to false.


ZakarFin:
Fixes oskariorg/oskari-docs#79

PR #661, opened 2018-11-07
Fix WFS layer initial appearing

data-ux:
And some lint fixes


ZakarFin:
Fixes an issue where WFS-features were not drawn on the map before user interacted with the map even though said features were available on the browser.

PR #671, opened 2018-11-13
Copy mapmodule images to assets

data-ux:


ZakarFin:
It's a nice solution for the problem, but I think we need to fix the core issue of having image references in js instead of working around the issue with custom build behavior for one bundle.

PR #672, opened 2018-11-13
Load mapmodule images in Webpack build

data-ux:


ZakarFin:
Note! Changes mapmodule.getImageUrl() logic and assumes the img-name being passed as param.

PR #673, opened 2018-11-13
Remove code moved into contrib repo

data-ux:


ZakarFin:
These files are now in https://github.com/oskariorg/oskari-frontend-contrib so this prevents accidental edits on files that should be maintained on that repository.

PR #674, opened 2018-11-13
Copy resources dir to dist/$VERSION/

data-ux:


ZakarFin:
Enables builds in app-specific/contrib repositories to refer to oskari-frontend resources like portal.css etc. Needs update on HTML from 

    <link rel="stylesheet" type="text/css" href="/Oskari/resources/css/portal.css">

To:

    <link rel="stylesheet" type="text/css" href="/Oskari/dist/[version]/resources/css/portal.css">

PR #675, opened 2018-11-14
Attribution for vector tile layers

jheijari:
Enables attribution control for vector tile layers.

### DIFF
--- a/api/mapping/mapmodule/vectortilelayerplugin.md
+++ b/api/mapping/mapmodule/vectortilelayerplugin.md
@@ -93,7 +93,7 @@ And each key has object value with one or both keys `featureStyle`, `optionalSty
 Hover describes how to visualize features on mouse hover and what kind of tooltip should be shown.
 Hover has two optional keys `featureStyle` and `content`.
 
-Content should be content of tooltip a as an array. Each object creates a row to the tooltip.
+Content should be content of tooltip as an array. Each object creates a row to the tooltip.
 Each row object has `key` or `keyProperty` and `valueProperty`.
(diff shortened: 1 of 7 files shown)


api/mapping/mapmodule/vectortilelayerplugin.md
```diff
@@ -88,6 +88,44 @@ And each key has object value with one or both keys `featureStyle`, `optionalSty
 }
 ```
 
+### Hover
+
+Hover describes how to visualize features on mouse hover and what kind of tooltip should be shown.
+Hover has two optional keys `featureStyle` and `content`.
+
+Content should tooltip's content as an array. Each object creates a row to the tooltip.
```
data-ux:
```suggestion
Content should be content of tooltip as an array. Each object creates a row to the tooltip.
```

data-ux:
Found type in docs. Otherwise good to go!

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -1017,6 +1017,26 @@ Oskari.clazz.define(
...
+            }
+            if (!Array.isArray(attributions)) {
+                attributions = [attributions];
+            }
+            return attributions.map(obj => {
+                if (typeof obj === 'string') {
+                    return obj;
+                }
+                const { label, link } = obj;
+                return link ? `<a href="${link}">${label}</a>` : label;
```
ZakarFin: Move UI specific formatting out of AbstractLayer to VectorTileLayerPlugin.js or even AbstractMapLayerPlugin.js

PR #678, opened 2018-11-15
App separation fixes

ZakarFin:
Packages jQuery and the common css-files to oskari.min.js so they don't have to be separately inserted to HTML. Fixes an issue where an app outside oskari-fronted (like one in oskari-frontend-contrib)  needed to refer to files in the app-specific folder for most part, but still have some of the files from oskari-frontend manually injected for the built application to work properly. Also tries to fix an issue with require.js and the maplayer admin functionality that has the same issue (AMD-modules and template-files need to be injected on the app-build instead of having them just on the oskari-frontend side).

Some files needed to be updated from using $() to jQuery() as that's the global that build-script is configured to expose.

Note! This needs some changes on the server-side as well. Namely references to /Oskari/resources/css and jQuery need to be removed from the page: oskariorg/oskari-server#286


data-ux:
Updated package-lock.json probably has newer version of eslint-standard than before. Resulting in eslint errors where previously were none.

ZakarFin:
Looks like the old one accepted trailing spaces on comment rows, but the new one doesn't.

PR #683, opened 2018-11-20
Documentation changes & small fixes

ZakarFin:
- Update readme to reflect latest build-script changes
- Replace some $ references to jQuery (as jQuery is now only exposed with the name)
- Add progress indicator for build
- Add placeholder for the default version of sample app for easier reference
- Change admin-layerselector location in build to version/assets as the bundle assumes after #678 instead of directly under version.

Sorry. Too many unrelated changes for a single PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,16 @@
 
 Documentation available at [http://www.oskari.org].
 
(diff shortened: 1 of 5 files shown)


data-ux:
Looks good!

PR #684, opened 2018-11-20
Remove download-basket bundle

data-ux:
(moved to contrib repo)


ZakarFin:
oskariorg/oskari-frontend-contrib#3

PR #687, opened 2018-11-22
Publisher: unsupported projection

data-ux:
Remove layers with unsupported projection when enterin publisher and restore when exiting.

Shows reason why layer is unsupported in Create Embedded Map flyout. Reason can be either "no permission" or "unsupported projection".

### DIFF
--- a/bundles/mapping/mapmodule/AbstractMapModule.js
+++ b/bundles/mapping/mapmodule/AbstractMapModule.js
@@ -2280,7 +2280,7 @@ Oskari.clazz.define(
                 layerFunctions = [],
                 sandbox = this.getSandbox();
 
-            if (!layer.isSupported(sandbox.getMap().getSrsName())) {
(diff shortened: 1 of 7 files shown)


bundles/mapping/mapmodule/AbstractMapModule.js
```diff
@@ -2280,7 +2280,7 @@ Oskari.clazz.define(
                 layerFunctions = [],
                 sandbox = this.getSandbox();
 
-            if (!layer.isSupported(sandbox.getMap().getSrsName())) {
+            if (!layer.isSupported(sandbox.getMap().getSrsName()) && sandbox.mapMode !== 'mapPublishMode') {
```
ZakarFin: I'd really like to get rid of sandbox.mapMode as it's just some random flag the publisher uses without any proper documentation or anything. What happens without this?

PR #697, opened 2018-11-27
MapBox styles support for MVT layers

jheijari:
Added external style support for MVT layers.
Styles should be defined in layer's options with key `externalStyles`.


bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileModelBuilder.js
```diff
@@ -1,16 +1,26 @@
-
 const Style = Oskari.clazz.get('Oskari.mapframework.domain.Style');
 
 export default class VectorTileModelBuilder {
     parseLayerData (layer, mapLayerJson, maplayerService) {
-        if (mapLayerJson.options.styles) {
-            Object.keys(mapLayerJson.options.styles).forEach((styleName) => {
+        const { options } = mapLayerJson;
+        if (options) {
```
data-ux: Early return better?

PR #709, opened 2018-12-03
MouseHoverEvent add isDrawing

okauppinen:
Add isDrawing to MouseHoverEvent and don't hover vectorlayer while drawing. If it's fine to add isDrawing to event, then API documentation should be updated.

MarkersPlugin: change marker's text stroke width to 3


ZakarFin:
I think it's ok to add the info to the event. Can you update the docs as well to the same PR?

PR #712, opened 2018-12-04
Remove bundle that has been moved to contrib-repository

ZakarFin:


data-ux:
Just a note: files under  bundles/integration/backbone/ haven't been moved into contrib.

ZakarFin:
True, but backbone isn't used anywhere. Same with integration/todo.

PR #720, opened 2018-12-08
Require.js config for Webpack-build

ZakarFin:
To make hierarchical layerselector admin compatible with Webpack build.


bundles/admin/admin-hierarchical-layerlist/components/Layer.js
```diff
@@ -62,6 +62,8 @@ Oskari.clazz.define('Oskari.admin.hierarchical-layerlist.Layer', function (insta
                 '_bundle': '../../../Oskari/bundles/integration/admin-layerselector'
             }
         };
+        // eslint-disable-next-line
+        requirementsConfig.paths._bundle = __webpack_public_path__ + 'assets/admin-layerselector';
```
data-ux: Why not put it directly on line 62?

PR #722, opened 2018-12-08
Fix layer filtering on hierarchical layerselector

ZakarFin:
Previously caused JS-error as groups don't have getLayers() function.

### DIFF
--- a/bundles/framework/hierarchical-layerlist/Flyout.js
+++ b/bundles/framework/hierarchical-layerlist/Flyout.js
@@ -171,7 +171,8 @@ Oskari.clazz.define('Oskari.framework.bundle.hierarchical-layerlist.Flyout',
             var layersCopy = layers.slice(0);
             var notLoadedBackend = {};
 
-            // TODO: is this needed anymore? Now that groups are coming with the layers...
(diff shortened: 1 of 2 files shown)


bundles/framework/hierarchical-layerlist/Flyout.js
```diff
@@ -162,15 +162,16 @@ Oskari.clazz.define('Oskari.framework.bundle.hierarchical-layerlist.Flyout',
...
+            var me = this;
+            var groupList = [];
 
             var allGroups = (me._currentFilter) ? me.mapLayerService.getFilteredLayerGroups(me._currentFilter) : me.mapLayerService.getAllLayerGroups();
 
             var layers = (me._currentFilter) ? me.mapLayerService.getFilteredLayers(me._currentFilter) : me.mapLayerService.getAllLayers();
             var layersCopy = layers.slice(0);
             var notLoadedBackend = {};
 
+            // TODO: is this needed anymore? Now that groups are coming with the layers...
```
data-ux: It would be preferable to answer this question and update the code instead of adding a TODO.

PR #723, opened 2018-12-12
Feature/resizable diagram

atlain:
Enables resizing thematic maps diagram flyout


bundles/framework/divmanazer/component/Chart.js
```diff
@@ -15,6 +15,7 @@ Oskari.clazz.define('Oskari.userinterface.component.Chart', function () {
     this._options = {
         colors: ['#555', '#555']
     };
+    this.resizable = false;
```
ZakarFin: Add a setter for this so and update references to use the function instead of toggling an "internal" variable.

PR #724, opened 2018-12-12
Revert "Fix legend rendering on load"

ZakarFin:
Reverts oskariorg/oskari-frontend#715 as it introduced new issues with the legend.


ZakarFin:
Nevermind, the fix is still better to have even with the new issues.

PR #726, opened 2018-12-12
MVT WFS event/request API implementation part 1

data-ux:
Implements all requests & events described in the bundle.md except `WFSSetFilter`, which has been separated into an upcoming PR. The signature of method `_applyInExtent` looks funny without the code that actually needs this structure.

Implemented `FeatureExposingMVTSource` for retrieving the feature properties of MVT features in view.


ZakarFin:
Just to remind myself for releasenotes. There's an Openlayers 5.2.0 -> 5.3.0 upgrade hidden in this PR.

PR #727, opened 2018-12-13
MVT WFS event/request API implementation part 2

data-ux:
Added handling for `WFSSetFilter` event for feature selection by intersecting geometry.

The Open Layers MVT source uses internally RenderFeature class to represent features in tiles. In order to test for intersection these RenderFeatures need to be converted to regular Feature class instances. That happens in `convertRenderFeatures`.

### DIFF
--- a/bundles/mapping/mapwfs2/plugin/components/convertRenderFeatures.js
+++ b/bundles/mapping/mapwfs2/plugin/components/convertRenderFeatures.js
@@ -1,0 +1,101 @@
+import GeometryType from 'ol/geom/GeometryType';
+import {linearRingIsClockwise} from 'ol/geom/flat/orient';
+import LineString from 'ol/geom/LineString';
+import MultiLineString from 'ol/geom/MultiLineString';
+import MultiPoint from 'ol/geom/MultiPoint';
+import MultiPolygon from 'ol/geom/MultiPolygon';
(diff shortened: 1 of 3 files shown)


bundles/mapping/mapwfs2/plugin/components/convertRenderFeatures.js
```diff
@@ -0,0 +1,104 @@
...
+        if (transformer) {
+            flatCoordinates = transformer(flatCoordinates.slice());
+        }
+        /**
+         *  Geometry conversion algorithm adapdted from
+         *  https://github.com/openlayers/openlayers/blob/e426af29c20c7b08ee323b2ee0f78d34b61c4c34/src/ol/format/MVT.js#L186-L211
+         * */
+        let geom;
+        if (geometryType === GeometryType.POLYGON) {
+            const endss = [];
```
ZakarFin: More descriptive names that could be used instead of "end", "ends" and "endss"? Or is the point that this is copy/paste from OL and it's easier to track changes if we don't modify it?

data-ux: Yes, the point is that if the variable names are the same as in the OL source, it's easier to see that the code is the same.

PR #728, opened 2018-12-14
Simplify case square

atlain:
Separate square shape handling from circle. Add features to _featuresValidity object so that area is calculated in sumMeasurements function.


atlain:
Accidentally pushed "Prevent empty metadata search" to this branch while this pr was open. Don't mind it.

PR #729, opened 2018-12-18
Utilize class division in diagram colors

atlain:
Utilizes indicators class division in thematic maps diagram colors

### DIFF
--- a/bundles/statistics/statsgrid2016/components/Diagram.js
+++ b/bundles/statistics/statsgrid2016/components/Diagram.js
@@ -248,14 +248,22 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Diagram', function (service, lo
      * @return colors[] containing colors
      */
     getColorScale: function () {
-        /*
+        var me = this;
         var stateService = this.service.getStateService();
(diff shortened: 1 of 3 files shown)


bundles/statistics/statsgrid2016/components/Diagram.js
```diff
@@ -248,14 +248,22 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Diagram', function (service, lo
...
      */
     getColorScale: function () {
-        /*
+        var me = this;
         var stateService = this.service.getStateService();
         var activeIndicator = stateService.getActiveIndicator();
         var classificationOpts = stateService.getClassificationOpts(activeIndicator.hash);
+        var currentRegionset = stateService.getRegionset();
+        var classification = {};
+        this.service.getIndicatorData(activeIndicator.datasource, activeIndicator.indicator, this.service.getIndicatorData(activeIndicator.datasource,
```
ZakarFin: The third param is this.service.getIndicatorData(). That seems erronous. Also getIndicatorData() is async although in practice it's probably always sync at this point. Would it be possible to pass the data as param to this method?

PR #740, opened 2019-01-02
Fix thematic maps legend

atlain:
Fixes an issue where thematic maps legend disappears when changing areal division

### DIFF
--- a/bundles/statistics/statsgrid2016/components/Legend.js
+++ b/bundles/statistics/statsgrid2016/components/Legend.js
@@ -21,6 +21,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function (sandbox, loc
         panels: {}
     };
     this._accordion = Oskari.clazz.create('Oskari.userinterface.component.Accordion');
+    this._renderQueue = [];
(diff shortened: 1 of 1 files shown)


bundles/statistics/statsgrid2016/components/Legend.js
```diff
@@ -163,15 +157,16 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Legend', function (sandbox, loc
      * Triggers a new render when needed (if render was called before previous was finished)
      */
     _renderDone: function () {
+        var me = this;
+        if (me._renderQueue.length) {
+            let render = me._renderQueue.shift();
```
data-ux: Do we need to render all requests in the queue or would just the most recent one be enough?

atlain: Good point, the most recent one probably is enough. I'll test it out.

atlain: Turns out all the requests in the queue have to be rendered. Otherwise there might be some indicators missing from the legend.

PR #744, opened 2019-01-03
Filter functionality to MVT WFS layers

jheijari:
Select features based on properties on MVT WFS

### DIFF
--- a/bundles/mapping/mapwfs2/plugin/components/ReqEventHandler.js
+++ b/bundles/mapping/mapwfs2/plugin/components/ReqEventHandler.js
@@ -1,5 +1,5 @@
 import olLayerVectorTile from 'ol/layer/VectorTile';
-import {propsAsArray, WFS_ID_KEY} from './propertyArrayUtils';
+import {propsAsArray, WFS_ID_KEY, WFS_FTR_ID_KEY} from './propertyArrayUtils';
 
 export default class ReqEventHandler {
     constructor (sandbox) {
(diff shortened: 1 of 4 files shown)


bundles/mapping/mapwfs2/plugin/components/ReqEventHandler.js
```diff
@@ -60,6 +112,34 @@ export default class ReqEventHandler {
...
+                if (!layer) {
+                    return;
+                }
+                const records = layer.getActiveFeatures();
+                if (!records || records.length === 0) {
+                    return;
+                }
+                const fields = layer.getFields();
+                const idIndex = fields.indexOf(WFS_FTR_ID_KEY);
+                let filteredIds = [];
```
data-ux: We could use a `Set` instead to avoid deduplication step later.

bundles/mapping/mapwfs2/plugin/components/ReqEventHandler.js
```diff
@@ -60,6 +112,34 @@ export default class ReqEventHandler {
...
+                // handle filter order (AND/OR)
+                const filterOrder = orderFilterOperations(event.getFilters().filters);
```
data-ux: `orderFilterOperations` returns a list of OR groups. We could reflect this in variable naming. So when we do `.forEach` later it will be clear what we are iterating over.

bundles/mapping/mapwfs2/plugin/components/ReqEventHandler.js
```diff
@@ -13,12 +13,64 @@ export default class ReqEventHandler {
...
+                }
+                if (filter.boolean === 'AND') {
+                    attributeFilters.push(filter);
+                } else if (filter.boolean === 'OR') {
+                    attributeFilters = [];
+                    filterOrder.push(attributeFilters);
+                }
+            });
+            return filterOrder;
+        };
```
data-ux: We should move the filtering logic above to a separate file.

bundles/mapping/mapwfs2/plugin/components/propertyArrayUtils.js
```diff
@@ -1,4 +1,5 @@
 export const WFS_ID_KEY = '_oid';
+export const WFS_FTR_ID_KEY = '__fid';
```
data-ux: We could change the server implementation to use `__fid` as the key so that we don't need two constants for the same thing.

PR #751, opened 2019-01-09
WIP: Methods for getting/setting values in oskari style json

atlain:
Enables getting/setting visualization form values in oskari json style.
Comments about the json form are welcome (especially how area/border values and dot color are represented).

### DIFF
--- a/bundles/framework/divmanazer/component/VisualizationForm.js
+++ b/bundles/framework/divmanazer/component/VisualizationForm.js
@@ -130,6 +130,97 @@ Oskari.clazz.define(
         },
 
         /**
+         * @public @method getOskariStyle
+         * Returns Oskari JSON style
+         *
+         * @return {JSON}
+         */
+        getOskariStyle: function () {
+            const values = this.getValues();
+            var oskariStyle = {
+                featureStyle: {
+                    fill: {
+                        color: values.dot.color,
+                        area: {
+                            color: values.area.fillColor,
+                            pattern: values.area.fillStyle
+                        }
+                    },
+                    stroke: {
+                        color: values.line.color,
+                        width: values.line.width,
+                        lineDash: values.line.style,
+                        lineCap: values.line.cap,
+                        lineJoin: values.line.corner,
+                        area: {
+                            color: values.area.lineColor,
+                            width: values.area.lineWidth,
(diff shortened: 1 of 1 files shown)


bundles/framework/divmanazer/component/VisualizationForm.js
```diff
@@ -129,6 +129,97 @@ Oskari.clazz.define(
...
+                    text: {
+                        labelText: values.dot.message
+                    },
+                    image: {
+                        shape: values.dot.shape,
+                        size: values.dot.size
+                    }
+                }
+            };
+            return JSON.stringify(oskariStyle);
```
ZakarFin: This could return the style object instead of a string.

ZakarFin: And maybe just return the featurestyle part. Other than the added parts like lineDash etc is this something that can be passed to mapmodule.getStyle()?

atlain: It seems that lineDash is expected to be an array in mapmodule.getStyle(), but since it is a string in visualization form it cannot be used unless the cases would be transformed somehow. Dot shape and area fill pattern are also problematic.

I was also thinking that maybe getOskariStyle() should return an object with oskari style JSONs separately for dot, line and area styles instead of trying to pack everything together?

bundles/framework/divmanazer/component/VisualizationForm.js
```diff
@@ -129,6 +129,96 @@ Oskari.clazz.define(
...
+         * Returns Oskari JSON style
+         *
+         * @return {Object}
+         */
+        getOskariStyle: function () {
+            const values = this.getValues();
+            var oskariStyle = {
+                fill: {
+                    area: {
+                        color: values.area.fillColor,
```
ZakarFin: In mapmodule getStyle() the fill color is just:

    {
        fill: {
            "color": "#FFBB00"
        },
        ...
    }

https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/mapping/mapmodule/MapModuleClass.ol.js#L861

ZakarFin: After this small change this looks good for merging

PR #733, opened 2018-12-19
Mapmodule user location

okauppinen:
Add location accuracy and error to UserLocationEvent.

Add options to GetUserLocationRequest including addToMap (adds point and accuracy circle to vectorlayer) and PositionOptions for geolocation api (timeout, enableHighAccuracy, maximumAge).

New Start/StopUserLocationTrackingRequests and handlers which uses Geolocation.watchPosition().

Added some functionalities to mapmodule to get geojson objects, track user's location and zoom map by location accuracy.

Changed MyLocationPlugin to use GetUserLocationRequest (with high accuracy option) and listen to UserLocationEvent for errors.

Increased GeolocationPlugins timeout to 30s because Chrome doesn't halt request while asking permission from user.

Note! Start/StopUserLocationTrackingRequests doesn't have API documentation and mylocationplugin doesn't have localization. This PR is made to review code. Missing documentation and localization will be pushed later.

PR #765, opened 2019-01-18
Added support configurable zoom

kuosman:
This PR add support to configure custom zoom level.


ZakarFin:
Thanks!

ZakarFin:
Note that there will be additional changes to this feature in #733

PR #766, opened 2019-01-18
Fix function name

jheijari:


ZakarFin:
Fixes an issue caused in #764

PR #773, opened 2019-01-23
Published map layer style select

data-ux:
User can allow layer style change on published map in the publisher tool for layer menu.

This selection is stored in a new conf flag `isStyleSelectable` for the LayerSelectionPlugin.

### DIFF
--- a/bundles/framework/publisher2/tools/LayerSelectionTool.js
+++ b/bundles/framework/publisher2/tools/LayerSelectionTool.js
@@ -12,12 +12,20 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.LayerSelectionTool',
         ],
         groupedSiblings: false,
         _templates: {
-            backgroundLayerSelector: jQuery('<div class="publisher2 background-layer-selector tool-options"><div class="header"></div><div class="layers"></div></div>'),
+            extraOptions: jQuery(`
+            <div class="publisher2 background-layer-selector tool-options">
+                <div class="style-selection">
+                    <label><input type="checkbox"/><span></span></label>
+                </div>
+                <div class="info"></div>
+                <div class="layers"></div>
+            </div>
+            `),
             backgroundCheckbox: jQuery('<div class="background-layers"><label>' +
                 '    <input class="baselayer" type="checkbox"/>' +
                 '  </label><div>')
         },
-        _backgroundLayerSelector: null,
+        _extraOptions: null,
 
         /**
     * Get tool object.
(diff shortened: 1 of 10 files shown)


bundles/framework/publisher2/tools/LayerSelectionTool.js
```diff
@@ -50,6 +58,7 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.LayerSelectionTool',
                     pluginConfig.config.baseLayers = layerSelection.baseLayers;
                     pluginConfig.config.defaultBaseLayer = layerSelection.defaultBaseLayer;
                 }
+                pluginConfig.config.isStyleSelectable = me.__plugin.getStyleSelectable();
```
ZakarFin: Shouldn't this come in the this.getPlugin().getConfig() call above?

bundles/mapping/mapmodule/plugin/layers/LayerSelectionPlugin.js
```diff
@@ -37,6 +37,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.LayerSelectionP
             },
             buttonGroup: 'mobile-toolbar'
         };
+        me._styleSelectable = !!this.getConfig().isStyleSelectable;
```
ZakarFin: What's the difference with getConfig() and the config that is received as parameter? BasicMapModulePlugin handles setting the config that's returned by getConfig()?

bundles/framework/publisher2/tools/LayerSelectionTool.js
```diff
@@ -211,32 +230,43 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.LayerSelectionTool',
      * @return {Boolean} true if layer must be preselect, other false
      */
         shouldPreselectLayer: function (id) {
+            const toolPluginMapfullConf = this._getToolPluginMapfullConf();
+            if (toolPluginMapfullConf) {
+                var isPluginConfig = !!((toolPluginMapfullConf && toolPluginMapfullConf.config &&
+                    toolPluginMapfullConf.config.baseLayers));
+
+                if (isPluginConfig) {
+                    var isFound = jQuery.inArray('' + id, toolPluginMapfullConf.config.baseLayers);
```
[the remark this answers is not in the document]
data-ux: It's old code that was moved around, but yeah we can refactor it while at it.

bundles/framework/publisher2/tools/LayerSelectionTool.js
data-ux: `getConfig()` does not return the changes made in the publisher to the plugin config. I guess it returns the initial config given to the plugin. That's why `baseLayers` and `defaultBaseLayer` is written separately, too, on the above lines.

bundles/mapping/mapmodule/plugin/layers/LayerSelectionPlugin.js
data-ux: Yes, `config` param of constructor is assigned to instance variable `_config` by BasicMapModulePlugin. `getConfig()` reads that.

PR #777, opened 2019-01-29
Use configurable property for highlighting features

jheijari:
Sets _oid as hover property for WFS-MVT. Hover property can be set in VectorTileLayerPlugin's hover state.


bundles/mapping/mapwfs2/plugin/WfsMvtLayerPlugin.js
```diff
@@ -15,6 +15,7 @@ Oskari.clazz.defineES('Oskari.wfsmvt.WfsMvtLayerPlugin',
             this._clazz = 'Oskari.wfsmvt.WfsMvtLayerPlugin';
             this._log = Oskari.log('WfsMvtLayerPlugin');
             this.layertype = 'wfs';
+            this.hoverState.property = '_oid';
```
ZakarFin: Add a comment why we want to do this
```suggestion
            // mvt only support numeric IDs and WFS-layers often have other characters in ID as well
            // fixes highlight on features spread to multiple tiles by using a generated _oid for "combining" features
            this.hoverState.property = '_oid';
```

PR #779, opened 2019-01-30
Webpack & ESLint support for React

data-ux:


ZakarFin:
After this we can start using React in bundles. Doesn't really add anything but the config and the dependency to React.

PR #782, opened 2019-01-31
Wfs mvt own style

jheijari:
Adds support for user's custom style to MVT implementation of WFS layer.


bundles/mapping/mapwfs2/request/ShowOwnStyleRequestHandler.js
```diff
@@ -15,6 +15,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapwfs2.request.ShowOwnStyleRequ
         this.plugin = plugin;
         this.localization = plugin.getLocalization('popup');
         this.visualizationForm = this.plugin.getVisualizationForm();
+        this._usesMvtPlugin = this.plugin._clazz === 'Oskari.wfsmvt.WfsMvtLayerPlugin';
```
data-ux: We could move the implementtation specific behaviour to the implementing plugins to avoid this.

bundles/mapping/mapwfs2/plugin/WfsMvtLayerPlugin.js
```diff
@@ -27,13 +29,45 @@ Oskari.clazz.defineES('Oskari.wfsmvt.WfsMvtLayerPlugin',
...
+         * Returns JSON presentation of bundles localization data for
+         * current language. If key-parameter is not given, returns
+         * the whole localization data.
+         *
+         * @param {String} key (optional) if given, returns the value for key
+         * @return {String/Object} returns single localization string or
+         *      JSON object for complete data depending on localization
+         *      structure and if parameter key is given
+         */
+        getLocalization (key) {
```
data-ux: We could refactor ShowOwnStyleRequestHandler.js so that we wouldn't need this method.

bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -905,25 +905,49 @@ export default class MapModule extends AbstractMapModule {
...
-        if (styleDef.stroke.width) {
-            stroke.width = styleDef.stroke.width;
+        if (width) {
+            stroke.width = width;
         }
-        if (styleDef.stroke.lineDash) {
-            if (_.isArray(styleDef.stroke.lineDash)) {
-                stroke.lineDash = styleDef.stroke.lineDash;
+        if (lineDash) {
+            if (_.isArray(lineDash)) {
```
ZakarFin: Could we use Array.isArray() instead so we don't have dependency to lodash?

PR #786, opened 2019-02-06
Fix layer order after transition from/to publisher

data-ux:
Fixed timing dependant problem with restoring layer order after 200ms from `UIChangeEvent`.

### DIFF
--- a/bundles/framework/publisher2/view/PublisherSidebar.js
+++ b/bundles/framework/publisher2/view/PublisherSidebar.js
@@ -518,6 +518,11 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PublisherSidebar
                 var event = Oskari.eventBuilder('Publisher2.ToolEnabledChangedEvent')(tool);
                 sb.notifyAll(event);
             });
+
+            // fix layer order after changing from/to publisher
+            setTimeout(() => {
+                sb.findRegisteredModuleInstance('MainMapModule').ensureLayerOrder();
+            }, 200);
         },
 
         /**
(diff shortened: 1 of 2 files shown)


bundles/mapping/mapmodule/AbstractMapModule.js
```diff
@@ -293,11 +293,15 @@ Oskari.clazz.define(
...
             AfterRearrangeSelectedMapLayerEvent: function (event) {
-                this.afterRearrangeSelectedMapLayerEvent(event);
+                this.afterRearrangeSelectedMapLayerEvent();
             },
             MapSizeChangedEvent: function (evt) {
                 this._handleMapSizeChanges({width: evt.getWidth(), height: evt.getHeight()});
             },
+            UIChangeEvent: function (evt) {
+                // fix layer order after changing from/to publisher
+                setTimeout(() => this.afterRearrangeSelectedMapLayerEvent(), 200);
```
ZakarFin: Can't this be handled in publisher.setEnabled()?

bundles/framework/publisher2/view/PublisherSidebar.js
```diff
@@ -518,6 +518,11 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PublisherSidebar
                 var event = Oskari.eventBuilder('Publisher2.ToolEnabledChangedEvent')(tool);
                 sb.notifyAll(event);
             });
+
+            // fix layer order after changing from/to publisher
+            setTimeout(() => {
+                sb.findRegisteredModuleInstance('MainMapModule').ensureLayerOrder();
```
ZakarFin: Does this really need the timeout? Here's a function that also does something with the layers so maybe move the ordering stuff here https://github.com/oskariorg/oskari-frontend/blob/179971a84ec2f93c10c3d70c1bbf20138f2e6346/bundles/framework/publisher2/instance.js#L267 or even publisher2/service/PublisherService.js

What makes the layers go out of order?

PR #787, opened 2019-02-07
Send requests in batches

jheijari:
Fixes an issue where service returned 403 Forbidden for sending too many requests at once.
Send only two parallel requests.

### DIFF
--- a/bundles/statistics/statsgrid2016/view/SearchFlyout.js
+++ b/bundles/statistics/statsgrid2016/view/SearchFlyout.js
@@ -352,8 +352,6 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
 
         // Run the searches to see if we get data from the service.
         const batchSize = 2;
-        let batchCounter = 0;
-        let batchIndex = 0;
         const batches = [];
         let batch;
         searchValues.forEach((search, index) => {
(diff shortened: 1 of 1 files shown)


bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -351,22 +351,48 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
-                let counter = 0;
-                const enoughData = !!Object.values(data).find(val => !isNaN(val) && ++counter > 1);
-                if (!enoughData) {
-                    searchFailed(search);
-                    return;
-                }
-                searchSuccessfull(search);
-            });
+        const batchSize = 2;
+        let batchCounter = 0;
```
data-ux: The variable `batchCounter` keeps track of received results, but the name implies it counts batches. Maybe rename?

bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -351,22 +351,48 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
+                batch = [];
+                batches.push(batch);
+            }
+            batch.push(search);
         });
+        const nextBatch = () => {
+            batchIndex++;
+            batchCounter = 0;
+            if (batchIndex < batches.length) {
+                consumeBatch(batchIndex);
```
data-ux: Instead of using `batchIndex` to access the batch, we could just `.pop()` the next one off `batches` and give it as argument to `consumeBatch()`. This way we coud get rid of `batchIndex` variable.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -351,22 +351,48 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
+            }
+        };
+        const consumeBatch = () => {
+            batches[batchIndex].forEach((search, index) => {
```
data-ux: Expanding on the comment above, if we had the current batch as argument, we could get rid of `batchCounter` variable by using `Promise.all()`. Or it that's too much, at least we could make `batchCounter` local variable of `consumeBatch()`.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -351,22 +351,48 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
+                    const enoughData = !!Object.values(data).find(val => !isNaN(val) && ++counter > 1);
+                    if (!enoughData) {
+                        searchFailed(search);
+                        return;
+                    }
+                    searchSuccessfull(search);
+                });
+            });
+        };
+        consumeBatch();
```
data-ux: In the above model we would start by calling `nextBatch`.

PR #788, opened 2019-02-08
Update measurement result

atlain:
Update measurement result label when modifying polygon or linestring

### DIFF
--- a/bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
+++ b/bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
@@ -554,15 +554,6 @@ Oskari.clazz.define(
             if (showMeasureUI) {
                 data.showMeasureOnMap = showMeasureUI;
             }
-
-            if (me._sketch && (me.getCurrentDrawShape() === 'Polygon' || me.getCurrentDrawShape() === 'LineString')) {
-                // Update measurement result in tooltip if drawing polygon or linestring
-                // TODO problem occurs when modifying polygon from its original starting point
-                const measurement = me.getCurrentDrawShape() === 'LineString' ? data.length : data.area;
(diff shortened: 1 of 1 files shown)


bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
```diff
@@ -555,6 +555,14 @@ Oskari.clazz.define(
                 data.showMeasureOnMap = showMeasureUI;
             }
 
+            if (me._sketch && (me.getCurrentDrawShape() === 'Polygon' || me.getCurrentDrawShape() === 'LineString')) {
+                // Update measurement result in tooltip if drawing polygon or linestring
+                // TODO problem occurs when modifying polygon from its original starting point
+                const measurement = me.getCurrentDrawShape() === 'LineString' ? data.length : data.area;
+                const drawMode = me.getCurrentDrawShape() === 'LineString' ? 'line' : 'area';
+                const formattedMeasurement = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule').formatMeasurementResult(measurement, drawMode);
```
ZakarFin: Could use this.getMapModule() instead of fetching reference from sandbox.

bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
```diff
@@ -555,6 +555,14 @@ Oskari.clazz.define(
...
+                jQuery('div.' + me._tooltipClassForMeasure + '.' + me._sketch.getId()).html(formattedMeasurement);
```
ZakarFin: Could we do this in https://github.com/oskariorg/oskari-frontend/blob/9bab4ab25102c0c636e8a3fc2f40a771bf06f7cd/bundles/mapping/drawtools/plugin/DrawPlugin.ol.js#L1009. This way we would have a single point where the measurement value is updated.

PR #792, opened 2019-02-15
Add Bing Maps layer plugin

jheijari:
Adds Bing Maps support. Admin can add a new Bing Maps layer with api key.


bundles/framework/mapfull/instance.js
```diff
@@ -177,6 +177,15 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapfull.MapFullBundleInstance',
                     });
                 }
 
+                let bingMapsPlugin = plugins.find(cur => cur.id === 'Oskari.mapframework.mapmodule.BingMapsLayerPlugin');
```
ZakarFin: Lets refactor this to some const automagicPlugins = ['Oskari.mapframework.mapmodule.VectorTileLayerPlugin', 'Oskari.mapframework.mapmodule.BingMapsLayerPlugin'] and loop over that for the automatic addition.

jheijari: Moved automagic plugins to a separate file.

PR #798, opened 2019-02-21
Update measurement result while modifying

atlain:
Update measurement result while modifying

### DIFF
--- a/bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
+++ b/bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
@@ -929,18 +929,16 @@ Oskari.clazz.define(
                 return;
             }
 
-            if (geometry.getCoordinates()[0].length >= 4) {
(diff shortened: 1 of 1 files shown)


bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
```diff
@@ -928,11 +928,21 @@ Oskari.clazz.define(
...
+                // The first call causes an error because the points differ from each other. Hence using try/catch here to avoid breaking the js process
+                try {
+                    if (!isValidOp.isValid(olParser.read(geometry))) {
+                        // lines intersect -> problem!!
+                        currentDrawing.setStyle(me._styles.intersect);
+                        me._featuresValidity[currentDrawing.getId()] = false;
+                        return;
+                    }
+                } catch (e) {
+                    return;
```
data-ux: It would be better to prevent the exception if we know it's going to happen instead of try-catch.

PR #808, opened 2019-02-28
Lock down versions

ZakarFin:
Building with webpack version 4.29.x doesn't work with dynamic imports (webpack/webpack#8656). Locking down the dependencies to workaround this and hopefully reduce these kind of random problems.


ZakarFin:
Rolled back even further to known working version https://github.com/oskariorg/oskari-frontend/commit/08dd49d502ff29146d41564151cea0eb54423b32

PR #813, opened 2019-03-05
Removed contrib bundle user guide

jheijari:
Remove analyse from user guide. It has been moved to the contrib repo.


ZakarFin:
See nls-oskari/pti-frontend#20 for an example how to modify these in extensions.

PR #816, opened 2019-03-12
WIP: React ui components

atlain:
Added antd dependency and initial versions of some of the new components

### DIFF
--- a/bundles/admin/admin-layereditor/components/Button.jsx
+++ b/bundles/admin/admin-layereditor/components/Button.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from 'antd/lib/button';
+import 'antd/lib/button/style/css';
+
+export const ButtonComponent = (props) => {
+    const {text, ...other} = props;
+    return (
+        <Button {...other}>{text}</Button>
+    );
+};
+
+ButtonComponent.propTypes = {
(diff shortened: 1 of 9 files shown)


bundles/admin/admin-layereditor/components/Button.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from 'antd/lib/button';
+import 'antd/lib/button/style/css';
+
+export const ButtonComponent = (props) => {
+    const {text, ...other} = props;
+    return (
+        <Button {...other}>{text}</Button>
```
data-ux: We could pass in `props.children` into the Button instead of `text` to allow more customizable buttons, not just text-buttons. https://reactjs.org/docs/composition-vs-inheritance.html#containment

bundles/admin/admin-layereditor/components/TextAreaInput.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Input from 'antd/lib/input';
+
+export const TextAreaInput = (props) => {
+    const { TextArea } = Input;
```
data-ux: We could move this out of the component scope so that it's done only once; not every render invocation.

PR #818, opened 2019-03-12
Storybook poc

ZakarFin:
- Rollback antd version so it works on NLS registry (need to update this later)
- Renames ButtonComponent to Button for convenience
- Add storybook support https://storybook.js.org/docs/guides/guide-react/

1) Do an npm install
2) npm run storybook
 -> starts a live server for React component preview without the need for oskari-server.
3) Create [something].stories.js files under oskari-frontend/bundles for your components.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,10 @@
+import { configure } from '@storybook/react';
+
+function loadStories() {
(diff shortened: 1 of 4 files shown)


jheijari:
Looks good to me.

PR #819, opened 2019-03-13
Add immutable.js

jheijari:
Using immutable.js makes sure the data structures used by React components won't be mutated elsewhere in the software.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "license": "MIT",
   "dependencies": {
     "antd": "3.15.0",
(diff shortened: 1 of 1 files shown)


data-ux:
Immutability can be best harnessed if the services providing the data to the React views use immutable data structures. I support implementing all future services in this way!

For the immutability library/tool I suggest Immer:
https://github.com/mweststrate/immer

ZakarFin:
We should still work towards this, but the PR is so old I'm closing it for now.

PR #820, opened 2019-03-13
React ui components

atlain:
Added Select and Collapse components
Few fixes for Button and TextAreaInput components mentioned here: https://github.com/oskariorg/oskari-frontend/pull/816


bundles/admin/admin-layereditor/components/Collapse.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Collapse from 'antd/lib/collapse';
```
jheijari: We could name Collapse as AntCollapse and then export our CollapseComponent as Collapse instead.

bundles/admin/admin-layereditor/components/Select.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Select from 'antd/lib/select';
```
jheijari: And do the same here. Rename Select -> AntSelect and export SelectComponent as Select.

bundles/admin/admin-layereditor/components/Button.jsx
```diff
@@ -1,15 +1,12 @@
...
 import 'antd/lib/button/style/css';
 
-export const ButtonComponent = (props) => {
-    const {text, ...other} = props;
-    return (
-        <Button {...other}>{text}</Button>
-    );
-};
+export const Button = (props) => (
+    <AntButton {...props}>{props.children}</AntButton>
```
data-ux: It would be cleaner no to pass `children` twice:
```
export const Button = ({children, ...other}) => (
<AntButton {...other}>{props.children}</AntButton>
);
```

bundles/admin/admin-layereditor/components/Collapse.jsx
```diff
@@ -0,0 +1,21 @@
...
+import 'antd/lib/collapse/style/css';
+
+const AntPanel = AntCollapse.Panel;
+
+export const Collapse = (props) => {
+    const {panels, ...other} = props;
+    return (
+        <AntCollapse {...other}>
+            {panels.map((panel, key) => (
+                <AntPanel {...panel} key={key}>{panel.content}</AntPanel>
```
jheijari: Passing content twice here

jheijari:
Should we add performance optimization by memorizing these components as well?

bundles/admin/admin-layereditor/components/Tabs.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import AntTabs from 'antd/lib/tabs';
+import 'antd/lib/tabs/style/css';
+
+const AntTabPane = AntTabs.TabPane;
+
+export const Tabs = (props) => {
+    const {tabPanes, ...other} = props;
```
data-ux: We should be able to first wrap `AntTabs.TabPane` and then use it via `props.children`. So we can have structure: 
```
<Tabs>
  <TabPane />
  <TabPane />
</Tabs>
```

data-ux:
> Should we add performance optimization by memorizing these components as well?

At this point, we can memoize components that only take primitive values as props. We cannot do it in the general case, because we are dealing with objects that might be mutated.

If we memoize components that take objects as props, and the objects mutate later on, we will miss out on renders that should happen.

bundles/admin/admin-layereditor/components/Tabs.jsx
atlain: Yep, I created TabPane component and modified Tabs to use it.

data-ux:
Should we wrap AntSelect.Option and AntCollapse.Panel, too?

ZakarFin:
> Should we wrap AntSelect.Option and AntCollapse.Panel, too?

Yes, let's be consistent. I have the AntD Steps with Steps.Step in #822

data-ux:
> Should we also use Pane or Panel (my pref) throughout even though AntD doesn't? Now we have both.

In Antd Pane and Panel are different things: TabPane and Collapse.Panel. The naming reflects established convention in UIs. I don't see a reason to give them the same name to different concepts.

ZakarFin:
Seems to me that collapse doesn’t have any controls in most of our cases so it should be pane instead of panel then?

data-ux:
Yes, to paraphrase, a panel is some bounded area (think "control panel" in physical machines) with UI controls. For example an accordion (or "collapse" in And terminology) usually has multiple small bounded areas that can be opened/closed. Scrollable accordion content would be very odd and is seldom used. So "accordion panel" or in Antd "collapse panel" makes sense.

PR #822, opened 2019-03-13
React-based admin wizard

ZakarFin:
Dummy wizard component that can be used when adding layers. The final step is the actual data gathering form that is implemented in an upcoming PR.

TODO: 
- capabilities fetching from server
- localization
- urlinput to have username/password capabilities
- error messaging


data-ux:
General comment about state handling. I would separate all state into only two places. Permanent state (saved to backend) and temporary state (exits only as long as the layer editor form is open). Permanent state would be handled by the service that will be created for admin-layereditor bundle and temporary state would be held as component state of the layer wizard/form (let's see later how up in the tree we need to push the statekeeping). All React components below the single stateful component should be stateless; they just signal to their parents (via function props) if something needs to change in the temporary state.

bundles/admin/admin-layereditor/view/LayerWizard/LayerURLForm.jsx
```diff
@@ -0,0 +1,58 @@
...
+    'WFS': ['3.0']
+};
+
+export class LayerURLForm extends React.Component {
+    constructor (props) {
+        super(props);
+        this.mutator = props.service;
+        this.state = {
+            url: props.layer.url
+        };
```
data-ux: Maximum one component should own the url state. I propose that this component would own the url state and that we make `UrlInput` a controlled component.
https://reactjs.org/docs/lifting-state-up.html

bundles/admin/admin-layereditor/view/LayerWizard.jsx
```diff
@@ -0,0 +1,99 @@
...
+import { LayerURLForm } from './LayerWizard/LayerURLForm';
+import { LayerDetails } from './LayerWizard/LayerDetails';
+import { LayerCapabilitiesListing } from './LayerWizard/LayerCapabilitiesListing';
+import { Button } from '../components/Button';
+import { LayerWizardService } from './LayerWizard/LayerWizardService';
+
+export class LayerWizard extends React.Component {
+    constructor (props) {
+        super(props);
+        this.service = new LayerWizardService(() => this.setState({ layer: this.service.getLayer() }));
```
data-ux: I think we should reserve the name "service" for global singleton objects managing state and server communication. Here we have a "state manager" object that handles only state for this specific form instance. Instead we could implement the _container & presentational component_ pattern:
https://medium.com/@yassimortensen/container-vs-presentational-components-in-react-8eea956e1cea

PR #823, opened 2019-03-14
React implementation of LayerList

jheijari:
First version of the next gen map layer list written using React.
Does not support nested grouping yet.


bundles/framework/layerlist/components/LayerCollapse.jsx
```diff
@@ -0,0 +1,89 @@
...
+import { Alert } from '../../../admin/admin-layereditor/components/Alert';
+
+const getLayers = (layers, otherProps) => layers.map(
+    (lyr, index) => {
+        const props = {
+            model: lyr,
+            even: index % 2 === 0,
+            ...otherProps
+        };
+        return <Layer key={index} {...props}/>;
```
data-ux: We should always use a real identifier as the key if possible. Layer id would be good here.
https://reactjs.org/docs/lists-and-keys.html#keys

bundles/framework/layerlist/components/LayerCollapse.jsx
```diff
@@ -0,0 +1,89 @@
...
+        };
+        const layers = filteredLayers || group.getLayers();
+        const listProps = {
+            dataSource: getLayers(layers, otherProps),
+            bordered: false
+        };
+        const panelProps = {
+            header: group.getTitle(),
+            extra: <Badge {...badgeProps}/>,
+            children: <List {...listProps}/>
```
data-ux: It would be more idiomatic to but <List> inside <Panel> in the returned JSX instead of passing as props.

bundles/framework/layerlist/components/LayerCollapse.jsx
```diff
@@ -0,0 +1,89 @@
...
+    const results = groups.map(group => {
+        const layers = group.getLayers()
+            .filter(lyr => group.matchesKeyword(lyr.getId(), keyword)); // and some other rule?
+        return {group, layers};
+    }).filter(result => result.layers.length !== 0);
+    return results;
+};
+
+export const LayerCollapse = props => {
+    const { groups, filterKeyword, ...rest } = props;
```
data-ux: Maybe destructure already in function signature?
`export const LayerCollapse = ({ groups, filterKeyword, ...rest }) => {...}`

bundles/framework/layerlist/components/LayerCollapse.jsx
```diff
@@ -0,0 +1,89 @@
...
+    if (!Array.isArray(groups) || groups.length === 0) {
+        return getEmptyMessage();
+    }
+    const filtered = getSearchResults(groups, filterKeyword);
+    if (filtered && filtered.length === 0) {
+        return getEmptyMessage(filterKeyword);
+    }
+    const collapseProps = {
+        children: getPanels(groups, filtered, rest),
```
data-ux: Same here with the children.

bundles/framework/layerlist/components/LayerCollapse/LayerTools.jsx
```diff
@@ -0,0 +1,132 @@
...
+    return !!subLayers.find(sub => !!sub.getMetadataIdentifier());
+};
+
+const getInfoIcon = layer => {
+    const classes = ['layer-info'];
+    if (layer.getMetadataIdentifier() || hasSubLayerMetadata(layer)) {
+        classes.push('icon-info');
+    }
+    const props = {
+        className: classes.join(' ')
```
data-ux: If we have only one or two props, it would look cleaner just to type it into the JSX:
`<RightFloatingIcon className={classes.join(' ')} >;`

bundles/framework/layerlist/components/LayerCollapse.jsx
jheijari: Replaced with layer id

bundles/framework/layerlist/components/LayerCollapse.jsx
jheijari: Passing children now in jsx instead of props

bundles/framework/layerlist/components/LayerCollapse.jsx
jheijari: Destructing in function signature done

bundles/framework/layerlist/components/LayerCollapse/LayerTools.jsx
jheijari: Done

bundles/framework/layerlist/view/LayerCollapse/StateHandler.js
```diff
@@ -0,0 +1,106 @@
...
+                // and some other rule?
+            return {group, layers};
+        }).filter(result => result.layers.length !== 0);
+        return results;
+    }
+    addListener (listener) {
+        this.listeners.push(listener);
+    }
+    notify () {
+        this.listeners.forEach(consumer => consumer());
```
data-ux: We should give the current state of the statehandler as argument to consumers, so that they can use it to `setState()` or ReactDOM.render();

bundles/framework/layerlist/view/LayerCollapse/StateHandler.js
```diff
@@ -0,0 +1,106 @@
...
+    constructor () {
+        this.listeners = [];
+        this.selectedLayerIds = [];
+        this.groups = [];
+        this.filterKeyword = null;
+        this.openGroupTitles = null;
+        this.filtered = null;
+        this.map = Oskari.getSandbox().getMap();
+    }
+    getState () {
```
data-ux: Ability to get the internal state should not be needed. We could make it private and only call in `notify(). Storybooks etc. can still use the provate function "illegally", if needed.

PR #824, opened 2019-03-15
React add util.jsx

okauppinen:


src/react/util.jsx
```diff
@@ -0,0 +1,25 @@
...
+            <GenericContext.Consumer>
+                {context => <Component {...props} {...context} />}
+            </GenericContext.Consumer>
+        );
+    };
+}
+
+// re-binds all object methods starting with 'handle',
+// so that methods can be used detached from instance. For use in React
+export function handleBinder (ob) {
```
data-ux: As @ZakarFin suggested, we could add an optional second parameter that would define the prefix. Default value would be "handle", but caller could give some other prefix if needed.

PR #825, opened 2019-03-19
Language selector

data-ux:
General language selector dropdown for Oskari. Implemented as React component in new bundle 'language-selector'. Uses currently avaialble languages in Oskari instance. Bundle setup in DB/appSetup as usual.

**Usage**
Selector can be used in application by adding an HTML element with id "language-selector-root" into the JSP template. Selector will be created inside on bundle startup.

Another way to use the selector is to include one of the React components into any React UI anywhere in the application. The bundle does not need to be added into DB/appSetup in this case. The components are:  `<LanguageChanger \>` that implements the global language change functionality and `<LanguageSelect \>` that is just a presentational language selection menu component.

**Styling**
Selector can be styled by overriding class `oskari-language-selector` descendants in overwritten.css.


bundles/framework/language-selector/instance.js
```diff
@@ -0,0 +1,21 @@
...
+const BasicBundle = Oskari.clazz.get('Oskari.BasicBundle');
+
+Oskari.clazz.defineES('Oskari.bundle.language-selector.instance',
+    class LanguageSelectorBundle extends BasicBundle {
+        constructor () {
+            super();
+            this.__name = 'language-selector';
+        }
+        _startImpl () {
+            const root = jQuery('#language-selector-root');
```
ZakarFin: This be just document.getElementById() since it's the only use of jQuery here?

bundles/framework/language-selector/components/LanguageSelector.jsx
```diff
@@ -0,0 +1,37 @@
...
+    } else if (window.location.href.indexOf('lang=') > -1) {
+        redirect = window.location.href.replace(/lang=[^&]+/, newLang);
+    } else {
+        redirect = window.location.href + '&' + newLang;
+    }
+    window.location.href = redirect;
+};
+
+export const LanguageSelector = () => {
+    const supported = getSupportedLanguages();
```
ZakarFin: Should the supported and current languages be passed as props instead? The only localization requires knowledge of Oskari? Maybe the change handle callback as well? Then this could be used to do any language selection in Oskari. For example some admin for selecting the default instance language etc.

ZakarFin: Maybe even pass the two strings needed from localization so the component could be "Oskari-free" :)

data-ux:
See updated first comment.

PR #832, opened 2019-03-25
Update to latest dependency versions

ZakarFin:
Looks like dynamic import works again with clean install of latest Webpack. CSS-loader has some issues when passing minimise:true as option (https://github.com/webpack-contrib/css-loader/issues/863) so commented it out. Needs some further testing still, but the updated versions worked ok on local laptop.


webpack.config.js
```diff
@@ -71,14 +71,18 @@ module.exports = (env, argv) => {
                     test: /\.css$/,
                     use: [
                         styleLoaderImpl,
-                        { loader: 'css-loader', options: { minimize: true } }
+                        { loader: 'css-loader', options: { } }
+                        // https://github.com/webpack-contrib/css-loader/issues/863
+                        //                        { loader: 'css-loader', options: { minimize: true } }
```
data-ux: Should we just remove it instead of commenting out? It seems the `minimize` option has been removed from the API. If we want to minimize the CSS, we need to start using some other way, like:
https://github.com/webpack-contrib/mini-css-extract-plugin#minimizing-for-production

data-ux:
Tested and runs ok on my dev machine, too.

PR #833, opened 2019-03-26
Publisher tool layout disable clicks

okauppinen:
Disable clicks and close control popups on tool layout edit mode.


ZakarFin:
Looks good. Just fix the linting errors.

PR #835, opened 2019-03-26
Divmanazer use ExtraFlyout with DefaultExtensions

okauppinen:
Use ExtraFlyout with DefaultExtension to get rid of overlapping flyout implementation (DefaultFlyout). Now all flyouts act similarly (dragging, click brings to top).

Other changes:
 - Prevent to drag flyout out of the viewport's top.
 - Trigger UIChangeEvent when Analysis is started to clean UI and select default tool on UIChangeEvent
 - Added flyout to setEl param. So reference to flyout can be stored and html/css class can be added when extension is started.
 - Added instance name/identifier to flyout and flyoutcontent.
 - Added new extension state oskari-closed which can be used to hide flyout without closing extension.
 - adjust z-indexes


bundles/framework/divmanazer/instance.js
```diff
@@ -307,18 +276,16 @@ Oskari.clazz.define('Oskari.userinterface.bundle.ui.UserInterfaceBundleInstance'
             el = null;
             if (flyoutPlugin !== null && flyoutPlugin !== undefined) {
                 flyout = me.createFlyout(extension, flyoutPlugin, count, extensionInfo);
-
+                /* Now uses ExtraFlyout's makeDraggable
```
ZakarFin: Remove commented out code

bundles/framework/divmanazer/instance.js
```diff
@@ -1290,13 +1281,16 @@ Oskari.clazz.define('Oskari.userinterface.bundle.ui.UserInterfaceBundleInstance'
          *
          */
         shuffleZIndices: function (toTop) {
+            toTop.css('z-index',  this.flyoutZIndexBase + Oskari.seq.nextVal());
+            return; //for debugging that is this really needed
```
ZakarFin: Draft stuff

bundles/framework/divmanazer/instance.js
```diff
@@ -1066,6 +1040,10 @@ Oskari.clazz.define('Oskari.userinterface.bundle.ui.UserInterfaceBundleInstance'
             restore: function (flyout, flyoutPlugin, extensionInfo) {
                 var me = this;
                 //    viewState;
+                 /*
+                 * to top
+                 */
+                me.shuffleZIndices(flyout);
```
ZakarFin: flyout.bringToTop() instead? Though shuffleZIndices() resets the z-indexes and bringToTop() just increases the value so after some time flyouts start to have higher z-indexes than the overlay-component?

ZakarFin: The z-value does have a 80k margin so I think we are pretty safe even if we don't reset.

bundles/mapping/toolbar/request/ShowMapMeasurementRequestHandler.ol.js
```diff
@@ -61,7 +61,6 @@ Oskari.clazz.define(
                 // if there is no content container, show the data in dialog
                 if (!me._dialogShown) {
                     dialog.show(me._title, '', me._buttons);
-                    dialog.makeDraggable();
```
ZakarFin: Remove draggable?

okauppinen: Similar popups are not draggable (e.g. map marker, select features). Maybe I was thinking that I don't make another pr.

PR #837, opened 2019-03-27
Fix session timeout

atlain:
Fix 'extend' button functionality


bundles/framework/statehandler/session-methods.js
```diff
@@ -60,13 +60,13 @@ Oskari.clazz.category(
                     const start = Date.now();
                     let diff;
                     let seconds;
-                    const interval = setInterval(timer, 1000);
+                    expireTimeout = setInterval(timer, 1000);
                     function timer () {
                         diff = expireIn - (((Date.now() - start) / 1000) | 0);
                         seconds = (diff % 60) | 0;
                         seconds = seconds < 10 ? 0 + seconds : seconds;
                         if (seconds === 0) {
```
ZakarFin: Change this to if (seconds < 1) since at least in Safari having this run on an inactive tab can make the timer go to a negative number without executing this statement.

PR #838, opened 2019-03-27
Refactor vector based wfs layer plugin

jheijari:
Combines `wfsmvt` & `wfsvector` layer plugins.
To select between mvt and vector implementations, set `renderMode` to `mvt` or `vector` in bundle config.
Wfs layer can also have `options.renderMode` which overrides the plugin setting.


ZakarFin:
This removes Oskari.wfsmvt.WfsMvtLayerPlugin and the current one to use is Oskari.wfsvector.WfsVectorLayerPlugin?

ZakarFin:
I've resolved the conflicts (due to ESLint update) on my local computer and can make a PR for those.

jheijari:
> This removes Oskari.wfsmvt.WfsMvtLayerPlugin and the current one to use is Oskari.wfsvector.WfsVectorLayerPlugin?
> 

Not sure if we should. Having different path name points out it's included in package `wfsvector` and not in `mapwfs2`. What do you think?

ZakarFin:
> > Maybe we should rename it to Oskari.mapwfs2.WfsVectorLayerPlugin to reference the src directory better?
> 
> Not sure if we should. Having different path name points out it's included in package `wfsvector` and not in `mapwfs2`. What do you think?

True, maybe we should just move these implementation files from under mapwfs2, but lets do that in another PR.

PR #852, opened 2019-04-10
Added support to export numeric values 

kuosman:
Added support to export numeric values and changed numeric check to Oskari.util.isNumber.

This PR need also https://github.com/oskariorg/oskari-server/pull/351 to work correct.


ZakarFin:
Thanks!

PR #853, opened 2019-04-10
Fix webpack sprite generation

tsallinen:
Fixed an issue with webpack sprite generation where the source icons where filtered out.

The source directory itself is passed to the copySync filter function and as this didn't match the filter, the destination directory remained empty, which in turn ended the execution of the script in tears:
`sprite.write(spritePath, function (err) {
TypeError: Cannot read property 'write' of null`

Also replaced deprecated `clobber` parameter with `overwrite`.


ZakarFin:
Thanks!

PR #855, opened 2019-04-12
Fix updateLayersInGroups method in MapLayerService to support updatin…

blackfolder:
…g layer group assignment

Related to issue: oskariorg/oskari-docs#105


ZakarFin:
Thanks!

PR #866, opened 2019-04-23
WFS vector userdata layers

okauppinen:
Changed MyPlaces to use oskari style:
- CategoryHandler adds oskari style to oskari layer's options field and updates if layer is edited
- CategoryHandler adds style for text/label. User can't change text style from visu form, so CategoryHandler adds same text style to all MyPlaces layers (styling copied from SLD)
- moved style validation from MyPlaces to visalization form (now only area and line stroke width is validated 1-50)

Changed edit userlayer (myplacesimport) to use oskari style. Moved setstyle null color check (no border/no stroke) from myplacesimport to visu form

WFS vector plugin:
- added _setFeatureLabel which sets feature's first existing value from labelProperty array. (e.g. [attention_text, name] -> if feature has attention_text value uses it for label otherwise name or finally empty string)
- use wfs layermodelbuilder to handle myplaces and userlayer styles. Could also use getPluginInstance and isLayerSupported instead of getLayerPlugin('wfs') and checking if layer is userlayer or myplaces


bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -63,7 +63,12 @@ Oskari.clazz.define(
             defaultStyle.setLegend('');
 
             const mapModule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
-            const wfsPlugin = mapModule.getLayerPlugins(layer.getLayerType());
+            let layerType = layer.getLayerType();
+            if (layerType === 'userlayer' || layerType === 'myplaces') {
+                layerType = 'wfs';
```
ZakarFin: This leaves me a bit cold to have references to userlayer and myplaces here. But I don't have a better suggestion for now so let's go with it and make it better in the future.

PR #870, opened 2019-04-24
Import mapmodule before bundles that refer to classes it provides

ZakarFin:
Fixes an error when opening an embedded map published from the sample app (class_system.js:48 Uncaught Error: No class registered with name: Oskari.mapframework.domain.Style)


jheijari:
Good catch!

PR #874, opened 2019-04-26
Admin layers wfs3

okauppinen:
Add version 3.0.0 option to admin wfs layerselector


ZakarFin:
I added a check for 3.0 support being there so people don't get confused: https://github.com/oskariorg/oskari-frontend/commit/85afd39a9e6da10c6950158e172a2f0fd0544ee4

PR #875, opened 2019-04-26
Admin options for wfs render mode 

jheijari:
Admin can select between `mvt` and `vector` render modes.


ZakarFin:
Changed WFS-admin footer to use the same variable as header: https://github.com/oskariorg/oskari-frontend/commit/6b359cc0518be4c5ea3fe5471888732caa410721

PR #877, opened 2019-04-29
Language selector style fix

jheijari:
Previous solution used an UI library that set global styles. Got rid of it.

### DIFF
--- a/bundles/framework/language-selector/components/LanguageChanger.jsx
+++ b/bundles/framework/language-selector/components/LanguageChanger.jsx
@@ -1,10 +1,10 @@
 import React from 'react';
-import {LanguageSelect} from './LanguageSelect';
-import {getSupportedLanguages} from './getSupportedLanguages';
+import { LanguageSelect } from './LanguageSelect';
+import { getSupportedLanguages } from './getSupportedLanguages';
 
-const handleLanguageChange = (language) => {
+const handleLanguageChange = (event) => {
     let redirect;
-    const newLang = 'lang=' + language;
+    const newLang = 'lang=' + event.target.value;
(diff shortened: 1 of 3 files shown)


bundles/framework/language-selector/components/LanguageChanger.jsx
```diff
@@ -1,18 +1,18 @@
...
-import {LanguageSelect} from './LanguageSelect';
-import {getSupportedLanguages} from './getSupportedLanguages';
+import { LanguageSelect } from './LanguageSelect';
+import { getSupportedLanguages } from './getSupportedLanguages';
 
-const handleLanguageChange = (language) => {
+const handleLanguageChange = (event) => {
     let redirect;
-    const newLang = 'lang=' + language;
+    const newLang = 'lang=' + event.target.value;
```
ZakarFin: I think the API was better with the language being sent to onChange.

bundles/framework/language-selector/components/LanguageChanger.jsx
```diff
@@ -1,18 +1,18 @@
...
     if (window.location.href.indexOf('?') === -1) {
         redirect = window.location.href + '?' + newLang;
     } else if (window.location.href.indexOf('lang=') > -1) {
         redirect = window.location.href.replace(/lang=[^&]+/, newLang);
     } else {
         redirect = window.location.href + '&' + newLang;
     }
-    window.location.href = redirect;
+    window.location.href = redirect.replace('#', '');
```
ZakarFin: Should we re-construct the URL from location.path + modified location.search + location.hash instead?

PR #881, opened 2019-05-02
Statistics fix classification rendering

okauppinen:
Indicator data wasn't always received from backend when components were rendered. Added status for indicatorData. Plugin doesn't render components if data is pending.

Removed service from general context and added mutator to props. Also had to add manualView to props because service isn't passed to manualClassification component.

Now classification plugin gathers props and keeps indicatorData status. This is not good solution, but other plugins depends also on same services and events. Should be refactored when other statistics functionalities are implemented with React.

### DIFF
--- a/bundles/statistics/statsgrid2016/plugin/ClassificationPlugin.js
+++ b/bundles/statistics/statsgrid2016/plugin/ClassificationPlugin.js
@@ -2,6 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { GenericContext } from 'oskari-ui/util';
 import { Classification } from '../components/classification/Classification';
+import { ManualClassificationView } from '../components/manualClassification/View';
 import '../resources/scss/classificationplugin.scss';
 /**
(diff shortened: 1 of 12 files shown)


bundles/statistics/statsgrid2016/plugin/ClassificationPlugin.js
```diff
@@ -78,25 +80,58 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationPlugin',
...
-                        onRenderChange = {this.rendered.bind(this)}/>
+                        onRenderChange = {this.rendered.bind(this)}
+                        indicatorData = {indicatorData}
+                        mutator = {mutator}
+                        manualView = {manualView}/>
                 </GenericContext.Provider>
             ), this.node);
         },
+        getIndicatorData: function (indicators) {
+            const active = indicators.active;
```
jheijari: Null check needed here and in `getIndicatorProps`.

PR #885, opened 2019-05-09
Fix loading indicator to mvt feature data

mikkomustikkamaa:
Implemented logic to send WFSStatusChangedEvents is done as follows.

With MVT layers:

- Loading event is send when first tileloadstart -event is received for the batch
- Error event is send when count of tileloadstart and tileloaderror- events are equal indicating that all started loadings have failed  
- Complete event is send when count of tileloadstart match to sum of tileloadend and tileloaderror events indicating that either successfull or unsuccessfull response has been received for all loading requests.

Counters are impelemented to keep track of event counts.
Timer is implemented as a temporaty solution to send Error event when loading of tile batch has been active 60 seconds.

With Vector layers:

- Loading event is send when feature loader is called
- Error event is send when error- response is received 
- Complete event is send when success -response is received


ZakarFin:
Please check the ESLint rules for formatting in your editor so you don't get so many formatting changes https://github.com/oskariorg/oskari-frontend/blob/develop/.eslintrc.js

bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/MvtLayerHandler.ol.js
```diff
@@ -92,14 +112,92 @@ export class MvtLayerHandler extends AbstractLayerHandler {
...
+                }
+                break;
+            case 'tileloadend':
+            case 'tileloaderror':
+
+                if (this._allStartedTileLoadingsFailed(tileCounter)) {
+                    this._resetTimer(layerId);
+                    super.sendWFSStatusChangedEvent(layerId, 'error');
+                    this._resetCounter(tileCounter);
+                } else if (this._allStartedTileLoadingsAreDone(tileCounter)) {
```
ZakarFin: Have you tried what happens if you remove the layer from the map when it's still loading? You probably need to reset the timer and counters on layer removal as well as a cleanup.

ZakarFin:
Aside from couple of conflicts to solve, cleanup of internal state and formatting changes this looks good to me.

bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/MvtLayerHandler.ol.js
mikkomustikkamaa: I didn't test mentioned scenario but i will do that and include needed fixes to this PR once implemented.

ZakarFin: Would probably work as well if you always reset things in _initializeCountersForLayerIfNeeded() instead of just when missing.

mikkomustikkamaa: I made change to reset counters every time layer is added to map. Also added some other fixes to problems noticed during error case testing and during testing of layer removal / re-adding layer.

mikkomustikkamaa: According to testing, this proposed solution works. However more advanced solution could be tracking counters and timers per layer instance instead of per layerid since user can add and remove same layer multiple times while tileloadings maybe pending in background.

PR #888, opened 2019-05-10
Fix unresponsive features

jheijari:
Extent of the feature is calculated once extent is requested for the first time.
If extent was requested before transforming the feature to projected srs coordinates, extent would have the internal tile coordinates.
 
This forces update on `RenderFeature` extents when internal tile coordinates is used.


jheijari:
This solution might not work on all projections.

PR #897, opened 2019-05-16
Feature loading indicator fix

mikkomustikkamaa:
This fixes the issue where feature data button would not refresh its error state when moving the map in area where tiles had already been loaded.

### DIFF
--- a/bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/AbstractLayerHandler.ol.js
+++ b/bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/AbstractLayerHandler.ol.js
@@ -1,5 +1,10 @@
 import { getFieldsAndPropsArrays } from '../util/props';
 const FEATURE_DATA_UPDATE_THROTTLE = 1000;
+const LOADING_STATUS_VALUE = {
+    COMPLETE: 'complete',
+    LOADING: 'loading',
+    ERROR: 'error'
+};
 
 export class AbstractLayerHandler {
     constructor (layerPlugin) {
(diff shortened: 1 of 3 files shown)


bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/AbstractLayerHandler.ol.js
```diff
@@ -79,6 +84,10 @@ export class AbstractLayerHandler {
             layer.setFields(fields);
             this.plugin.setLayerLocales(layer);
         }
+        if (layer.getActiveFeatures() && layer.getActiveFeatures().length) {
```
jheijari: It would be more clear to have the length check > 0

bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/VectorLayerHandler.ol.js
```diff
@@ -91,12 +92,17 @@ export class VectorLayerHandler extends AbstractLayerHandler {
...
      */
     _getFeatureLoader (layer, source) {
+        let lastExtent = null;
         return (extent, resolution, projection) => {
             const olLayers = this.plugin.getOLMapLayers(layer.getId());
 
             if (olLayers !== undefined && olLayers.length > 0 && !olLayers[0].getVisible()) {
                 return;
             }
+            if (lastExtent && extentEquals(extent, lastExtent) && source.getFeatures().length !== 0) {
```
jheijari: Insted of checking length !== 0 could be length > 0

PR #904, opened 2019-05-28
Add unit tests for util.js

mikkomustikkamaa:


package.json
```diff
@@ -59,9 +62,11 @@
     "geostats": "1.7.0",
     "gm": "1.23.1",
     "imports-loader": "0.8.0",
+    "jest": "^24.8.0",
     "loader-utils": "1.2.3",
     "merge": "1.2.1",
     "mini-css-extract-plugin": "0.5.0",
+    "mobile-detect": "^1.4.3",
```
ZakarFin: Take out the ^ as it's not used in the others and change this https://github.com/oskariorg/oskari-frontend/blob/1.52.0/webpack/oskari-core.js#L7 to use the version from npm instead of the saved on under libraries. The one under libraries can be removed after this. Maybe even split this change to another pr?

src/util.test.js
```diff
@@ -0,0 +1,845 @@
...
+
+        const functionCallCount = 5;
+        const wait = 200;
+        const mockFunction = jest.fn();
+
+        const throttledFunction = OskariMock.util.throttle(mockFunction, wait);
+
+        for (let i = 0; i < functionCallCount; i++) {
+            throttledFunction();
+            await sleep(wait);
```
ZakarFin: Shouldn't you first call the throttledFunction a number of times, check that it hasn't been called but once or 0 times, then sleep, then check that is has been called 5 times?

ZakarFin: Or add an assert on both sides of the sleep? This doesn't really test the throttling but that the function has been called 5 times after we call it 5 times?

src/util.test.js
```diff
@@ -0,0 +1,845 @@
...
+
+    test('returns sort number correctly with double parameters ascending', () => {
+        expect.assertions(1);
+        expect(OskariMock.util.naturalSort(10.1, 22.1, sortDescending)).toEqual(1);
+    });
+
+    test('returns sort number correctly with double parameters descending', () => {
+        expect.assertions(1);
+        expect(OskariMock.util.naturalSort(10.1, 22.1, sortAscending)).toEqual(-1);
+    });
```
ZakarFin: We could add something to test the natural part of the sorting as well like the example from wikipedia:

Alphabetical sorting:

1) z11
2) z2

Natural sorting:

1) z2
2) z11

src/util.test.js
mikkomustikkamaa: Mentioned test did not test actual throttling as you stated. I updated this test to indicate more clearly what is the purpose of that test. Updated tests should now test also actual throttling better .

src/util.test.js
mikkomustikkamaa: Proposed tests added to test natural part of sorting.

package.json
mikkomustikkamaa: I will do this change in another pr.

mikkomustikkamaa: I decided to include these changes to this pr in spite of my previous comment.

src/util.test.js
```diff
@@ -0,0 +1,874 @@
...
+        const functionCallCount = 5;
+        const wait = 500;
+        const mockFunction = jest.fn();
+
+        const throttledFunction = OskariMock.util.throttle(mockFunction, wait);
+
+        for (let i = 0; i < functionCallCount; i++) {
+            throttledFunction();
+            if (i === 0) {
+                expect(mockFunction.mock.calls.length).toEqual(1);
```
ZakarFin: This should work when moved out of the for loop and called without the if.

src/util.test.js
```diff
@@ -0,0 +1,874 @@
...
+        expect.assertions(functionCallCount);
+
+        const wait = 200;
+        const mockFunction = jest.fn();
+
+        const throttledFunction = OskariMock.util.throttle(mockFunction, wait);
+
+        for (let i = 0; i < functionCallCount; i++) {
+            throttledFunction();
+            await sleep(wait);
```
ZakarFin: Move sleep and expect to another loop so calls are made immediately but execution should trickle to completion?

mikkomustikkamaa: Mentioned test intentionally tests that function should be called immediately when calling interval is equal to wait time provided to throttle function.
In other words tests that throttling should not occur when calling interval is equal to wait time.

According to test starting from https://github.com/mikkomustikkamaa/oskari-frontend/blob/add-first-unit-tests/src/util.test.js#L39, 
throttle function sets at most one function call to be executed with delay in situation where multiple function calls are issued before timeout of first function has elapsed. 
(Setting function execution with timeout is done only if timeout variable evaluate to false at line https://github.com/mikkomustikkamaa/oskari-frontend/blob/add-first-unit-tests/src/util.js#L65).

src/util.test.js
mikkomustikkamaa: Mentioned tested is refactored to be more clear by removing unnecessary  for loop.

src/util.test.js
ZakarFin: Oh ok so the test is that if the call is NOT spammed the function WILL be called multiple times in contrast to other tests that spam calls and expect to only have it called once. That's what it says on the test description now that I read it with some thought :)

ZakarFin:
Awesome!

PR #905, opened 2019-05-31
FeatureData ordered fields and locales for WFS vector layers

okauppinen:
Use ordered fields from getLocalizedPropertyNames response. If layer doesn't have localized fields names then set layer's fields from features' properties (visible in extent). Get features' properties as defined in layer's fields (keys and order).

Use localized fields in FeatureData flyout's dropdown if given.


ZakarFin:
Requires oskariorg/oskari-server#393

PR #907, opened 2019-05-31
MarkersPlugin fix icon fill color

okauppinen:


ZakarFin:
Fixes an issue where any marker added to the map with AddMarkerRequest was colored black. Will be hotfixed in 1.52.1 today or early next week.

PR #910, opened 2019-06-06
Fix thematic map search spinner positioning

mikkomustikkamaa:
Unfortunately first solution in commit a763a34 did not work when search button was pushed by pressing enter key instead of clicking button with mouse. When button was pressed with enter, needed y- coordinate containing location of button (issuer of event) was not included in event. For this reason solution in commit b58dcae is implemented which works in both mentioned cases. In proposed solution spinner is placed 100 pixels above buttons in Get content of data search. With data source and indicator selections, spinner should be located as near these selections (works as before).

### DIFF
--- a/bundles/statistics/statsgrid2016/view/SearchFlyout.js
+++ b/bundles/statistics/statsgrid2016/view/SearchFlyout.js
@@ -66,21 +66,26 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
     getNewSearchElement: function () {
         var me = this;
         var container = jQuery('<div></div>');
-
         var selectionComponent = Oskari.clazz.create('Oskari.statistics.statsgrid.IndicatorSelection', me.instance, me.sandbox);
         container.append(selectionComponent.getPanelContent());
+
+        var buttonContainer = jQuery('<div id="statistics-search-flyout-button-container"></div>');
+        container.append(buttonContainer);
 
         var btn = Oskari.clazz.create('Oskari.userinterface.component.Button');
         btn.addClass('margintopLarge');
         btn.setPrimary(true);
         btn.setTitle(this.loc('panels.newSearch.addButtonTitle'));
         btn.setEnabled(false);
-        btn.insertTo(container);
+        btn.insertTo(buttonContainer);
 
         btn.setHandler(function (event) {
             event.stopPropagation();
-            me.setSpinnerVerticalPosition(event, selectionComponent.spinner);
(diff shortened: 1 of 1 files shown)


bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -66,28 +66,34 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
         btn.setTitle(this.loc('panels.newSearch.addButtonTitle'));
         btn.setEnabled(false);
-        btn.insertTo(container);
+        btn.insertTo(buttonContainer);
 
         btn.setHandler(function (event) {
             event.stopPropagation();
-            me.setSpinner(selectionComponent.spinner);
+            const progressSpinner = Oskari.clazz.create('Oskari.userinterface.component.ProgressSpinner');
+            progressSpinner.insertTo(jQuery('#statistics-search-flyout-button-container'));
```
ZakarFin: You can just pass the buttonContainer variable. Also remove the id as we try to use as little those as possible.

ZakarFin: Actually you can do the setup outside of the button handler and only start the spinner in the handler.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
```diff
@@ -66,28 +66,34 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.SearchFlyout', function (t
...
+            progressSpinner.opts.top = -100;
+            me.setSpinner(progressSpinner);
+            me.getSpinner().start();
```
ZakarFin: Nothing stops the spinner?

mikkomustikkamaa: Spinner is stopped at  https://github.com/mikkomustikkamaa/oskari-frontend/blob/b58dcaec61779b58dadfa54b49aa0a73248cf3b9/bundles/statistics/statsgrid2016/view/SearchFlyout.js#L522. Mentioned code in _addIndicators function was already implemented so i decided to use it.

bundles/statistics/statsgrid2016/view/SearchFlyout.js
mikkomustikkamaa: In latest commit, spinner setup logic is moved from button event handler function. Also id attribute is removed from button container div.

PR #912, opened 2019-06-11
Set wfs layers always supported

jheijari:


ZakarFin:
Features from WFS endpoints are transformed to map projection so they are always supported.

PR #915, opened 2019-06-20
jstree upgraded 3.3.5 -> 3.3.8

kuosman:
This upgrade fixes Chrome issue:
jstree.js:9743 [Deprecation] document.registerElement is deprecated and will be removed in M73, around March 2019. Please use window.customElements.define instead. See https://www.chromestatus.com/features/4642138092470272 for more details.


ZakarFin:
Any chance to use this from npm? https://www.npmjs.com/package/jstree

kuosman:
Ok, i check

ZakarFin:
Might need to configure webpack/babel a bit: https://github.com/vakata/jstree/issues/1675

kuosman:
babel-loader exlude not work, same jstree is not function error come

kuosman:
and when added provideplugin (jQuery) come csrf token is null error and says Unable to start because of this (getappsetup action throw error)

--> Cannot use npm

ZakarFin:
Ok, I guess we'll take another crack at npm libs when migrating to React.

PR #916, opened 2019-06-20
Refactored layer support checking

jheijari:
Refactored the way unsupported layers are being handled.

Added new functions to `Oskari.getSandbox().getMap()`: 
* `isLayerSupported`
* `addLayerSupportCheck`
* `getUnsupportedLayerReasons`

Added new domain class `UnsupportedLayerReasons` that provides more information about the reason why layer is not supported. The class also provides means to generate action links to the selected layers list.


bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -1211,17 +1213,25 @@ Oskari.clazz.define(
...
             }
             return this._srsList.indexOf(projection) !== -1;
         },
+        /**
+         * @method @deprecated isSupported
+         * @param {String} projection
+         * @return {Boolean} true if no data about support or param found in supported
+         */
+        isSupported: function (projection) {
+            return this.isSupportedSrs(projection);
```
ZakarFin: Add Oskari.log('AbstractLayer').deprecated() call here so it's easier to spot usage.

PR #921, opened 2019-08-01
Reset remaining session time

mikkomustikkamaa:
PR contains following additions:

- Copied existing function to get error text from jQuery.ajax failure objects to util to make it reausable. Unit tests are added to verify functionality of mentioned function.
- Added logic to send backend request to newly created ActionRoute with 5 minute throttle to extend user session when mousemove event is done anywhere in website (bound to document object)


bundles/framework/statehandler/instance.js
```diff
@@ -495,6 +497,30 @@ Oskari.clazz.define('Oskari.mapframework.bundle.statehandler.StateHandlerBundleI
...
+                    type: 'GET',
+                    url: Oskari.urls.getRoute('ResetRemainingSessionTime'),
+                    error: (jqXHR, textStatus, errorThrown) => {
+                        const errorText = Oskari.util.getErrorTextFromAjaxFailureObjects(jqXHR, errorThrown);
+                        this._log.error(errorText);
+                    },
+                    success: (response) => {
+                        this._log.debug(response);
+
+                        if (Oskari.user().isLoggedIn() && sessionLengthInMinutes > 0) {
```
ZakarFin: This if statement could be moved to the beginning of this method and if the user is not logged in -> skip the rest of the method.

mikkomustikkamaa: Good point. Method updated as recommended.

PR #922, opened 2019-08-02
Ant design configuring and custom themes

jheijari:
~~Load Ant Design components using `babel-plugin-import`. Configured `babel-plugin-import` to~~ Uses `less` styles to enable theme overwriting.

Application specific themes can be set using webpack build environment variable `theme` which should point to a `less` file declaring Ant Design theme variables. Available theme variables can be found in [Ant Design Git repo.](https://github.com/ant-design/ant-design/blob/master/components/style/themes/default.less.)

Prevented Ant global styles by replacing core styles with an empty file during webpack build.

~~NOTICE !! Requires application repositories to install `babel-plugin-import`.~~


ZakarFin:
Looks good but let's hold this on pause for a moment while we figure out how to handle these kind of changes required in app-specific dependencies

jheijari:
Removed babel-plugin-import so there's no more app-specific dependencies.

PR #923, opened 2019-08-07
Wfs feature clustering

jheijari:
Client side clustering for geojson wfs layers.
Supports mixed layers (lines, polygons, points...) but only points are clustered.

Cluster configuring is done in the layer options. Setting `{ clusteringDistance: 20 }` on layer options will create clusters with 20 px buffering area.

`MyPlacesLayerPlugin` and `UserLayersLayerPlugin` can also take `clusteringDistance` in their plugin configurations.

### DIFF
--- a/bundles/mapping/mapuserlayers/plugin/UserLayersLayerPlugin.ol.js
+++ b/bundles/mapping/mapuserlayers/plugin/UserLayersLayerPlugin.ol.js
@@ -29,8 +29,9 @@ Oskari.clazz.define(
          */
         _initImpl: function () {
             const layerClass = 'Oskari.mapframework.bundle.myplacesimport.domain.UserLayer';
+            const { clusteringDistance } = this._config;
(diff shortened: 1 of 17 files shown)


bundles/mapping/mapuserlayers/plugin/UserLayersLayerPlugin.ol.js
```diff
@@ -29,8 +29,9 @@ Oskari.clazz.define(
          */
         _initImpl: function () {
             const layerClass = 'Oskari.mapframework.bundle.myplacesimport.domain.UserLayer';
+            const { clusteringDistance } = this._config;
```
ZakarFin: this._config isn't saved like in MyPlacesLayerPlugin so it's always undefined?

ZakarFin:
One minor change to UserLayersLayerPlugin.ol.js but looks good for merging.

PR #931, opened 2019-08-22
Select features from selected analysis layer

mikkomustikkamaa:
PR contains fix to select features with drawing tools from analysis layer if selected instead of all or top most layer.


ZakarFin:
Analysis now being in oskari-frontend-contrib we shouldn't see it referenced in oskari-frontend. So there's a problem with having getAnalysisWFSLayerId() in WFSLayerService.js but it's a problem we need to solve in some other context.

PR #939, opened 2019-09-02
Use smaller snap tolerance when drawing on desktop

ZakarFin:
Also modifies Oskari.util.isMobile(). Now you can pass true as param to ignore size of the map to know if the device is mobile or desktop.


jheijari:
Looks good

PR #941, opened 2019-09-03
Default version from package.json instead of 'devapp' for builds

ZakarFin:
Allows npm run build parameters for --env.appdef to be included in application specific package.json. Previously you couldn't because the version usually changes and it was part of the folder reference for apps. Now you can still give it as a parameter but it defaults to version in package.json for build and "devapp" for webpack-dev-server (npm start).

Which means you can usually just modify package.json and always use `npm run build` instead of `npm run build -- --env.appdef=THE_ONE_THING_THAT_CHANGES:applications`


jheijari:
Good improvement

PR #946, opened 2019-09-10
Make domain field optional

jheijari:
Set map name the first element.
Made domain field optional.

### DIFF
--- a/bundles/framework/publisher2/view/PanelGeneralInfo.js
+++ b/bundles/framework/publisher2/view/PanelGeneralInfo.js
@@ -17,17 +17,17 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelGeneralInfo
         this.loc = localization;
         this.sandbox = sandbox;
         this.fields = {
-            domain: {
-                label: localization.domain.label,
-                placeholder: localization.domain.placeholder,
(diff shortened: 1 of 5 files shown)


bundles/framework/publisher2/view/PanelGeneralInfo.js
```diff
@@ -65,34 +62,32 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelGeneralInfo
...
-                inputField.setValue(value);
-                sanitizedValue = Oskari.util.sanitize(value);
-                if (sanitizedValue !== value) {
+
+            me.fields.domain.field.setContentCheck(true, me.loc.error.domainIllegalCharacters);
+            me.fields.domain.field.setValidator(function (inputField) {
+                const value = inputField.getValue();
+                const name = inputField.getName();
+                const errors = [];
+                if (value.indexOf('http') === 0 || value.indexOf('www') === 0) {
```
ZakarFin: We could just remove anything before :// if the field has protocol included?

ZakarFin: Atleast the www check should be removed.

ZakarFin:
Localization needs changes as well?

PR #953, opened 2019-09-13
Refactor MapMoveRequest

jheijari:
Moved srsName to options object.
Added animation parameter.


ZakarFin:
We don't actually have impl for the animation yet but its coming. This is the API change.

PR #954, opened 2019-09-13
Fix timeseries printing

mikkomustikkamaa:


bundles/framework/printout/view/BasicPrintout.js
```diff
@@ -298,8 +299,15 @@ Oskari.clazz.define('Oskari.mapframework.bundle.printout.view.BasicPrintout',
 
             contentPanel.append(mapTitle);
 
+            me.areLayersWithTimeSeriesSelected =
```
ZakarFin: Change this to function scoped let/const instead so we don't rely on a side-effect later on.

bundles/framework/printout/view/BasicPrintout.js
```diff
@@ -770,6 +778,16 @@ Oskari.clazz.define('Oskari.mapframework.bundle.printout.view.BasicPrintout',
             if (selections.scaleText) {
                 url = url + '&scaleText=' + selections.scaleText;
             }
+
+            if (me.areLayersWithTimeSeriesSelected) {
```
ZakarFin: Do this again instead of using an instance variable:

     me.instance.sandbox.findAllSelectedMapLayers()
                    .filter(l => l.getAttributes().times).length > 0;

ZakarFin: Add a check for timeseriesPlugin not being undefined as well.

bundles/framework/printout/view/BasicPrintout.js
mikkomustikkamaa: Fixed as proposed.

bundles/framework/printout/view/BasicPrintout.js
mikkomustikkamaa: Fixed as proposed.

PR #955, opened 2019-09-16
Disable clustering from multipoints

jheijari:
Disable clustering for MultiPoints if they actually contain more than one point.


ZakarFin:
Fixes an issue where multipoints were shown as just the first point.

PR #958, opened 2019-09-18
Default to GeoJSON instead of MVT for WFS

ZakarFin:
Since MVT-mode has a hard limit for zoom-level (disregarding min/maxscale set by the admin) and users might have WFS-layers that can be viewed from further away in previous versions.


bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin.ol.js
```diff
@@ -20,7 +20,7 @@ export class WfsVectorLayerPlugin extends AbstractMapLayerPlugin {
         this._config = config;
         this.__name = 'WfsVectorLayerPlugin';
         this._clazz = 'Oskari.wfs.WfsVectorLayerPlugin';
-        this.renderMode = config.renderMode;
+        this.renderMode = config.renderMode || RENDER_MODE_VECTOR;
```
jheijari: The default value is set in init based on mapmodules 3D capabilities.
The change should be made there.

ZakarFin: Changed init to only change renderMode IF 3D mapmodule is used.

PR #967, opened 2019-09-23
Fix hierarchical layerlist publisher2 error

kuosman:
This PR fixes following issue:
* Used hierarchical-layerlist bundle instead of layerselector2
* Open published map or open new publication
* Try to add new layer by pressing **Add map layer**-button
--> Nothing happens, because this JS error come:

> ShowFilteredLayerListRequestHandler.js:30 Uncaught TypeError: n.setActiveFilter is not a function
> at Oskari.clazz.define.handleRequest.handleRequest (ShowFilteredLayerListRequestHandler.js:30)
> at o.clazz.define.getLog.processRequest (sandbox.js:280)
> at a (sandbox.js:354)

Fix:
* Changed filter function same name than layerselector2 bundle


ZakarFin:
Looks good. There seems to be some trouble with the Travis <> GitHub connection why this failed the check. I'll try rerunning it again and wait a bit if we need a hotfix for this.

PR #970, opened 2019-09-24
Os 441 rpc kameran ajot

Somac:
https://jira.nls.fi/browse/OS-441

### DIFF
--- a/bundles/mapping/mapmodule/MapModuleClass.ol.js
+++ b/bundles/mapping/mapmodule/MapModuleClass.ol.js
@@ -538,17 +538,16 @@ export class MapModule extends AbstractMapModule {
      * @param {Object} lonlat coordinates to move the map to
      * @param {Number} zoom absolute zoomlevel to set the map to
      * @param {String} animation animation name
+     * @param {Number}  duration animation duration time
      * @return {Boolean} success
      */
-    animateTo (lonlat, zoom, animation) {
-        lonlat = this.normalizeLonLat(lonlat);
+    _animateTo (lonlat, zoom, animation, duration = 3000) {
         if (!this.isValidLonLat(lonlat.lon, lonlat.lat)) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -538,24 +599,39 @@ export class MapModule extends AbstractMapModule {
...
+                duration: 3000,
+                easing: easeOut
+            });
+        } else {
+            this._animateTo(lonlat, zoomN, options.animation, options.duration);
+        }
+
         this.updateDomain();
         if (suppressEnd !== true) {
             this.notifyMoveEnd();
```
ZakarFin: With this implementation the map moved event comes at the beginning of the animation. We would need to trigger it at the end.

ZakarFin: Also I think we need to have centerMap() to skip any animation by default as users might expect the map to be on a given location after this is called.

bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -538,24 +599,39 @@ export class MapModule extends AbstractMapModule {
...
         if (!this.isValidLonLat(lonlat.lon, lonlat.lat)) {
             return false;
         }
-        this.getMap().getView().setCenter([lonlat.lon, lonlat.lat]);
         if (zoom === null || zoom === undefined) {
             zoom = this.getMapZoom();
         }
-        this.getMap().getView().setZoom(zoom);
+
+        const zoomN = zoom.type === 'scale' ? view.getZoomForResolution(zoom.value) : zoom.value;
```
ZakarFin: We need to check if zoom was a number and treat it as zoom.value and use this logic if zoom was an object to be backwards compatible.

jheijari:
Closing as a duplicate. See https://github.com/oskariorg/oskari-frontend/pull/981.

PR #974, opened 2019-09-25
Fix broken icon on admin-hierarchical-layerlist

ZakarFin:
Edit-layer icon was added to Oskari icons in 1.53.0 which broke the edit-layer icon on admin-hierarchical-layerlist. We need to force styling here as Oskari icons use !important. Proper fix would be trying to remove the !importants from Oskari icons.


kuosman:
Looks great and tested it with hierarchical layerlist. Now layer edit icon is same than previous.

PR #981, opened 2019-10-01
Animated map move and MapTourRequest

Somac:
https://jira.nls.fi/browse/OS-442 & https://jira.nls.fi/browse/OS-441

Test in console:

```
var _sb = Oskari.getSandbox()
_sb.postRequestByName('MapTourRequest', [
    [
        { 
            lon: _sb.getMap().getX() + Math.floor(Math.random() * 300000) - 150000, 
            lat: _sb.getMap().getY() + Math.floor(Math.random() * 300000) - 150000,
            duration: 5000
        },
        { 
            lon: _sb.getMap().getX() + Math.floor(Math.random() * 300000) - 150000, 
            lat: _sb.getMap().getY() + Math.floor(Math.random() * 300000) - 150000,
        },
        { 
            lon: _sb.getMap().getX() + Math.floor(Math.random() * 300000) - 150000, 
            lat: _sb.getMap().getY() + Math.floor(Math.random() * 300000) - 150000,
            animation: 'panZoom',
            zoom: 6
        }
    ],
    { 
        zoom: 8, 
        animation: 'fly', 
        duration: 3000, 
        cancelled: () => console.log('cancelled'),
        completed: () => console.log('completed'),
        srsName: 'EPSG:3067',
    }
]);
```

With additional 3d stuff:
```
var _sb = Oskari.getSandbox()
_sb.postRequestByName('MapTourRequest', [
    [
        { 
            lon: _sb.getMap().getX() + Math.floor(Math.random() * 300000) - 150000, 
            lat: _sb.getMap().getY() + Math.floor(Math.random() * 300000) - 150000,
            duration: 5000,
            angles: { 
                heading: -2,
                roll: 0.5,
                pitch: -0.1
            },
            zoom: 1
        },
        { 
            lon: _sb.getMap().getX() + Math.floor(Math.random() * 300000) - 150000, 
            lat: _sb.getMap().getY() + Math.floor(Math.random() * 300000) - 150000,
            angles: { 
                heading: -0.4,
                roll: 0,
                pitch: -0.1
            }
        },
        { 
            lon: _sb.getMap().getX() + Math.floor(Math.random() * 300000) - 150000, 
            lat: _sb.getMap().getY() + Math.floor(Math.random() * 300000) - 150000,
            zoom: 0.1,
            angles: {
            	heading: 10,
                roll: 0,
                pitch: 0.15
            }
        }
    ],
    { 
        zoom: 10, 
        animation: 'fly', 
        duration: 3000, 
        cancelled: () => console.log('jou'),
        completed: () => console.log('valmis'),
        heading: -0.14003,
        roll: 0.00001,
        pitch: -0.35
    }
]);
```


api/mapping/mapmodule/event/MapTourEvent.md
```diff
@@ -0,0 +1,109 @@
...
+### getLocation()
+Returns map location
+
+### getCompleted
+Returns completion status of tour
+
+### getCancelled
+Returns cancellation status of tour
+
+## Examples
```
ZakarFin: You can remove the example from the event doc (for maintainability so it doesn't go out of sync with the one in request doc in some future update)

bundles/framework/rpc/instance.js
```diff
@@ -176,6 +176,7 @@ Oskari.clazz.define(
                     'MapModulePlugin.RemoveMarkersRequest',
                     'MapModulePlugin.MarkerVisibilityRequest',
                     'MapMoveRequest',
+                    'MapTourRequest',
```
ZakarFin: Add the event to allowedEvents as well.

bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -530,32 +531,173 @@ export class MapModule extends AbstractMapModule {
...
+                    // set Delay for next point
+                    delay = location.delay || delayOption;
+                } else {
+                    me.notifyTourEvent(status);
+                }
+            } else {
+                me.notifyTourEvent(status, true);
+            }
+        };
+        next(true);
```
ZakarFin: Could this be written like this. So always send an event excluding the first call (index is -1). Then we wouldn't have to have so many elses and wouldn't have to set a timeout inside another timeout.

```suggestion
        const next = (more) => {
            if (index !== -1) {
                // trigger if this wasn't the first call
                me.notifyTourEvent(status, !more);
            }
            if (!more) {
                // if we are done we can stop here
                return;
            }
            // go to the next step
            ++index;
            if (index < coordinates.length) {
                // Check if location has overrides for animation values
                const location = coordinates[index];
                const zoomValue = location.zoom || zoomDefault;
                const animationValue = location.animation || animation;
                const durationValue = location.duration || duration;
                status = { ...status, step: index + 1 };

                setTimeout(function () {
                    me._animateTo(location, zoomValue, animationValue, durationValue, next);
                }, delay);

                // set Delay for next point
                delay = location.delay || delayOption;
            }
        };
        next(true);
```

bundles/mapping/mapmodule/event/MapTourEvent.js
```diff
@@ -0,0 +1,77 @@
...
+        },
+
+        /**
+         * @method getParams
+         */
+        getParams: function () {
+            return {
+                status: this.getStatus(),
+                location: this.getLocation(),
+                completed: this.getCompleted()
```
ZakarFin: Not sure if we need canceled or if it can be calculated here from status. But now it's not available for RPC (only in geoportal with the getCancelled() method).

ZakarFin: I guess status can be calculated with status.steps !== status.step (if the completed step wasn't the last one -> was canceled)?

bundles/mapping/mapmodule/AbstractMapModule.js
```diff
@@ -1068,6 +1070,15 @@ Oskari.clazz.define(
             var evt = Oskari.eventBuilder('AfterMapMoveEvent')(lonlat.lon, lonlat.lat, this.getMapZoom(), this.getMapScale(), creator);
             sandbox.notifyAll(evt);
         },
+
+        notifyTourEvent: function (status, cancelled = false) {
```
ZakarFin: Should we switch canceled to completed and calculate canceled based on status.steps !== status.step?

api/mapping/mapmodule/event/MapTourEvent.md
Somac: Removed

bundles/framework/rpc/instance.js
Somac: Added

bundles/mapping/mapmodule/event/MapTourEvent.js
Somac: If we add me.notifyTourEvent(status, true) to the !more check, we get cancellation event from tourMap.

bundles/mapping/mapmodule/AbstractMapModule.js
Somac: It now gets completed if the current step === last step, and cancelled if the animation event doesn't get to finish (for example if user moves map).

bundles/mapping/mapmodule/MapModuleClass.ol.js
Somac: Yeah this so much more beautiful than what it was before. I wondered if it would be logical to send status to event on first call also, but have the step be 0 for example?

bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -530,32 +531,173 @@ export class MapModule extends AbstractMapModule {
...
+
+        this._animateTo(lonlat, zoomValue, animation, duration);
+
         this.updateDomain();
         if (suppressEnd !== true) {
-            this.notifyMoveEnd();
+            setTimeout(() => {
+                this.notifyMoveEnd();
+            }, notifyDuration);
         }
```
ZakarFin: Just realized that we need to call updateDomain() _after_ the map has moved. So do something like this:
```suggestion
    setTimeout(() => {
        this.updateDomain();
        if (suppressEnd !== true) {
            this.notifyMoveEnd();
        }  
   }, notifyDuration);
```
Otherwise the domain object will have the location of the map _before_ the animation starts and not what the location is _after_ the animation.

bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -530,32 +531,173 @@ export class MapModule extends AbstractMapModule {
...
+
+        const zoomDefault = zoom.type === 'scale' ? view.getZoomForResolution(zoom.value) : zoom.value;
+        let index = -1;
+        let delay = 0;
+        let status = { steps: coordinates.length };
+
+        const next = (more) => {
+            if (index !== -1) {
+                // trigger if this wasn't the first call
+                me.notifyTourEvent(status, !more);
```
ZakarFin: Add calls to:

this.notifyMoveEnd();
this.updateDomain();

So components that update their UI based on map location refresh accordingly

ZakarFin: Scratch that. This https://github.com/oskariorg/oskari-frontend/blob/1.53.1/bundles/mapping/mapmodule/MapModuleClass.ol.js#L101 gets triggered with the animate method so the events and domain get updated.

bundles/mapping/mapmodule/MapModuleClass.ol.js
ZakarFin: Scratch that. This https://github.com/oskariorg/oskari-frontend/blob/1.53.1/bundles/mapping/mapmodule/MapModuleClass.ol.js#L101 gets triggered with the animate method so the events and domain get updated.

ZakarFin: This can be removed as https://github.com/oskariorg/oskari-frontend/blob/1.53.1/bundles/mapping/mapmodule/MapModuleClass.ol.js#L101 gets triggered with the animate method so the events and domain get updated automatically.

bundles/mapping/mapmodule/MapModuleClass.ol.js
ZakarFin: Like a "tour starting" event. Sure why not.

bundles/mapping/mapmodule/request/MapMoveRequestHandler.js
```diff
@@ -29,37 +29,27 @@ Oskari.clazz.define(
...
-            var zoomChange = (zoom || zoom === 0);
-
-            // if zoom is about to change -> Suppress the event
-            this.mapModule.centerMap(lonlat, null, !!zoomChange);
-            if (zoomChange) {
-                const { left, top, bottom, right } = zoom;
-
-                if (left && top && bottom && right) {
-                    const zoomOut = top === bottom && left === right;
-                    this.mapModule.zoomToExtent(zoom, zoomOut, zoomOut);
```
ZakarFin: Forgot this too. This isn't handled now by the centerMap() call. So we need to accept zoom that is an object with left, top, bottom, right keys.

bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -624,16 +625,18 @@ export class MapModule extends AbstractMapModule {
             return false;
         }
 
+        if (zoom && top && bottom && left && right) {
+            const zoomOut = top === bottom && left === right;
+            this.zoomToExtent(zoom, zoomOut, zoomOut);
+            view.setCenter(lonlat);
+            return true;
```
ZakarFin: Did you check does this trigger the AfterMapMoveEvent or do we need to add the notifyMoveEnd() call here to be used with view.setCenter()

ZakarFin: The event is sent.

PR #983, opened 2019-10-03
remove spinner from autocomplete

jheijari:
Removed faulty spinner from search autocomplete.


ZakarFin:
Autocompletion should be performant enough not to show spinner and the user hasn't actually done anything but written on the field for it to trigger -> no need for a spinner to indicate "something is running"

PR #988, opened 2019-10-04
Add domain validation to map publishing

mikkomustikkamaa:
Added domain validation to map publishing using regular expression available at https://miguelmota.com/bytes/validate-domain-regex/. Yellow warning icon with black text is shown to user with 1 second delay when needed.

### DIFF
--- a/bundles/framework/publisher2/view/PanelGeneralInfo.js
+++ b/bundles/framework/publisher2/view/PanelGeneralInfo.js
@@ -38,7 +38,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelGeneralInfo
                 '</div>')
         };
         this.panel = null;
-        this.domainWarningTemplate = jQuery('<div class="domain-warning">TODO: tekstit lokaalisaatiotiedostosta</div>');
+        this.domainWarningTemplate = jQuery('<div class="domain-warning">' + this.loc.domain.inputWarning + '</div>');
     }, {
         /**
          * Creates the set of Oskari.userinterface.component.FormInput to be shown on the panel and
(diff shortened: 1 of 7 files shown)


bundles/framework/publisher2/view/PanelGeneralInfo.js
```diff
@@ -58,6 +59,24 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelGeneralInfo
                     field.setLabel(data.label);
                     field.setTooltip(data.tooltip, data.helptags);
                     field.setPlaceholder(data.placeholder);
+
+                    const keyUpHandler = Oskari.util.throttle((event) => {
```
ZakarFin: Define the function outside the loop and rename it to something like domainValidatorKeyUpHandler as it's only used for domain.

bundles/framework/publisher2/view/PanelGeneralInfo.js
```diff
@@ -125,6 +144,16 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelGeneralInfo
 
             me._languageChanged(selectedLang);
         },
+        /*
+         * Function implemented by modifying function introduced in https://miguelmota.com/bytes/validate-domain-regex/
+         */
+        _isValidDomain (v) {
```
ZakarFin: This should be moved to Oskari.util.

bundles/framework/publisher2/view/PanelGeneralInfo.js
mikkomustikkamaa: Fixed as proposed.

bundles/framework/publisher2/view/PanelGeneralInfo.js
mikkomustikkamaa: Function moved to Oskari.util and some unit tests is added.

ZakarFin:
Adds Oskari.util.isValidDomain("domain as string")

PR #990, opened 2019-10-08
MyLocationPlugin publisher preview

kuosman:
This PR fixes MyLocationPlugin publisher preview (https://github.com/oskariorg/oskari-docs/issues/132).


ZakarFin:
Thanks!

PR #993, opened 2019-10-09
Initial integration of layer wizard and layer editor

ZakarFin:
New layers will show the wizard. Editing existing layer shows the form. 

Includes mock-data for capabilities and the form doesn't fully work when adding layers (js errors when wizard shows the form).


bundles/admin/admin-layereditor/view/AdminLayerFormService.js
```diff
@@ -119,28 +174,29 @@ export class AdminLayerFormService {
...
-            opacity: layer ? layer.getOpacity() : 100,
-            minScale: layer ? layer.getMinScale() : 1,
-            maxScale: layer ? layer.getMaxScale() : 1,
-            style: layer ? layer.getCurrentStyle().getName() : '',
-            styleTitle: layer ? layer.getCurrentStyle().getTitle() : '',
+            groupId: layer.getAdmin().organizationId,
+            organizationName: layer.getOrganizationName(),
+            maplayerGroups: [...layer.getGroups()],
+            opacity: layer.getOpacity() || 100,
+            minScale: layer.getMinScale() || 1,
```
jheijari: The scale defaults still need some work, but it's not the focus here, so this is ok.

bundles/admin/admin-layereditor/view/AdminLayerFormService.js
```diff
@@ -333,6 +394,6 @@ export class AdminLayerFormService {
     }
 
     notify () {
-        this.listeners.forEach(consumer => consumer());
+        this.listeners.forEach(consumer => consumer(this.getLayer()));
```
jheijari: The layer doesn't seem to be used. It's accessed from the getter instead.

bundles/admin/admin-layereditor/view/LayerWizard.jsx
```diff
@@ -1,100 +1,106 @@
...
-        case 1:
-            this.service.getMutator().setVersion();
-            break;
-        case 2:
-            this.service.getMutator().setName();
-            break;
-        }
+function setStep (mutator, requested) {
+    switch (requested) {
+    case 0:
```
jheijari: Use constants instead of magic numbers

bundles/admin/admin-layereditor/view/LayerWizard.jsx
```diff
@@ -1,100 +1,106 @@
...
+                    onSelect={(type) => mutator.setType(type)} />
+            }
+            { getStep(layer) === 1 &&
+                <div>
+                    <LayerURLForm
+                        layer={layer}
+                        loading={loading}
+                        service={mutator} />
+                    <hr/>
+                    <Button onClick={() => setStep(mutator, 0)}>{getMessage('cancel')}</Button>
```
jheijari: Create local reusable cancel button?

bundles/admin/admin-layereditor/view/AdminLayerFormService.js
ZakarFin: True, though I'd much rather get the modified result as param as a consumer/listener but since we don't use it now I'll remove it.

bundles/admin/admin-layereditor/view/LayerWizard.jsx
ZakarFin: Removed step specific back button and using a history(-1) type logic for it.

jheijari:
Looks good. Merging.

PR #994, opened 2019-10-09
Fix layer support check

mikkomustikkamaa:
Use isLayerSupported logic from map to check need to show warning icon to user on layerlist


ZakarFin:
Good job adding tests!

PR #996, opened 2019-10-11
Support optional params in Oskari.urls.getRoute()

ZakarFin:
Oskari.urls.getRoute() now accepts an optional second parameter for url-params that should be included in the url. Also added sanity checks for url instead of assuming it ends with ? or &

### DIFF
--- a/src/oskari.es6.js
+++ b/src/oskari.es6.js
@@ -18,6 +18,32 @@ function getUrl (key) {
     return _urls[key];
 }
 
+function encodeParams(params) {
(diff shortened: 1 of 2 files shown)


src/oskari.es6.js
```diff
@@ -80,11 +106,11 @@ const Oskari = {
              * @param  {String} route [description]
              * @return {String} url to use when making API calls
```
jheijari: Add `@param` documentation for `optionalParams`. String, array, object?

PR #998, opened 2019-10-11
Server integration for layer admin

ZakarFin:
Work in progress but mergeable for integrating oskariorg/oskari-server#461 and the new layer admin.

What works is the workflow for adding new layers and showing the edit form for an existing one. Pretty much everything else still needs work. Also the server response needs some modifications still.


bundles/admin/admin-layereditor/view/LayerHelper.js
```diff
@@ -0,0 +1,157 @@
+export const LayerHelper = (Oskari) => {
```
jheijari: Export LayerHelper as a class and leave private functions out of it.

bundles/admin/admin-layereditor/view/LayerWizard/LayerCapabilitiesListing.jsx
```diff
@@ -1,20 +1,127 @@
...
+`;
+export const StylePopUl = styled.ul`
+max-width: 400px;
+`;
+export const StylePopLi = styled.li`
+overflow: hidden;
+text-overflow: ellipsis;
+`;
+
+export const LayerCapabilitiesListing = (props) => {
```
jheijari: Destruct props for clarity.

bundles/admin/admin-layereditor/view/LayerWizard/LayerCapabilitiesListing.jsx
```diff
@@ -1,20 +1,127 @@
...
+    return (
+        <React.Fragment>
+            <LayerCapabilitiesFilter
+                placeholder="Filter layers"
+                filter={filter}
+                onChange={(value) => setfilter(value)}/>
+            <List dataSource={layers} rowKey="name" renderItem={item => getItem(props.onSelect, item)}></List>
+        </React.Fragment>);
+};
+
```
jheijari: Move `propTypes` next to the element declaration.

bundles/admin/admin-layereditor/view/Flyout.js
```diff
@@ -41,46 +41,53 @@ export class LayerEditorFlyout extends ExtraFlyout {
...
-
     update () {
-        const me = this;
         const el = this.getElement();
         if (!el) {
             return;
         }
+        let uiCode = this.getEditorUI();
+        if (this.service.isLoading()) {
+            uiCode = <Spin>{ uiCode }</Spin>;
```
jheijari: Move this logic to `LayerWizard`.

jheijari:
Suggesting a few minor changes. The functionality is still clearly work in progress and all suggestions might not be needed at this point.

bundles/admin/admin-layereditor/view/LayerHelper.js
ZakarFin: I'll rename it getLayerHelper and pass an array of supported languages instead of Oskari as param. It's mostly code moved from LayerFormService to a helper file to make the service file more readable. The service itself doesn't even use the new mutator mixin so this needs more work to be ready.

bundles/admin/admin-layereditor/view/Flyout.js
ZakarFin: Good idea. There's a lot of stuff that should probably be moved around still but since we are prioritizing on getting the new layerlisting ready instead of the admin functionality it's good to get the "pending" work towards the admin merged even if it's not ready since the version in develop isn't ready either.

ZakarFin: Added TODO

PR #999, opened 2019-10-11
Creating layer list tabs in react and refactoring service logic

jheijari:
Massive refactor for `layerlist` bundle. Moved away from `layerselector2`'s instance and flyout.
Everything except `Flyout` and `Tile` in this bundle is now in react.

* Improved performance
* Focusing on text search
* Fresh look for tabs

Todo:
* Bundle's Oskari state is not yet supported.
* Suggestive text search


bundles/admin/admin-layereditor/components/Button.jsx
```diff
@@ -1,10 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 import { Button as AntButton } from 'antd';
 import 'antd/es/button/style/index.js';
 
-export const Button = ({children, ...other}) => (
-    <AntButton {...other}>{children}</AntButton>
+const Styled = styled(AntButton)`
+    > .anticon {
```
ZakarFin: This is because we configured the ant-styles out of the global scope so we need to do some extra tweaking to make the icon align perfectly as with the global styles. Add something along these lines as a comment as the reasoning for this.

ZakarFin: Or should we gather some readme.md on the components folder to explain this and have some info about the styling in general as well (some mention about webpack config for these)?

bundles/framework/layerlist/Flyout.js
```diff
@@ -0,0 +1,120 @@
...
+        getOptions: function () { },
+
+        /**
+         * Set content state
+         * @method  @public setContentState
+         * @param {Object} state a content state
+         */
+        setContentState: function (state) {
+            // TODO; Set filter and selected tab
+            console.warn('Unimplemented layer list flyout function: setContentState');
```
ZakarFin: Use Oskari.log('layerlist').warn() instead.

bundles/framework/layerlist/Flyout.js
```diff
@@ -0,0 +1,120 @@
...
+
+        /**
+         * @method startPlugin
+         *
+         * Interface method implementation, assigns the HTML templates that will be used to create the UI
+         */
+        startPlugin: function () {
+            var elParent = this.container.parentElement.parentElement;
+            var elId = jQuery(elParent).find('.oskari-flyouttoolbar .oskari-flyouttools .oskari-flyouttool-close');
+            elId.attr('id', 'oskari_layerselector2_flyout_oskari_flyouttool_close');
```
ZakarFin: I think the ID is for Selenium testing so lets remove it for now. We can use the new t_something class for this if it's still required by tests but as this is a new bundle it's perfect time to correct previous sins.

bundles/framework/layerlist/Tile.js
```diff
@@ -0,0 +1,110 @@
...
+         * Creates the UI for a fresh start
+         */
+        refresh: function () {
+        }
+    }, {
+        /**
+         * @property {String[]} protocol
+         * @static
+         */
+        'protocol': ['Oskari.userinterface.Tile']
```
ZakarFin: If you extend 'Oskari.userinterface.extension.DefaultTile' this would be far fever lines. For now the custom tile isn't needed but we need a custom implementation for the "selected layers count" label that we will add in a future update.

bundles/framework/layerlist/service/layerlist.js
```diff
@@ -0,0 +1,60 @@
...
+                return this.__qname;
+            },
+            /**
+             * @method getName
+             * @return {String} service name
+             */
+            getName: function () {
+                return this.__name;
+            },
+            getMutator () {
```
ZakarFin: getMutator() is something that has been left from a previous version as this.mutator isn't initialized?

bundles/framework/layerlist/view/LayerList/LayerCollapse/CollapseService.js
```diff
@@ -0,0 +1,206 @@
...
+        return this.map.getLayers().map(layer => layer.getId());
+    }
+
+    addLayer (id) {
+        if (!id || this.state.selectedLayerIds.includes(id)) {
+            return;
+        }
+        const selectedLayerIds = [...this.state.selectedLayerIds, id];
+        this.updateState({ selectedLayerIds });
+        setTimeout(() => this.sandbox.postRequestByName('AddMapLayerRequest', [id]), ANIMATION_TIMEOUT);
```
ZakarFin: Timeout is used for performance reasons. If UI is updated for selected layer during animation there is a notable performance hit and the animation becomes stuttery. This should be commented on the code as my first reaction would be to remove the timeout :)

bundles/framework/layerlist/view/LayerList/LayerCollapse/util.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {String} method layer method name to sort by
+ */
+const comparator = (a, b, method) => {
+    var nameA = a[method]().toLowerCase();
+    var nameB = b[method]().toLowerCase();
+    if (nameA === nameB && (a.getName() && b.getName())) {
+        nameA = a.getName().toLowerCase();
+        nameB = b.getName().toLowerCase();
+    }
+    if (nameA < nameB) {
```
ZakarFin: We could use Oskari.util.naturalSort(nameA, nameB) here after we have determined what we need to compare instead of a simple comparison.

src/react/mutatorMixin.js
```diff
@@ -0,0 +1,24 @@
+export const mutatorMixin = (Service, mutatingMethods) => class extends Service {
+    constructor (...args) {
+        super(...args);
+
+        this.mutatingMethods = mutatingMethods || [];
+        if (!this.mutatingMethods) {
```
ZakarFin: This is never true. Add .length to check if the array is empty.

bundles/admin/admin-layereditor/components/Button.jsx
jheijari: Fixed ant style handling by modifyin ant-globals.less. Also relocated ant-globals.less and ant-theme.less under oskari-ui.
Added README.md under oskari-ui.

src/react/mutatorMixin.js
jheijari: Updated array check.

bundles/framework/layerlist/service/layerlist.js
jheijari: Refactored singleton layerlist service in layerselector2 and removed this file.

bundles/framework/layerlist/view/LayerList/LayerCollapse/util.js
jheijari: Using natural sort

bundles/framework/layerlist/view/LayerList/LayerCollapse/CollapseService.js
jheijari: Added comment

bundles/framework/layerlist/Tile.js
jheijari: Using the DefaultTile directly for now

PR #1002, opened 2019-10-16
WIP: Layer permissions

mikkomustikkamaa:
Work in progress. Initial permission layout based on fetched possible user roles and permission types.


bundles/admin/admin-layereditor/view/AdminLayerFormService.js
```diff
@@ -372,6 +373,34 @@ export class AdminLayerFormService {
...
+            me.getMutator().setMessage('messages.errorFetchUserRolesAndPermissionTypes', 'error');
+            me.notify();
+        });
+    }
+
+    getRolesAndPermissionTypes () {
+        return this.rolesAndPermissionTypes;
+    };
+
+    setListener (consumer) {
```
ZakarFin: This isn't called anywhere.

bundles/admin/admin-layereditor/view/AdminLayerFormService.js
```diff
@@ -372,6 +373,34 @@ export class AdminLayerFormService {
...
     }
 
+    fetchRolesAndPermissionTypes () {
+        const me = this;
+        this.loading = true;
+        fetch(Oskari.urls.getRoute('GetAllRolesAndPermissionTypes')).then(response => {
+            if (response.ok) {
+                return response.json();
+            } else {
+                return Promise.reject(Error('Fetching user roles and permission types failed'));
```
ZakarFin: Should this be "new Error" instead of just "Error"?

mikkomustikkamaa: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error states "When Error is used like a function -- without new, it will return an Error object. Therefore, a mere call to Error will produce the same output that constructing an Error object via the new keyword would." I admit that it is clearer for most readers if new key word is included.

bundles/admin/admin-layereditor/view/AdminLayerFormService.js
mikkomustikkamaa: :) setListener() is now removed

bundles/admin/admin-layereditor/view/AdminLayerFormService.js
mikkomustikkamaa: new keyword added

bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionsTabPane.jsx
```diff
@@ -1,7 +1,70 @@
...
+        <StyledListItem>
+            <PermissionRow {...rowModel}/>
+        </StyledListItem>
     );
 };
+
+const PermissionsTabPane = (props) => {
+    const { getMessage, rolesAndPermissionTypes } = props;
+    var permissionDataModel;
+    if (rolesAndPermissionTypes) {
```
ZakarFin: This is set as required in proptypes so I guess the if isn't needed. Anyway if it IS needed I would refactor to:

  if (!rolesAndPermissionTypes) {
    return;
  }

bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionTabPane/PermissionRow.jsx
```diff
@@ -0,0 +1,81 @@
...
+                    { props.rowText}
+                </HeaderTextColumn>
+                { permissionTypes }
+            </RowContainer>
+            : <RowContainer>
+                <TextColumn>
+                    { props.rowText}
+                </TextColumn>
+                { permissionTypes }
+            </RowContainer>
```
ZakarFin: How about doing two returns here to make this more readable?

  if (props.isHeaderRow) {
     return (...);
  }
  return (...);

ZakarFin: Or consider doing two exports. One for header row and another for normal one. Then you could take out the isHeaderRow prop.

mikkomustikkamaa: I ended up doing conditional styling in TextColumn styled.div so HeaderTextColumn is removed as obsolete and if statement is no longer needed in render

bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionsTabPane.jsx
mikkomustikkamaa: Changed roles and permissions as not mandatory and added if statement as recommended. Earlier these roles and permissions might be undefined if loading from backend is in progress. With recently implemented spinner logic, permission tab should not be rendered with undefined roles and permissions. Still i thought that marking this prop as not required is way to go.

bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionTabPane/PermissionRow.jsx
mikkomustikkamaa: With refactoring done in latest commits, code is still work in progress in scope of whole layer permissions handling but it should be in better shape to add actual state and saving logic to permissions later on in other pull request(s).

bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionsTabPane.jsx
```diff
@@ -1,7 +1,76 @@
...
+import { PermissionRow } from './PermissionTabPane/PermissionRow';
+import { List, ListItem, Checkbox } from 'oskari-ui';
+import { withLocale } from 'oskari-ui/util';
 
-export const PermissionsTabPane = (props) => {
+const StyledListItem = styled(ListItem)`
+    &:first-child > div {
+        font-weight: bold;
+    }
+    &:not(:first-child) {
```
ZakarFin: TRy using even/odd from css instead of additional prop:
```
tr:nth-child(even) {background: #ffffff}
tr:nth-child(odd) {background: #f3f3f3}
```

bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionTabPane/PermissionRow.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+export const PermissionRow = (props) => {
+    const checkboxDivs = [];
+    props.checkboxes.forEach(checkbox => {
+        if (props.isHeaderRow) {
+            checkboxDivs.push(<StyledPermissionDiv key={checkbox.key}><HeaderPermissionText >{checkbox.props.selectionText}</HeaderPermissionText><Break/><SelectAllDiv>{checkbox}</SelectAllDiv></StyledPermissionDiv>);
+        } else {
+            checkboxDivs.push(<StyledPermissionDiv key={checkbox.key}>{checkbox}</StyledPermissionDiv>);
+        }
+    });
```
ZakarFin: Would this work/be more readable:
```
    const checkboxDivs = props.checkboxes.map(checkbox => {
        const content = checkbox;
        if (props.isHeaderRow) {
            content = (<HeaderPermissionText >{checkbox.props.selectionText}</HeaderPermissionText><Break/><SelectAllDiv>{content}</SelectAllDiv>);
        } 
        return (<StyledPermissionDiv key={checkbox.key}>{content}</StyledPermissionDiv>);
    )};
```
Or maybe even:
```
    const checkboxDivs = props.checkboxes.map(checkbox => {
        const content = checkbox;
        if (props.isHeaderRow) {
            content = (
                <React.Fragment>
                    <HeaderPermissionText>{checkbox.props.selectionText}</HeaderPermissionText>
                    <Break/>
                    <SelectAllDiv>{content}</SelectAllDiv>
                </React.Fragment>
            );
        } 
        return (
            <StyledPermissionDiv key={checkbox.key}>
                {content}
            </StyledPermissionDiv>
        );
    )};

Comment (mikkomustikkamaa) [reply bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionTabPane/PermissionRow.jsx]: Fixed as suggested

File: bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionsTabPane.jsx (hunk 5 above)
Comment (mikkomustikkamaa) [reply bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionsTabPane.jsx]: Fixed as suggested

File: bundles/admin/admin-layereditor/view/AdminLayerForm/PermissionsTabPane.jsx [hunk 7]
```diff
@@ -1,7 +1,76 @@
...
+
+    const rolePermissionRows = roles.map((role, index) => {
+        const rolePermissionCheckboxes = permissionTypes.map((permission) => {
+            return <Checkbox key={permission.id + '_' + role.id}
+                selectionText={getMessage('rights.' + permission.id)}
+                permission={permission.id} role={role.id}
+                onChange = {checkboxOnChangeHandler}/>;
+        });
+        return <PermissionRow key={role.id} even={index % 2 === 0} isHeaderRow={false} text={role.name} checkboxes={rolePermissionCheckboxes}/>;
+    });
```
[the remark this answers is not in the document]
mikkomustikkamaa: Fixed as suggested

PR #1008, opened 2019-10-21
Refactored OtherLanguagesPane and LocalizedLayerInfo to own reusable components

Somac:


ZakarFin:
I think the API for these should be something like:
```   
<LocalizedContent languages={array like ['fi', 'sv', 'en']}>
    {children}
</LocalizedContent>
```

In layer admin this would be used with something like this:
```
<LocalizedContent name="locale" languages={Oskari.getSupportedLanguages()} onChange={console.log}>
    <input name="title" />
    <input name="subtitle" />
</LocalizedContent>
```
This would generate an UI that has 2 input fields for the main language (first language in languages prop array) and repeating those two fields in a collapse panel for the other languages. The onchange would log something like this to the console:
```
 {
     locale: {
         fi: {
            title: 'What the user wrote',
            subtitle: 'What the user wrote'
         }
     }
 }
 ```

 Where fi is a value in the languages prop and any other languages that are filled out would be saved in a similar structure next to each other.
  Locale in the output is the name given to LocalizedContent and title/subtitle are the names given to children.

File: bundles/admin/admin-layereditor/view/AdminLayerForm/LocalizedLayerInfo.jsx [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+    const name = layer[`name_${lang}`];
+    const description = layer[`title_${lang}`];
+    const labels = getLocalizedLabels(lang, getMessage);
+    const onNameChange = evt => service.setLocalizedLayerName(lang, evt.target.value);
+    const onDescriptionChange = evt => service.setLocalizedLayerDescription(lang, evt.target.value);
+    const nameInput = <TextInput type='text' value={name} onChange={onNameChange} />;
+    const descInput = <TextInput type='text' value={description} onChange={onDescriptionChange} />;
+    if (selectedLang === lang) {
+        return (
+            <>
```
ZakarFin: Use React.Fragment here as well for consistency.

ZakarFin:
Resolve the conflict and fix the React.Fragment and lets merge this. We can fix the localization component later to have a better API.

PR #1009, opened 2019-10-21
Dropdown and Menu antDesign components made for oskari-ui

Somac:


ZakarFin:
Styling will probably need the manual import for like here https://github.com/oskariorg/oskari-frontend/commit/b7f2b198057e37a66e878d2408a7e89e113f5275

    import 'antd/es/popover/style/index.js';

PR #1017, opened 2019-10-24
Move 3d bundles

okauppinen:
Move 3d related code from pti-frontend repo: dimension-change, tiles3d, mapmodule.olcs, LayersPlugin.olcs. These files have commit history, but git doesn't show moved history without --forward flag.

Packages are copied without commit history because them doesn't contain major changes.
I don't know why everything mapmodule packages is under ol3, but now added dimension-change and tiles3d straight under mapmodule (not olcs). I think that only duplicate mapmodule files should be under ol3 and olcs. These files have different mapmodule and LayerPlugin implementations/references (ol/olcs). Other references are same.

Added 3d fixes to ToolSelectionHandler. Updated import and package references.


bundles/mapping/dimension-change/resources/locale/fi.js
```diff
@@ -0,0 +1,12 @@
+Oskari.registerLocalization(
+    {
+        "lang": "fi",
+        "key": "dimensionchange",
+        "value": {
+            "to3Dview": "Siirry 3D-näkymään. Käytettävissä vain osa karttatasoista ja työkaluista.",
+            "backTo2Dview": "Palaa Paikkatietoikkunan oletusnäkymään.",
```
ZakarFin: Paikkatietoikkuna reference here that should be removed if we merge this to Oskari. Maybe we should consider moving the dimension change to Paikkatietoikkuna repository instead of having it on oskari-frontend or make a more generic "appsetup selector" bundle for Oskari.

bundles/mapping/tiles3d/plugin/Tiles3DModelBuilder.js
```diff
@@ -0,0 +1,18 @@
+const Style = Oskari.clazz.get('Oskari.mapframework.domain.Style');
+
+export class Tiles3DModelBuilder {
+    parseLayerData (layer, mapLayerJson, maplayerService) {
+        if (!mapLayerJson.options) {
+            return;
+        }
+        const options = mapLayerJson.options;
+        const styles = jQuery.extend({}, options.styles || {}, options.externalStyles || {});
```
ZakarFin: Could we use object spread instead of jQuery here?

bundles/mapping/dimension-change/resources/locale/fi.js
ZakarFin: Let's go with having the dimension-change bundle in oskari-frontend for now and just remove the PTI reference from the toolip.

PR #1018, opened 2019-10-25
Openlayers 6

mikkomustikkamaa:
Changes related upgrade to openlayers 6


bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/MvtLayerHandler/FeatureExposingMVTSource.js
```diff
@@ -117,7 +116,7 @@ export class FeatureExposingMVTSource extends olSourceVectorTile {
      * @param {ol/VectorTile} tile
      */
     _getRenderFeatureExtent (renderFeature, tile) {
-        if (intersects(renderFeature.getExtent(), tile.getExtent())) {
+        if (intersects(renderFeature.getExtent(), tile.extent)) {
             renderFeature.extent_ = null;
```
ZakarFin: Should this be extent as well instead of extent_?

ZakarFin:
Add some notes about the path changes to oskariorg/oskari-server/MigrationGuide.md

bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/MvtLayerHandler/FeatureExposingMVTSource.js
mikkomustikkamaa: Getter for tile extent is dropped in OL6 so direct field reference is used. Feature does not contain setter for extent (https://openlayers.org/en/latest/apidoc/module-ol_render_Feature-RenderFeature.html) so property is set directly to null as with OL5. Checked with debugger that features extent field ends to underscore. 

Minimum effort approach is used with this FeatureExposingMVTSource implementation to get it work with OL6 in same way it worked with OL5. There has been some discussion to contact open layers developers to maybe request them to add needed features to API (or maybe we could do PR) so we could achieve this MVT property fetching trough the API:s instead of current inheritance and direct field access method. Mentioned actions and improvements are out of scope of this pull request.

mikkomustikkamaa:
Added notes about path changes to oskari-server/MigrationGuide.md in https://github.com/oskariorg/oskari-server/pull/466

PR #1027, opened 2019-11-01
Fixes to thematic map classification

mikkomustikkamaa:
Changes in this PR:

- Allow classification when value count is two using other methods than natural intervals (jenks). According to testing and values used in it geostats library produces some jenks class boundaries with undefined values when input data contains only two values. In these cases classification now defaults to  quantile method and natural intervals classification method selection is disabled from classification editing.
-  geostats library is upgraded to latest release (https://github.com/simogeo/geostats/releases/tag/v1.8.0) as a attempt to solve classification problems mentioned above. It seems that upgrade did not solve these problems but since it includes performance upgrade related to jenks classification (https://github.com/simogeo/geostats/pull/49), usage of this latest version is included in this PR along with code changes to use jenks classification from it.


bundles/statistics/statsgrid2016/service/ClassificationService.js
```diff
@@ -105,10 +105,7 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationService',
             const setBounds = (stats) => {
                 if (opts.method === 'jenks') {
                     // Luonnolliset välit
-                    // geostats.js has performance problems when calculating jenks natural breaks
-                    // hence we are using different function implemented in GeostatsHelper for that
-                    const geostatsHelper = new GeostatsHelper();
-                    response.bounds = geostatsHelper.getJenks(stats.serie, opts.count);
+                    response.bounds = stats.getClassJenks(opts.count);
```
ZakarFin: Do we need geostatshelper after this at all? Could it be removed?

mikkomustikkamaa: There is no use for GeostatsHelper so it is now removed as obsolete.

PR #1035, opened 2019-11-06
Fix indicator layout

mikkomustikkamaa:
Implemented util method to cut sentences to given length and used it in thematic map indicator listing to prevent layout problems with long indicator names.


bundles/statistics/statsgrid2016/components/IndicatorList.js
```diff
@@ -66,8 +66,9 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.IndicatorList', function (servi
         var indicators = me._getIndicators();
         indicators.forEach(function (ind, id) {
             me.service.getUILabels(ind, function (labels) {
+                const regexFourDigitYearBetweenParenthesesAtTheEnd = /\(\d{4}\)$/;
                 var indElem = jQuery(me.__templates.indicator({
-                    name: labels.full,
+                    name: Oskari.util.cutSentenceToLength(labels.full, 60, regexFourDigitYearBetweenParenthesesAtTheEnd),
```
ZakarFin: You can use other keys from labels variable to reconstruct the parameters part and just "cut" the name part. This way you don't need the util method or regexp.

mikkomustikkamaa: Changed implementation to construct indicator name from indicator and params.

bundles/statistics/statsgrid2016/resources/scss/style.scss
```diff
@@ -959,9 +959,12 @@ background-color: #888888;
 /* ********************************************
  * Indicator list
  * ********************************************/
- ol.statsgrid-indicator-list li:nth-child(2n) {
-    background-color: #f3f3f3;
-}
+ ol.statsgrid-indicator-list {
+    min-width: 430px !important;
```
ZakarFin: Check if !important is really required. Otherwise looks good for merging.

mikkomustikkamaa: Removed !important since it is not required.

PR #1036, opened 2019-11-07
Ol6 tuning

ZakarFin:
Workaround for https://github.com/openlayers/openlayers/blob/master/changelog/upgrade-notes.md#the-view-extent-option-now-applies-to-the-whole-viewport. EPSG:3857 and 4326 projections are now limited to extent on viewport but other projections work as before (viewport can show outside extent but center can't go beyond extent). 

(Note: constrainCenterOnly mentioned in ol 6 upgrade notes is actually constrainOnlyCenter https://openlayers.org/en/latest/apidoc/module-ol_View-View.html)


jheijari:
Looks good.

PR #1038, opened 2019-11-07
Optimizing hover performance

jheijari:
Updating map hover to use OL 6 `getFeatures` function for hover hit detection.
`getFeatures` has better performance, but is less accurate than previously used `forEachFeatureAtPixel`.

### DIFF
--- a/bundles/mapping/mapmodule/service/VectorFeatureService.ol.js
+++ b/bundles/mapping/mapmodule/service/VectorFeatureService.ol.js
@@ -30,6 +30,7 @@ Oskari.clazz.defineES('Oskari.mapframework.service.VectorFeatureService',
             this.layerTypeHandlers = {};
             this.defaultHandlers = {};
             this._registerEventHandlers();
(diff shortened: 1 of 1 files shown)


ZakarFin:
Thanks! This was a good test but the functionality seems a bit off at least for point features with this change. Not sure if the reason is in OpenLayers since even on https://openlayers.org/en/latest/examples/hitdetect-vector.html the functionality feels "buggy" in a sense that when the cursor is not on any feature -> some feature is always "highlighted". Maybe we'll try this again on a later OL-version.

ZakarFin:
Still digging this old rabbit hole and found a related issue: https://github.com/openlayers/openlayers/issues/17288#issuecomment-3764415318

Apparently getFeatures() not probably worth it, but we could tinker around with:

```
 if (map.getView().getInteracting() || map.getView().getAnimating()) {
    return;
  }
```
And take a look how these relate to our own `sandbox.getMap().isMoving()` that is used when handling hovering and using `forEachFeatureAtPixel()`.

PR #1042, opened 2019-11-08
Add blink indicator to selected layers

jheijari:
Added `blink` to  `LayerViewTabs` state. When `true`, selected layers tabs badge starts blinking.


bundles/framework/layerlist/resources/scss/style.scss
```diff
@@ -1,5 +1,3 @@
-$filter-border-color: #a4a4a4;
-
 @keyframes blinking {
```
ZakarFin: These seem to repeat as a styled-component. Could these be removed as well?

jheijari: Unfortunately no, these are for the Tile.

bundles/framework/layerlist/view/LayerViewTabs/LayerViewTabs.jsx
```diff
@@ -51,7 +51,15 @@ export const LayerViewTabs = ({ tab, layerList, selectedLayers, autoFocusSearch,
             <TabPane tab={tabs.layerList} key={TABS_ALL_LAYERS}>
                 <LayerList ref={searchTermInputRef} {...layerList.state} mutator={layerList.mutator} locale={locale} />
             </TabPane>
-            <TabPane tab={<SelectedTab num={selectedLayers.state.layers.length} text={tabs.selectedLayers} />} key={TABS_SELECTED_LAYERS}>
+            <TabPane
+                tab={
+                    <SelectedTab
+                        key={selectedLayers.state.layers.length}
+                        blink={blink}
```
ZakarFin: I would argue that blink is not a good prop for this. It's more of an internal state/implementation specific thing for the tab to do WHEN the state changes. Blink as a name is not perfect either. Would rather see something like "indicateChange" but then again shouldn't it be part of the component to know if it should signal change or not based on num-prop (would count be a better name for num?) ? 

TL;DR: change blink prop name to indicateChange so the naming doesn't dictate how the change should be indicated. Or even better yet try baking the logic inside SelectedTab.

PR #1061, opened 2019-11-15
Refresh layer visibility on layer support changes

jheijari:
Fixes an issue where 3D layer was visible in layer selector when it was not supported (2D map).
Issue occurred when the layer was initially added to the map on page load.

### DIFF
--- a/bundles/framework/layerlist/view/LayerViewTabs/LayerViewTabsHandler.js
+++ b/bundles/framework/layerlist/view/LayerViewTabs/LayerViewTabsHandler.js
@@ -98,6 +98,19 @@ class UIStateHandler extends StateHandler {
                     this.getLayerListHandler().getFilterHandler().useStashedState();
                 }
             },
+            'MapLayerEvent': event => {
+                if (!['update', 'sticky', 'tool'].includes(event.getOperation())) {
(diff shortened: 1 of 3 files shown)


bundles/framework/layerlist/view/LayerViewTabs/LayerViewTabsHandler.js
```diff
@@ -98,6 +98,19 @@ class UIStateHandler extends StateHandler {
                     this.getLayerListHandler().getFilterHandler().useStashedState();
                 }
             },
+            'MapLayerEvent': event => {
+                if (!['update', 'sticky', 'tool'].includes(event.getOperation())) {
```
ZakarFin: Shouldn't the layers be updated after these as well? Or is it handled somewhere else?

ZakarFin: For example on update the layer name might have changed or with tool another tool have been added.

jheijari: Yes. The condition is actually negated, so update, sticky and tool will update the state.

PR #1063, opened 2019-11-18
Get camera request

Somac:
Request that gets current 3d map location and camera orientation and sends an event that handles this information.


ZakarFin:
Could these be added to mapmoveevent for 3d?

ZakarFin:
Remove GetMapCameraRequest and GetMapCameraEvent. Add camera info to AfterMapMoveEvent.

ZakarFin:
Good for merging. I'll merge after the 1.54 release.

PR #1069, opened 2019-11-20
Excluding core-js from transpiling should fix IE

ZakarFin:
According to https://github.com/zloirock/core-js/issues/514


ZakarFin:
Tested with paikkatietoikkuna.fi app that on current develop build the app _doesn't_ start on IE and with this change it _does_ start.

PR #1070, opened 2019-11-20
Fix bug with unexpected feature geometry

mikkomustikkamaa:
Fix bug where error occured if feature geometry didn't match any of the cases. Now in these situations javascript errors no longer occur, features can be seen in Feature Data flyout but feature cannot be visualized to map


ZakarFin:
GeometryCollection has geometries array. Maybe we should check the first geometry in the array and return a style based on that geometry type. Just that we get something visible on the screen instead of failing silently. Maybe we could log some info level stuff when this happens so it's easier to debug any issues caused by this using developer console.

PR #1075, opened 2019-11-25
ZoomLevel to altitude formula for 3d map

Somac:
Works with scale and number. The value is not 100% correct, cause 3d map also factors camera orientation to altitude level, but approximate of what the zoomLevel from 2d to 3d map is.

Formula taken from here (and fine tuned a bit):
https://stackoverflow.com/questions/36544209/converting-altitude-to-z-level-and-vice-versa/37142662#37142662

### DIFF
--- a/bundles/mapping/mapmodule/mapmodule.olcs.js
+++ b/bundles/mapping/mapmodule/mapmodule.olcs.js
@@ -170,50 +170,49 @@ class MapModuleOlCesium extends MapModuleOl {
         cam.moveStart.addEventListener(this.notifyStartMove.bind(this));
         cam.moveEnd.addEventListener(this.notifyMoveEnd.bind(this));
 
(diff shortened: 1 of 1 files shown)


bundles/mapping/mapmodule/mapmodule.olcs.js
```diff
@@ -528,14 +526,29 @@ class MapModuleOlCesium extends MapModuleOl {
...
     }
 
-    adjustZoom (zoom) {
+    _altitudeToZoom (altitude) {
+        const A = 40487.57;
+        const B = 0.00007096758;
+        const C = 91610.74;
+        const D = -40467.74;
+
+        return (D + (A - D) / (1 + Math.pow(altitude / C, B))) - 4;
```
ZakarFin: Add the link to the source here.

PR #1092, opened 2019-12-09
Make Message component localized

jheijari:
Wrapped `Message` with `withLocale` so it can be directly imported from `oskari-ui` and used without `bundleKey` property.

Removed `Message` property from `LocaleContext`.


ZakarFin:
Perfect!

PR #1096, opened 2019-12-10
Adding data providers and themes

jheijari:
Admin tools for adding data providers and map layer groups.
Contains commits from https://github.com/oskariorg/oskari-frontend/pull/1089


bundles/framework/divmanazer/extension/ExtraFlyout.js
```diff
@@ -98,6 +98,9 @@ Oskari.clazz.define('Oskari.userinterface.extension.ExtraFlyout',
                 } else {
                     me.options.container.append(popup);
                 }
+                if (me.options.showHeading === false) {
+                    popup.find('.oskari-flyoutheading').css('display', 'none');
```
ZakarFin: This makes a flyout that can't be closed or dragged by normal means other than something that that is inside the content closing it programmatically?

jheijari: Actually this is just a visual change taking off the topmost handle bar above the title row where the closing icon is. The flyout can still be closed by normal means.

ZakarFin: I think this is not something we need for this. Sure it's a small change but with an overall UI-refresh coming up I'm not comfortable adding more options to current UI-components.

bundles/admin/admin-layereditor/instance.js
```diff
@@ -56,16 +65,49 @@ Oskari.clazz.defineES('Oskari.admin.admin-layereditor.instance',
...
             this.sandbox.notifyAll(event);
+        }
 
+        /**
+         * Adds admin tools to layer list
+         */
+        _setupAdminTooling () {
             // add layerlist tool for adding new layers
             const toolingService = this.sandbox.getService('Oskari.mapframework.service.LayerListToolingService');
             if (toolingService) {
```
ZakarFin: Switch this to:
```
if (!toolingService) {
    return;
}
```
To get rid of the massive if block

bundles/admin/admin-layereditor/view/LocalizingFlyout.jsx
```diff
@@ -0,0 +1,175 @@
...
+        super();
+        this.getMsg = Oskari.getMsg.bind(null, instance.getName());
+        this.setState({
+            loading: false,
+            headerMessageKey: null,
+            value: {},
+            labels: {
+                en: this.getMsg('en.lang'),
+                fi: this.getMsg('fi.lang'),
+                sv: this.getMsg('sv.lang')
```
ZakarFin: Hardcoded languages. These should come from the Oskari env.

bundles/admin/admin-layereditor/instance.js
```diff
@@ -168,5 +237,86 @@ Oskari.clazz.defineES('Oskari.admin.admin-layereditor.instance',
...
+                headerMessageKey: 'dataProviderName'
+            });
+            this.dataProviderFlyout.makeDraggable({
+                handle: '.oskari-flyouttoolbar',
+                scroll: false
+            });
+            this.dataProviderFlyout.setAction(value => {
+                // TODO add discreet notifications
+                jQuery.ajax({
+                    type: 'PUT',
```
ZakarFin: Currently "action" is only for adding? I think we need some modifications for making editing work. Also we might want to add delete to the popup so this could be setSaveAction() or something like that?

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -603,6 +603,15 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
                 }
             });
         },
+
+        /**
+         * @method addLayerGroup
+         * @param {Oskari.mapframework.domain.MaplayerGroup} group map layer group to add.
+         */
+        addLayerGroup: function (group) {
+            this._layerGroups.push(group);
```
ZakarFin: This assumes no hierarchy but I guess its ok at this point.

bundles/admin/admin-layereditor/instance.js
jheijari: Indeed, the flyout can perform only single action at this point. Refactoring is required once we implement editing and deleting data.

bundles/framework/divmanazer/extension/ExtraFlyout.js
jheijari: Removed the option.

bundles/admin/admin-layereditor/instance.js
jheijari: Done.

bundles/admin/admin-layereditor/view/LocalizingFlyout.jsx
jheijari: Using Oskari.getSupportedLanguages

bundles/admin/admin-layereditor/instance.js
jheijari: Renamed `setAction` to `setSaveAction`.

PR #1098, opened 2019-12-11
Set time plugin

Somac:
Oskari.app.playBundle(
{
  bundlename : 'shadow-plugin-3d'
});

Shadowing plugin for 3d maps. Gets time and date from user and changes shadowing on map accordingly.


bundles/mapping/shadow-plugin-3d/instance.js
```diff
@@ -0,0 +1,62 @@
...
+            return this.sandbox;
+        },
+        start: function (sandbox) {
+            if (this._started) {
+                return;
+            }
+            this._started = true;
+            sandbox = sandbox || Oskari.getSandbox();
+            this.setSandbox(sandbox);
+            this._mapmodule = sandbox.findRegisteredModuleInstance('MainMapModule');
```
ZakarFin: We could check if the map module supports 3D and stop if it doesn't

bundles/mapping/shadow-plugin-3d/view/ShadowTool/ShadowTool.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+    return (
+        <Background>
+            <Row>
+                <Col>
+                    <StyledIcon type="calendar" style={{ color: '#d9d9d9', fontSize: '18px' }} />
+                    <StyledInput value={dateValue} onChange={changeDate} />
+                </Col>
+                <Col>
+                    <StyledButton onClick={setCurrentTime}>Nykyhetki</StyledButton>
```
Somac: I will replace this hardcoded string on the next ticket. Already made so much changes to this file on the next part so too much hassle merging these branches if I make the change here.

bundles/mapping/shadow-plugin-3d/plugin/ShadowingPluginClass.js
```diff
@@ -0,0 +1,188 @@
...
+
+    _addToMobileToolBar () {
+        const resetMapStateControl = jQuery('.toolbar_mobileToolbar').find('.mobile-reset-map-state');
+        jQuery(this._element).insertAfter(resetMapStateControl);
+    }
+
+    _addToPluginContainer () {
+        const panButtonsControl = jQuery('.mappluginsContent').find('.panbuttons');
+        jQuery(this._element).insertAfter(panButtonsControl);
+    }
```
ZakarFin: Use "index" to assign "priority"/placement among other plugins: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/coordinatetool/plugin/CoordinateToolPlugin.js#L29

bundles/mapping/shadow-plugin-3d/view/ShadowTool/ShadowTool.jsx
```diff
@@ -0,0 +1,75 @@
...
+    const validateDate = (target) => {
+        const matches = /^(0[1-9]|1[0-9]|2[0-9]|3[0-1]|[1-9])[/](0[1-9]|1[0-2]|[1-9])$/.exec(target);
+        if (matches === null) {
+            return false;
+        }
+        const d = parseInt(matches[1]);
+        const m = matches[2] - 1;
+        const dateObject = new Date(2019, m, d);
+        return dateObject.getDate() === d && dateObject.getMonth() === m;
+    };
```
ZakarFin: Add a comment that these are duplicated in the SetTimeRequestHandler. Perfect option would be to have some util.js on 3d-mapmodule that could be imported and used both here and in the requesthandler but the comment is enough for now.

ZakarFin: There will be a follow up PR that has changes to this so merging for now.

PR #1100, opened 2019-12-16
Add 3D camera controls to publisher

mikkomustikkamaa:
PR contains changes to handle 3D camera controls in publisher.

### DIFF
--- a/bundles/mapping/camera-controls-3d/tool/CameraControls3dTool.js
+++ b/bundles/mapping/camera-controls-3d/tool/CameraControls3dTool.js
@@ -1,0 +1,72 @@
+Oskari.clazz.define('Oskari.mapping.cameracontrols3d.CameraControls3dTool',
+    function () {
+    }, {
+        index: 3,
+        allowedLocations: ['top left', 'top right', 'bottom left', 'bottom right'],
+        lefthanded: 'top left',
+        righthanded: 'top right',
+        allowedSiblings: [
+            'Oskari.mapframework.bundle.featuredata2.plugin.FeaturedataPlugin',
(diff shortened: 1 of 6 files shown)


ZakarFin:
Does this work with editing an existing embedded map? I think CameraControls3dTool needs at least init method to detect if the tool should be on or not when initializing the publisher for editing.

ZakarFin:
Otherwise looks ok for merging.

bundles/mapping/camera-controls-3d/tool/CameraControls3dTool.js
```diff
@@ -0,0 +1,72 @@
...
+    *
+    * @returns {Object} tool value object
+    */
+        getValues: function () {
+            var me = this;
+
+            if (me.state.enabled) {
+                return {
+                    configuration: {
+                        mapfull: {
```
ZakarFin: This is not part of mapfull plugins but a bundle called "camera-controls-3d". So getValues() should return something different. The featuredata bundle works in the same way so you could look for an example here https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/framework/publisher2/tools/FeaturedataTool.js#L62

ZakarFin: You will also need to whitelist the bundle for publishing in oskari-server: https://github.com/oskariorg/oskari-server/blob/master/control-base/src/main/java/fi/nls/oskari/control/view/AppSetupHandler.java#L149

mikkomustikkamaa: Thank you for the tips. Camera controls are now handled as a bundle. Also https://github.com/oskariorg/oskari-server/pull/497 is created to whitelist the bundle.

PR #1101, opened 2019-12-17
Improve 3D camera rotate mode

mikkomustikkamaa:
Target to look at is now determined with ray from globe surface. This leads to similar rotate experience compared to situation when rotate is done in default move mode using keyboard Ctrl pressed down. Also constrainedAxis handling is removed as obsolete.

### DIFF
--- a/bundles/mapping/mapmodule/mapmodule.olcs.js
+++ b/bundles/mapping/mapmodule/mapmodule.olcs.js
@@ -538,10 +538,10 @@ class MapModuleOlCesium extends MapModuleOl {
 
         if (Cesium.defined(lookAtTarget)) {
             camera.lookAt(lookAtTarget, camera.position);
+            this._disableMapMoveControls();
         } else {
             this._log.warn('LookAtTarget cannot be determined. Maybe looking at space.');
         }
-        this._disableMapMoveControls();
     }
 
     setCameraToMoveMode () {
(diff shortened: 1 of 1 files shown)


bundles/mapping/mapmodule/mapmodule.olcs.js
```diff
@@ -524,21 +525,27 @@ class MapModuleOlCesium extends MapModuleOl {
...
+        ));
+
+        const lookAtTarget = scene.globe.pick(ray, scene);
+
+        if (Cesium.defined(lookAtTarget)) {
+            camera.lookAt(lookAtTarget, camera.position);
+        } else {
+            this._log.warn('LookAtTarget cannot be determined. Maybe looking at space.');
+        }
         this._disableMapMoveControls();
```
ZakarFin: Should _disableMapMoveControls() call be in the if with lookAt()?

mikkomustikkamaa: I thought that in those situation user probably needs to choose "reset to original state" anyway. But i can move disable inside if.

mikkomustikkamaa: In might have shot my self in the leg with using useState in CameraControls3d.jsx because active icon is also set to rotate in these failing cases because mentioned react component does not currently get any response about did operation actually succeeded. Probably would be better to extract state from React component to mapmodule.olcs or separate Handler and implement notify loop where component gets active mode as prop.

PR #1106, opened 2019-12-20
Set time plugin timeline and playing feature

Somac:
Adds playing and timeline sliders to set time plugin.

Some problems with statehandling right now so opened as a draft.


bundles/mapping/shadow-plugin-3d/plugin/ShadowingPluginClass.js
```diff
@@ -46,15 +45,6 @@ class ShadowingPlugin extends BasicMapModulePlugin {
         this._createUI(mapInMobileMode);
     }
 
-    _createEventHandlers () {
-        return {
-            TimeChangedEvent: function (event) {
```
ZakarFin: This would work better with RPC etc programmatic changes to time done outside this UI.

PR #1108, opened 2020-01-01
Admin form localized inputs

jheijari:
Refactored localized layer names and descriptions in admin layer form to make use of the `LocalizationComponent`.

Contains commits from https://github.com/oskariorg/oskari-frontend/pull/1107


bundles/admin/admin-layereditor/view/AdminLayerForm/GeneralTabPane.jsx
```diff
@@ -42,10 +63,18 @@ const GeneralTabPane = (props) => {
...
-                </StyledComponent>
+                <LocalizationComponent
+                    labels={localized.labels}
+                    value={localized.values}
+                    languages={Oskari.getSupportedLanguages()}
+                    onChange={controller.setLocalizedNames}
+                    LabelComponent={PaddedLabel}
+                >
+                    <TextInput type='text' name='name'/>
+                    <Padding/>
```
ZakarFin: <Padding /> is used because inputs inside LocalizationComponent need to be at the "root level"? So we can't just have a StyledTextInput with the padding? There should probably be a comment about this if that is the reason. Otherwise it will get "fixed"/broken in near future.

jheijari: Yes, that's why. And we could make a custom TextInput, but then the style should have padding settings with *!important* since ant-design overrides it otherwise with 0.

PR #1112, opened 2020-01-08
Layer type support

jheijari:
Supporting different layer types in admin form.
A new `LayerComposingModel` (or `LayerModelComposer`?) class to describe layer types for the admin form. LayerPlugins register `LayerComposingModels` for editable layer types.


bundles/admin/admin-layereditor/view/AdminLayerForm/AdditionalTabPane.jsx
```diff
@@ -1,37 +1,57 @@
...
+    const attributesInput =
+        <Fragment>
+            <Message messageKey='attributes'/>
+            <StyledComponent>
+                <TextAreaInput
+                    rows={6}
+                    value={JSON.stringify(layer.attributes || {}, null, 2)}
+                    onChange={(evt) => controller.setAttributes(evt.target.value)} />
+            </StyledComponent>
+        </Fragment>;
```
ZakarFin: Maybe premature optimization but if we would replace metainfoInput, gfiContentInput and attributesInput with method calls or components outside the render function they would not be created even in memory if they are not needed?

PR #1119, opened 2020-01-13
Use credentials in layer wizard step 2

mikkomustikkamaa:
This pull request contains following improvements to admin layer wizard step 2:

- Credentials section from UrlInput component is displayed to user. Inserted username and password values are stored to AdminLayerFormHandler and passed in capabilities fetching request to backend.
- Error handling is added and error messages are displayed to user at the top of flyout when needed. In case of HTTP 401 also response credentials collapse is opened.
- Some minor layout fixes

### DIFF
--- a/bundles/admin/admin-layereditor/view/LayerWizard.jsx
+++ b/bundles/admin/admin-layereditor/view/LayerWizard.jsx
@@ -53,7 +53,7 @@ function getStep (layer) {
 const LayerTypeTitle = ({ layer, LabelComponent }) => (
     <React.Fragment>
         <Message messageKey='wizard.type' LabelComponent={LabelComponent} />
-        { layer.type && <React.Fragment><div style={{ display: 'inline' }}>:</div><div>{layer.type}</div></React.Fragment>}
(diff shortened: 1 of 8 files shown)


bundles/admin/admin-layereditor/view/LayerWizard.jsx
```diff
@@ -44,9 +53,10 @@ function getStep (layer) {
 const LayerTypeTitle = ({ layer, LabelComponent }) => (
     <React.Fragment>
         <Message messageKey='wizard.type' LabelComponent={LabelComponent} />
-        { layer.type && `: ${layer.type}` }
+        { layer.type && <React.Fragment><div style={{ display: 'inline' }}>:</div><div>{layer.type}</div></React.Fragment>}
```
ZakarFin: `<span>` instead of `<div style={{ display: 'inline' }}>`?

ZakarFin:
After the span-fix this is ready for merging

bundles/admin/admin-layereditor/view/LayerWizard.jsx
mikkomustikkamaa: Fixed to use span as suggested

PR #1129, opened 2020-01-16
Simplify localized name handling

jheijari:
Remove complexity


jheijari:
Hold on with this for a while, something not right with it

jheijari:
Now it's golden!

PR #1137, opened 2020-01-20
Admin default style option

jheijari:
Localized default style option based on wfs bundle.


jheijari:
Closing as outdated and irrelevant.

PR #1139, opened 2020-01-20
Payload and other fixes

jheijari:
Refactored `LayerHelper`.

Fixed issues with styles, data provider, groups etc.
Removed an unmaintained storybook story.


bundles/admin/admin-layereditor/view/AdminLayerForm/GeneralTabPane/DataProvider.jsx
```diff
@@ -9,7 +9,7 @@ export const DataProvider = ({ layer, dataProviders, controller }) => (
         <Message messageKey='dataProvider'/>
         <StyledComponent>
             <Select
-                value={layer.organizationName}
+                value={layer.groupId}
```
ZakarFin: Should this be dataProviderId instead so it's not confused to layer group id?

jheijari: Yes, I'll refactor it. I'll also change mapLayerGroups to just groups on layer data.

PR #1140, opened 2020-01-21
Add data provider and theme editing

mikkomustikkamaa:
With changes in this pull request, admin users can edit themes and data providers.


bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -79,7 +83,11 @@ class ViewHandler extends StateHandler {
     updateLayerGroups () {
         const { searchText, activeId: filterId } = this.filter;
         const layers = filterId ? this.mapLayerService.getFilteredLayers(filterId) : this.mapLayerService.getAllLayers();
-        let groups = groupLayers([...layers], this.groupingMethod);
+        const tools = Object.values(this.toolingService.getTools()).filter(tool => tool.getTypes().includes('layergroup'));
+        const isUserAdmin = tools.length > 0;
+        const themes = isUserAdmin ? this.mapLayerService.getAllLayerGroups() : [];
```
ZakarFin: Non-admins get an empty array? Does this work ok? Seems a bit misleading atleast. These are the empty groups that will be appended/prepended to groups that are determined from layers?

mikkomustikkamaa: Yes this logic relates to feature to show empty groups for admin users. There are logic in https://github.com/oskariorg/oskari-frontend/blob/2811e5d72769349934f614cd2a033c91aa19e8fd/bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/util.js#L54
to add themes and data providers which does not contain layers at the top of layerlist. Idea here is that for admin users all themes and data providers are provided to groupLayers funtion to achieve mentioned result. For non admin users empty list is provided as a themes and dataproviders parameter and with this, only dataproviders and themes determined from layers are shown. According to testing should work properly.

ZakarFin: Should these be renamed with like dataProvidersIncludingEmpty etc for clarity? At some point I would like to refactor the whole structure to be based on this information with layers being injected into the groups instead of the current code where groups are determined from the layers. Now these are used just for getting info on any empty groups. Or maybe add a comment to clarify this? Both here and in groupLayers()?

mikkomustikkamaa: Comments added and variables renamed to clarify this.

PR #1159, opened 2020-01-28
removed functions that don't work with new WFS

osqq:
This can be ignored if there is a chance coming to this.

WFSLayer doesn't get "clickedGeometries" with new WFS because they are set by Mediator, which is not currently used.


ZakarFin:
Awesome. Please check the conflict though so it can be merged.

PR #1163, opened 2020-01-29
Data provider delete

mikkomustikkamaa:
This pull request contains logic to delete data providers. Remove button is added to Edit group flyout. On remove click confirmation popup is shown to user. If map layer count in group is greater than zero, checkbox is shown to user in confirmation popup with text "Delete the map layers in the group (x)" where x is count of map layers belonging to the group. When checkbox checked, map layers are removed from system. When checkbox is unchecked, data provider to be deleted is removed from the map layers and map layers remain in the system.


bundles/admin/admin-layereditor/view/LocalizingFlyout.jsx
```diff
@@ -158,10 +185,17 @@ const LocalizedContent = ({ loading, labels, value, headerMessageKey, controller
             >
                 <TextInput />
             </LocalizationComponent>
+            {id && layerCountInGroup > 0 &&
+                <DeleteLayersCheckbox checked={deleteLayers} onChange={ evt => controller.setDeleteLayers(evt.target.checked)}>{deleteMapLayersText + ' (' + layerCountInGroup + ')'}</DeleteLayersCheckbox>
```
ZakarFin: We should add a confirmation popup for delete and add this as an option to the confirmation. Now it's a bit scary looking field in the form if the user just wants to update the name. It could also be just 2 different buttons on the popup. One that deletes the provider including layers and one that only removes the provider and moves layers to the "ungrouped" group.

mikkomustikkamaa: Confirmation popup is now added.

PR #1164, opened 2020-01-29
Map scales to zoom levels, improve styling

jheijari:
Scale steps based on zoom levels.


bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
```diff
@@ -11,6 +11,7 @@ const DEFAULT_TAB = 'general';
 class UIHandler extends StateHandler {
     constructor (consumer) {
         super();
+        const mapmodule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
```
ZakarFin: I'm not sure if we are already too deep in the rabbit hole with this one but I would much prefer that these kind of things were passed as constructor params instead of accessed through the global. Just to make it more testable.

But we can refactor it later as there is the existing mapLayerService here as well.

PR #1177, opened 2020-01-31
Use layer params to create request url

jheijari:


ZakarFin:
Related to oskariorg/sample-server-extension#6

PR #1183, opened 2020-01-31
Fix single permission set for new layers

jheijari:


ZakarFin:
Fixed test in https://github.com/oskariorg/oskari-frontend/commit/fcdede5e7fc715705b1b76251e653dd2ae0a73a5

PR #1185, opened 2020-02-06
add camera settings to params

hvisuri:
Return also camera settings with params


ZakarFin:
Tuned it a bit to not confuse people only using 2D: https://github.com/oskariorg/oskari-frontend/commit/09e0ee84ba9a6de064dbbc334f3f287bd890baa3

PR #1196, opened 2020-02-23
removed mozilla specific css params from hierarchical scss

osqq:
Removed mozilla specific CSS params that caused elements to be misplaced


ZakarFin:
Resolves https://github.com/oskariorg/oskari-docs/issues/156

PR #1197, opened 2020-02-24
Layerlist 3d layer filter

okauppinen:
Add filter for 3d layers and button (option for select). LayerlistService isn't ready/registered when mapmodule's plugins are started/registered so made lazyStartPlugin functionality. I'm not sure if this is the right way to handle this.

Localization doesn't have tooltip text for 3d filter. Layerlist select options have both antd tooltip and normal tooltip. We should check tooltips and add tooltip text for 3d.


ZakarFin:
I'm not comfortable introducing new API-methods to mapplugins at the moment as the API is kinda messy already. I would rather have the plugin listen to `app.start` event and trying to add the filter on it's own without being "lazy" from mapmodule perspective:

            Oskari.on('app.start', function (details) { ... try adding filter ...     });

PR #1212, opened 2020-03-02
Error handling for opacity slider (in wrong place)

ZakarFin:
Opacity in map link parameters are thrown in as strings to opacity slider which prints out 20 lines of error output as it expects to receive a number. This fixes the issue but it should be fixed at its root.


ZakarFin:
Seems I can't reproduce this anymore so closing as not relevant.

PR #1244, opened 2020-03-27
Statistics diagram series data fix

okauppinen:
Changed diagram to use own getIndicatorData instead of service getCurrentDataset. getCurrentDataset doesn't work if there is more than one series and selected value is outside of value range. Diagram shows data only for active indicator so it doesn't need whole dataset. Even data table doesn't use services getCurrentDataset. Maybe we should remove it.

ClassificationChangedEvent had getPrevious method but previous classification wasn't set. Changed to getChanged which returns updated/changed object. User can update only single values with classification plugin (or min,max pair with slider). It's easy to catch changed value and components doesn't need whole previous classification. So component doesn't have to check if specific value is updated (current.value !== previous.value). Data table is re-rendered only if fractionDigits is changed.

Some event listeners were missing or changed to listen parameter changed instead of active indicator.

Added selected effect for data table header to get fast response for user click.

### DIFF
--- a/bundles/statistics/statsgrid2016/components/Datatable.js
+++ b/bundles/statistics/statsgrid2016/components/Datatable.js
@@ -400,6 +400,11 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Datatable', function (sandbox, 
             log.debug('Indicator parameter changed! ');
             me._handleRegionsetChanged();
         });
+        this.service.on('StatsGrid.ClassificationChangedEvent', function (event) {
+            if (event.getChanged().hasOwnProperty('fractionDigits')) {
+                me._handleRegionsetChanged();
+            }
+        });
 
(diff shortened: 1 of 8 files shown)


bundles/statistics/statsgrid2016/components/Datatable.js
```diff
@@ -399,6 +400,11 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.Datatable', function (sandbox,
             log.debug('Indicator parameter changed! ');
             me._handleRegionsetChanged();
         });
+        this.service.on('StatsGrid.ClassificationChangedEvent', function (event) {
+            if (event.getChanged().hasOwnProperty('fractionDigits')) {
+                me._handleRegionsetChanged();
```
ZakarFin: Maybe we should rename _handleRegionsetChanged() to something _updateWholeUI() or something like that? But we can do it later.

PR #1253, opened 2020-03-30
Try throttling log-messages

ZakarFin:
To increase logging performance.

This is work-in-progress, duplicates code from Oskari.util and generally just moves the "stop the world" logging moment to the future. Not production ready but something to tinker with in the future while code is compiling.


ZakarFin:
We haven't needed this so closing as unnecessary.

PR #1254, opened 2020-04-01
Screenshot fix

okauppinen:
html-to-image works only with Firefox and Chrome.

https://github.com/openlayers/openlayers/pull/10669/files


ZakarFin:
Restores getScreenshot() support for IE.

PR #1264, opened 2020-04-09
Statistics tile handling

okauppinen:
Changed divmanazer to close other attached tiles which doesn't have flyout. Now statistics tile is closed when other extension is attached. 

When tile is closed search flyout is closed. Search flyout is shown when tile is attached and no indicator's selected. Table, diagram and own inicator flyouts are closed when statistics state is reseted (layer or last indicator removed).


ZakarFin:
Looks good but I'll wait merging until there's time to test this more.

PR #1268, opened 2020-04-24
Camera3d guidedtour

okauppinen:
Add guided tour for 3D camera tools


bundles/mapping/camera-controls-3d/instance.js
```diff
@@ -43,5 +56,37 @@ Oskari.clazz.defineES('Oskari.mapping.cameracontrols3d.instance',
...
+                }
+            }
+
+            function handler (msg) {
+                if (msg.id === 'guidedtour') {
+                    sendRegister();
+                }
+            }
+
+            const tourInstance = me.sandbox.findRegisteredModuleInstance('GuidedTour');
```
ZakarFin, line 84: Could `sandbox.hasHandler('Guidedtour.AddToGuidedTourRequest')` be used here instead of checking for the bundle?

PR #1269, opened 2020-04-24
GetinfoPlugin wfs result fix

okauppinen:
Publisher stops and restarts plugins. When getinfo plugin is restarted it gets mapclickedevent after VectorLayerPlugin. VectorLayerPlugin gets gfi result faster than wms because it gathers result in frontend. GetinfoPlugin mapclick closes opened wfs popup.  Added location check for mapclick to fix this issue. If popup is opened in same location than mapclick, GetInfoPlugin doesn't close popup.

Fixed issue where wfs result pan map cancels wms requests. Refresh didn't pan map.
Fixed issue with renderPopup options param.


ZakarFin:
Feels kinda hacky with the click location stuff (that was already there) but if it works I guess it's ok for now.

PR #1283, opened 2020-05-07
Build improvements

ZakarFin:
And updated jsts to 2.2.0.

Tried to minimize code that is run through babel to improve build time. Changed config so instead of excluding some dependencies from transpiler the build now uses a list that should be included from node_modules. Namely 'oskari-frontend', 'oskari-frontend-contrib', 'jsts', 'olcs'. These all include ES6 code that need to be transpiled before they are passed to UglifyPlugin. Adding the olcs that is only needed for 3D apps makes the build take 3x or more time. So now we know what is making the build slow but not yet how to handle this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "imports-loader": "0.8.0",
     "jest": "24.9.0",
     "jquery": "3.5.1",
-    "jsts": "2.0.8",
+    "jsts": "2.2.0",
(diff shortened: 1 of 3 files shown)


ZakarFin:
This shouldn't change anything for the user but can affect apps built on Oskari. Merging so we can test it.

PR #1289, opened 2020-05-22
Printout fix custom styles

okauppinen:
Fix custom styles handling.  

It's was annoying that user can save style without name. It was easier to add Own style 1.. name than add error message.

Also moved id from style to metadata.

### DIFF
--- a/bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin.ol.js
+++ b/bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin.ol.js
@@ -386,7 +386,7 @@ export class WfsVectorLayerPlugin extends AbstractMapLayerPlugin {
         let name = this.visualizationForm.getOskariStyleName();
         if (!name) {
             const existing = this.userStyleService.getUserStylesForLayer(layerId);
-            name = Oskari.getMsg('MapWfs2', 'own-style') + ' ' + existing.length + 1;
+            name = Oskari.getMsg('MapWfs2', 'own-style') + ' ' + (existing.length + 1);
         }
         const styleWithMetadata = { name, style, id };
         layer.saveUserStyle(styleWithMetadata);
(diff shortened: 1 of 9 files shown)


bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin.ol.js
```diff
@@ -380,14 +380,16 @@ export class WfsVectorLayerPlugin extends AbstractMapLayerPlugin {
...
-        const styleWithMetadata = {
-            name: name,
-            style: oskariStyle
-        };
+    saveUserStyle (layer, id) {
+        const style = this.visualizationForm.getOskariStyle();
+        const layerId = layer.getId();
+        let name = this.visualizationForm.getOskariStyleName();
+        if (!name) {
+            const existing = this.userStyleService.getUserStylesForLayer(layerId);
```
ZakarFin, line 388: If this would always return an array the line for count could always be  existing.length + 1

bundles/mapping/mapwfs2/service/UserStyleService.js
```diff
@@ -9,7 +9,7 @@ export class UserStyleService {
             return;
         }
         const layerStyles = this.styles.get(layerId) || [];
```
ZakarFin, line 11: Call this.getUserStylesForLayer(layerId) and change it to `return this.styles.get(layerId) || [];`. This will remove a lot of if statements from the code.

bundles/mapping/mapwfs2/service/UserStyleService.js
```diff
@@ -32,7 +32,7 @@ export class UserStyleService {
 
         const layerStyles = this.styles.get(layerId);
```
ZakarFin, line 33: Use `this.getUserStylesForLayer(layerId)` and remove the following `if (layerStyles) {` block.

ZakarFin:
Other than cleaning up getUserStylesForLayer() this looks good for merging.

PR #1290, opened 2020-05-22
Printout fix full scale preview

okauppinen:
Use printmap and png on preview click


okauppinen:
https://github.com/oskariorg/oskari-frontend/pull/1291

PR #1300, opened 2020-06-05
Fixes to Lodash changes overwriting the global variable

ZakarFin:
Turns out doing `import { cloneDeep, sortBy } from 'lodash';` writes/modifies the `_ ` global variable with full lodash (of the version in package.json) overwriting the (other version) one we link from libraries here: https://github.com/oskariorg/oskari-frontend/blob/1.55.2/packages/framework/bundle/oskariui/bundle.js#L49-L51 which causes all sorts of compatibility issues.


ZakarFin:
Fixes issues caused by #1297. Might cause problems with markers but lets deal with those if they surface.

PR #1305, opened 2020-06-17
Admin single tile

okauppinen:
Add single tile option for admin layer editor for wms layers


bundles/admin/admin-layereditor/view/AdminLayerForm/AdditionalTabPane/SingleTile.jsx
```diff
@@ -0,0 +1,30 @@
...
+import { InfoTooltip } from '../InfoTooltip';
+import { StyledFormField } from '../styled';
+import styled from 'styled-components';
+
+const SpacedLabel = styled('div')`
+    display: inline-block;
+    margin-left: 10px;
+`;
+
+export const SingleTile = LocaleConsumer(({ layer, controller }) => {
```
ZakarFin, line 14: I think this doesn't need to be wrapped in LocaleConsumer for it to work. You only need LocaleConsumer if you need the getMessage() function but since this uses Message-tag (like it should) this should work without the wrapper.

PR #1306, opened 2020-06-17
Mapfull add selected layers 

okauppinen:
Try to add selected layers which are not available (myplaces, userlayer, analysis) when app is started.


ZakarFin:
With this we could remove the special handling for myplaces layers in myplaces bundle?

PR #1318, opened 2020-08-12
Fix map layer group/data provider admin popup height

ZakarFin:
Don't set height since it varies depending on number of locales available


ZakarFin:
Fixes oskariorg/oskari-docs#195

PR #1319, opened 2020-08-13
Publisher map tool fix

okauppinen:
Add tool button if config buttons includes tool. Fixes the issue where selected tools/buttons doesn't appear and unselected does to published map.


ZakarFin:
This was broken in the lodash removal. The original line was `if (_.indexOf(me._buttons, tool) !== -1) {` so looks like it was reverted on the removal.

ZakarFin:
Merging to develop as this is easy to cherry-pick for a hotfix if we decide to do one.

PR #1320, opened 2020-08-20
Fix coordinate check

jheijari:
Fixes a bug in `MapMoveRequestHandler`. When move is requested in other srs than the map's native srs, request fails on bound check.


ZakarFin:
Awesome, thanks!

PR #1321, opened 2020-08-27
Fixed defined scale selection check

kuosman:
This PR fixes JavasScript error that comes up when printing with defined scale selection (print not opened because selection check cause JavaScript exception).


ZakarFin:
Thanks

PR #1334, opened 2020-09-17
Printout load file with post request

okauppinen:
Load file with post request instead of opening preview window. If pdf has title then use it for filename otherwise print.pdf (before this: action.pdf). 

Post method is used when WFS layer has custom style. Fixes the issue with Chrome pdf viewer which tries to download file from url (before this used form with post method). 

Removed layout params, tile and table data and geojson from request because server doesn't use them. Didn't remove methods (e.g. _stringifyTileData())

"Chrome's built in PDF viewer will download the pdf file through the PDF's origin URL. So if the PDF is generated at server runtime and if it's not stored in the sever, the download could fail"

### DIFF
--- a/bundles/framework/printout/view/BasicPrintout.js
+++ b/bundles/framework/printout/view/BasicPrintout.js
@@ -632,6 +632,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.printout.view.BasicPrintout',
             };
             const errorCb = error => {
                 Oskari.log('BasicPrintout').error(error);
+                const popup = Oskari.clazz.create('Oskari.userinterface.component.Popup');
+                popup.show(this.loc.error.title, this.loc.error.saveFailed, [popup.createCloseButton()]);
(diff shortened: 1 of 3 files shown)


bundles/framework/printout/view/BasicPrintout.js
```diff
@@ -598,36 +598,43 @@ Oskari.clazz.define('Oskari.mapframework.bundle.printout.view.BasicPrintout',
...
+                    elem.download = fileName;
+                    document.body.appendChild(elem);
+                    elem.click();
+                    document.body.removeChild(elem);
+                    window.URL.revokeObjectURL(url);
+                }
+                showSpinner(false);
+            };
+            const errorCb = error => {
+                Oskari.log('BasicPrintout').error(error);
```
ZakarFin, line 634: Should we give the user some friendlier error message? Now just the spinner stops and dev-console gets a message.

PR #1342, opened 2020-10-14
Publisher select width

okauppinen:
Language select was too wide if application didn't have printout bundle. Updated select to 80% and fixed printout css leakage.


ZakarFin:
Nice catch!

PR #1344, opened 2020-10-14
New MyPlaces form

kasparia:
Replace old jQuery form with new GenericForm React-component


bundles/framework/myplaces3/view/MainView.js
```diff
@@ -162,25 +164,32 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.MainView',
          * @param {Oskari.mapframework.bundle.myplaces3.model.MyPlace} place prepoluate form with place data (optional)
          */
         showPlaceForm: function (location, place) {
-            const me = this;
             let layerId;
+            const me = this;
             const sandbox = this.instance.sandbox;
+            const categoryHandler = this.instance.getCategoryHandler();
+            this.categories = categoryHandler.getAllCategories();
```
ZakarFin, line 171: If there's not a good reason to use `this` just do a `const` instead.

bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -9,49 +13,82 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
      * @method create called automatically on construction
      * @static
      */
-
-    function (instance, options) {
+    function (instance, options, categories) {
```
ZakarFin, line 16: Looks like `instance` isn't really needed here. Provide a saveCallback instead.

bundles/framework/myplaces3/view/MainView.js
```diff
@@ -107,8 +107,10 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.MainView',
                 if (!event.getSticky()) {
                     return;
                 }
-                // changed tool
+
+                // changed tool -> clean popup
                 this.cleanupPopup();
+                this.form.destroy();
```
ZakarFin, line 113: Should form.destroy() be in cleanPopup()?

src/react/components/GenericForm.jsx
```diff
@@ -10,6 +11,49 @@ import 'antd/es/input/style/index.js';
 
 const { TextArea } = Input;
 
+const zIndexValue = 999999;
```
ZakarFin, line 14:
```suggestion
// If the form is shown on popup the Select dropdown opens behind popup without this
// FIXME: this will probably not work with modal popups (dropdown might be over the modal overlay)
const zIndexValue = 999999;
```

src/react/components/GenericForm.jsx
```diff
@@ -18,112 +62,180 @@ const { TextArea } = Input;
...
- */
-const createFormInput = (field) => {
-    if (!field) {
-        return null;
+                    >
+                        { field.value.map((singleOption) => {
+                            return (
+                                <Select.Option
+                                    value={ singleOption.categoryId }
+                                    key={ singleOption.categoryId }
```
ZakarFin, line 152: `option.categoryId` doesn't make sense on a generic form. Use `option.value` instead and map it accordingly on myplaces functionality.

src/react/components/GenericForm.jsx
```diff
@@ -18,112 +62,180 @@ const { TextArea } = Input;
...
+                                >
+                                    { singleOption.name }
```
ZakarFin, line 154: To make this more generic 
```suggestion
                                    { singleOption.label || singleOption.name || singleOption.value }
```

src/react/components/GenericForm.jsx
ZakarFin: Maybe even support non-object options? 
```suggestion
                                    key={ singleOption.value || singleOption }
```

src/react/components/GenericForm.jsx
```diff
@@ -18,112 +62,180 @@ const { TextArea } = Input;
...
-        formSettings,
-        fields
-    } = props;
 
-    const [form] = Form.useForm();
+export class GenericForm extends React.Component {
+    constructor (props) {
+        super(props);
+        
+        this.formRef = React.createRef();
```
ZakarFin, line 70: I'm not sure about using ref. Can you tell a bit more why we need to use it here?

bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -9,49 +13,82 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
...
+            },
+            () => ({
+                validator: (_, value) => value ? Promise.resolve(value) : Promise.reject(new Error('Must validate'))
+            })
+        ];
+
+        // Default form settings
+        this.defaultProps = {
+            formSettings: {
+                label: 'Form settings label',
```
ZakarFin, line 81: Localize label OR remove it?

bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -9,49 +13,82 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
...
+                    if (typeof value !== 'undefined'  && value !== '') {
+                        return Oskari.util.sanitize(value) === value ? Promise.resolve(value) : Promise.reject(new Error(this.loc('validation.descIllegal')));
+                    } else {
+                        return Promise.resolve('');
+                    }
+                }
+            })
+        ];
+
+        this.testRules = [
```
ZakarFin, line 68: This is for testing? Consider moving validator rules to different file that can be imported here.

ZakarFin:
Only english translation has been updated. Does fi/sv need to be changed as well or do they make sense without changes?

ZakarFin:
PlaceForm.js feels a bit clunky at the moment but if it works I guess it's a step to the right direction anyways.

bundles/framework/myplaces3/view/MainView.js
```diff
@@ -163,60 +164,30 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.MainView',
...
-                actions,
-                layerId
-            }];
 
-            var options = {
-                hidePrevious: true
-            };
-            var request = Oskari.requestBuilder('InfoBox.ShowInfoBoxRequest')(this.popupId, me.loc('placeform.title'), content, location, options);
-            sandbox.request(me.getName(), request);
+            me.form.getForm(categories, place); // Get form
```
ZakarFin, line 189: The getForm() is a misleading name since the return value is not used. Does it show the form?

bundles/framework/myplaces3/view/MainView.js
kasparia: Changed to `const`.

bundles/framework/myplaces3/view/MainView.js
kasparia: Moved to cleanPopup().

bundles/framework/myplaces3/view/PlaceForm.js
kasparia: Removed

bundles/framework/myplaces3/view/PlaceForm.js
kasparia: Removed for now

src/react/components/GenericForm.jsx
kasparia: Removed ref since we don't need that anymore with `initialValue` prop.

bundles/framework/myplaces3/view/MainView.js
kasparia: Changed `getForm` -> `showForm`

bundles/framework/myplaces3/view/PlaceForm.js
kasparia: Replaced with `saveCallback`

src/react/components/GenericForm.jsx
```diff
@@ -10,120 +11,226 @@ import 'antd/es/input/style/index.js';
...
+                        key={ fieldKey }
+                        placeholder={ field.placeholder }
+                        dropdownStyle={{ zIndex: zIndexValue }}
+
+                    >
+                        { field.value.map((singleOption) => {
+                            return (
+                                <Select.Option
+                                    value={ singleOption.value }
+                                    key={ singleOption.label || singleOption.value }
```
ZakarFin, line 149: We fallback to support non-object options for label. We should do it here as well.
```suggestion
                                    value={ singleOption.value || singleOption }
                                    key={ singleOption.label || singleOption.value || singleOption }
```

bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -177,35 +198,15 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
...
-                    measurementDiv.html(this.loc('placeform.measurement.' + drawMode) + ' ' + measurement);
-                } else {
-                    measurementDiv.remove();
-                }
-            }
+            // const measurement = place.getMeasurement();
             this.place = place;
         },
-        setMeasurementResult: function (measurement, drawMode) {
+        setMeasurementResult: function (measurement, drawMode, instance) {
```
ZakarFin, line 204: This should be refactored to take a single param for the pre-formatted measurement. We can remove drawMode and instance params, do the formatting on the caller and have this be more simple/testable without external dependencies.

bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -301,5 +286,176 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
...
+            this._initializePlace(); // initialize place so we have empty place to fill on
+            this._populateForm(); // populate form with data from place
+
+            this.dialog = Oskari.clazz.create('Oskari.userinterface.component.Popup'); // Create popup dialog
+            this.dialog.makeDraggable();
+
+            // add new dialog to ui
+            this.dialog.show(this.loc('placeform.title'), '<div class="places-edit-dialog"></div>');
+
+            [this.container] = jQuery('.places-edit-dialog');
```
ZakarFin, line 312: You can get a reference to the popup content from the popup. No need to search the whole document for it and possibly hitting the wrong element:
```suggestion
            [this.container] = this,dialog.getJqueryContent().find('.places-edit-dialog');
```

Also I'm not sure if it's a good idea to store another reference to the element here. If it's not required to have the container as a variable for other purposes you could just send the reference as a parameter for `this._renderForm(´)`

bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -301,5 +286,176 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
...
+
+            this._renderForm();
+
+            this.dialog.moveTo('div.personaldata ul li select', 'right');
```
ZakarFin, line 316: Just to be sure we target a select in the right tab?
```suggestion
            this.dialog.moveTo('div.personaldata .tab-content.myplaces ul li select', 'right');
```

bundles/framework/myplaces3/view/PlaceForm.js
kasparia: Refactored as suggested.

bundles/framework/myplaces3/view/PlaceForm.js
kasparia: Got rid of `container` variable and passed it straight to renderForm method.

PR #1351, opened 2020-10-23
Request for layer deletion through RPC

kasparia:


bundles/mapping/mapmodule/plugin/vectorlayer/VectorLayerPlugin.ol.js
```diff
@@ -590,6 +590,16 @@ Oskari.clazz.define(
                     var evt = Oskari.eventBuilder('MapLayerEvent')(layer.getId(), 'update');
                     this._sandbox.notifyAll(evt);
                 }
+                if (options.remove) {
+                    const request = Oskari.requestBuilder('RemoveMapLayerRequest')(layer.getId());
+                    this._sandbox.request(this, request);
+
+                    const mapLayerService = this._sandbox.getService('Oskari.mapframework.service.MapLayerService');
```
ZakarFin, line 597: We already have the service in a variable earlier in this function. Make that const instead and drop this line. Also I would move the remove portion to the start of the function and return early if removal was requested to make it easier to read. Also that way any layer properties are not unnecessarily updated for any reason just to get the removal done.

kasparia: Check!

bundles/mapping/mapmodule/plugin/vectorlayer/VectorLayerPlugin.ol.js
```diff
@@ -572,6 +572,18 @@ Oskari.clazz.define(
...
+            if (options.remove) {
+                const request = Oskari.requestBuilder('RemoveMapLayerRequest')(layer.getId());
+                this._sandbox.request(this, request);
+
+                mapLayerService.removeLayer(layer);
+
+                this.removeMapLayerFromMap(layer);
+                delete this._oskariLayers[layer.getId()];
+
+                return;
```
ZakarFin, line 585: For consistency:
```suggestion
                return layer;
```

PR #1354, opened 2020-10-27
Layerlist raster layer filter

okauppinen:
Add layer filter for raster layers. Raster filter is like newest filter (registered in layerlist and defined in maplayerservice). Now maplayer service contains hard coded list of raster layer types.  Maybe maplayer plugins should push/register layer type/model with raster/vector/timeseries... additional information which could be used for filters (and layer tools). Vector layer filter uses hasFeatureData().


ZakarFin:
Looks like a good compromise for now. I think we can do it like this for now and make it more extendable through the plugins or layer type registration for maplayerservice in the future.

PR #1355, opened 2020-10-27
Layerlist fr and ru localization

okauppinen:
Add fr and ru localization for layerlist based on layerselector2 and selectedlayers2.


ZakarFin:
Not sure about including the english translations for missing ones manually as the build does it automatically. It makes it harder to detect missing translations programmatically (essentially they are not detected as missing now that they have values) but makes it easier to detect them for humans who are actually adding the texts if they use the JSON-format.

PR #1362, opened 2020-11-04
Myplaces form fixes

kasparia:


bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -42,14 +42,24 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
...
         this.nameRules = [
             {
                 required: true,
                 message: this.loc('validation.placeName')
             },
             () => ({
-                validator: (_, value) => Oskari.util.sanitize(value) === value ? Promise.resolve(value) : Promise.reject(new Error(this.loc('validation.placeNameIllegal')))
+                validator: (_, value) => {
+                    console.log(value);
+                    console.log(_);
```
ZakarFin, line 56: Remove console.log() lines

src/react/components/GenericForm.jsx
```diff
@@ -187,20 +214,26 @@ export class GenericForm extends React.Component {
                     </Space>
                 </Row>
                 );
+            case 'additionalinfo':
```
ZakarFin, line 217: Could 'info' be used instead of adding a new type?

src/react/components/GenericForm.jsx
```diff
@@ -13,7 +14,15 @@ const { TextArea } = Input;
 
 // If the form is shown on popup the Select dropdown opens behind popup without this
 // FIXME: this will probably not work with modal popups (dropdown might be over the modal overlay)
-const zIndexValue = 999999;
+const zIndexValue = 99998;
+
+const GlobalStyle = createGlobalStyle`
```
ZakarFin, line 19: This isn't the right place for this. Move it here: https://github.com/oskariorg/oskari-frontend/blob/master/src/react/components/Tooltip.jsx

ZakarFin: Though I'm not sure about this either. Does it create the global style just once or each time you use the `<Tooltip />`?

kasparia: Moved global styles to `portal.css` since Tooltip breaks down when adding child components.

src/react/components/GenericForm.jsx
kasparia: Changed this to `info` and removed some padding from Card-component.

PR #1364, opened 2020-11-05
Improve JSDoc and add Tooltip support for GenericForm

kasparia:


resources/css/portal.css
```diff
@@ -492,5 +492,5 @@ select:not([class*="ant-"]) {
 }
 
 .ant-tooltip {
-  z-index: 999999;
+  z-index: 999999 !important;
```
kasparia, line 495: I think this isn't good solution for this but couldn't figure any other workaround since AntD styles overwrite this later on with `z-index: 50001;` or something similar.

src/react/components/GenericForm.jsx
```diff
@@ -86,39 +86,65 @@ export class GenericForm extends React.Component {
...
-            );
+            if(field.showTooltip) {
+                return (
+                    <Tooltip title={ field.placeholder || field.label } trigger={ ['focus', 'hover'] }>
+                        { this._createFormItem( field ) }
+                    </Tooltip>
+                );
+            } else {
+                return this._createFormItem( field );
+            }
```
ZakarFin, line 108: Minor thing but doing it this way makes it clear the form item itself is always created the same way:

```suggestion
    const formItem = this._createFormItem(field);
    if (field.showTooltip) {
        return addTooltip(formItem, field.placeholder || field.label);
    }
    return formItem;
```

This could be added somewhere earlier in the file (doesn't need to be inside the react component):
```
const addTooltip = (formItem, text) => {
    if (!text) {
        return formItem;
    }
    return (<Tooltip title={ text } trigger={ ['focus', 'hover'] }>{ formItem }</Tooltip>);
}
```

ZakarFin: I would also change showTooltip from boolean to "tooltip" with a value of the tooltip text that we want to show. Then the caller can decide what to show as tooltip instead of it being placeholder or label.

kasparia: Check!

PR #1367, opened 2020-11-06
Confirm reset state

okauppinen:
Add confirm popup for reset map state buttons. Removed some dots from tools' tooltips (most simple tooltips doesn't have end dot).

Added createConfirmButtons() for Popup. I'm not sure if this is useful for other components and should it be added. At least "yes" and "no" localization is only in divmanazer's localization and PanButtons and default-buttons doesn't have to use other bundles localization or copy-paste them. Also it's shorter/simpler to create confirm popups for PanButtons and default-buttons.


ZakarFin:
Good idea with the confirm buttons 👍

PR #1372, opened 2020-11-11
Stateful bundle handling for links

ZakarFin:
Modernized code, added some unit-tests, added a new flag for `sandbox.generateMapLinkParameters()` `[statefulBundle].getStateParameters()` to generate an "optimized" link parameter. The intent is to allow bundles to drop things from the state parameters that does not affect the view. Like leaving out layers that have been set as full transparency (opacity = 0) or layers that are outside their scale limits. We use generateMapLinkParameters() to give the printing functionality info about layers that are on the map and we don't really need to send references to layers that are not shown to the enduser. Currently it only results in unnecessary requests made by the backend.

TODO: 

- [x] optimization for leaving out layers with scale limits/outside coverage area etc -> now checks with the map library if the layer is shown
- [ ] more testing

### DIFF
--- a/bundles/framework/mapfull/instance.js
+++ b/bundles/framework/mapfull/instance.js
@@ -551,7 +551,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapfull.MapFullBundleInstance',
                 const json = {
                     id: layer.getId(),
                     opacity: layer.getOpacity()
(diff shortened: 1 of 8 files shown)


ZakarFin:
This will be merged after the 2.1 release/for 2.2.

PR #1376, opened 2020-11-16
announcements bundle

osqq:
Announcements bundle made for creating announcements f.ex. about maplayer changes or service breaks. Normal user has access to valid announcements in their own tile, some of which can be shown as a pop-up if so needed (set when creating a new announcement). The announcements show title, content and date range when the announcement is valid. 

Admin-announcements bundle is for creating new announcements. Admin can create new, edit old and delete announcements. You can set title, content, date range and toggle the announcement to be shown as a pop-up for the users.

TODOS: 

- The bundle.md files contain some todos, few that might be currently needed and few that are for further development


api/framework/announcements/bundle.md
```diff
@@ -0,0 +1,31 @@
...
+<table class="table">
+  <tr>
+    <th> Dependency </th><th> Linked from </th><th> Purpose </th>
+  </tr>
+  <tr>
+    <td> [Oskari admin-announcements](/documentation/bundles/framework/admin-announcements) </td>
+    <td> Expects to be present in application setup </td>
+    <td> Admin-announcements bundle to create the announcements</td>
+  </tr>
+</table>
```
ZakarFin, line 31: I hope the admin isn't a dependency for the announcements. If you want to indicate that there is another bundle that relates to this maybe you can just tell that on the desciption with "Announcements have an admin UI implemented as bundle [bundle]"?

api/framework/admin-announcements/bundle.md
```diff
@@ -0,0 +1,54 @@
+# admin-announcements
```
ZakarFin, line 1: I think the admin bundles should all be under the admin "namespace" instead of framework. At least these new ones.

api/framework/admin-announcements/bundle.md
```diff
@@ -0,0 +1,54 @@
...
+
+Admin bundle for making announcements.
+
+## Description
+
+The bundle allows admins to create announcements to be displayed in announcements tab. Admins can optionally also set the announcement to be displayed in a pop-up. 
+
+## Parameters
```
ZakarFin, line 9: Are these the values that admins can set for an announcements? I first thought that these are parameters for some Oskari request or XHR request. Maybe add some small sentence what these are/change the title to make it more apparent?

bundles/framework/admin-announcements/Tile.js
```diff
@@ -0,0 +1,54 @@
+/*
```
ZakarFin, line 1: Doesn't really look like you need a custom tile. What does it do differently than the default tile?

bundles/framework/admin-announcements/Tile.js
```diff
@@ -0,0 +1,54 @@
...
+            const isInstanceId = !!((isContainer && this.instance.mediator.instanceId));
+
+            if (isInstanceId && !this.container.hasClass(this.instance.mediator.instanceId)) {
+                this.container.addClass(this.instance.mediator.instanceId);
+            }
+            if (isBundleId && !this.container.hasClass(this.instance.mediator.bundleId)) {
+                this.container.addClass(this.instance.mediator.bundleId);
+            }
+        }
+        _blink (element) {
```
ZakarFin, line 44: This function name starts with an underscore that usually means it's "private"/internal, but I don't see calls to it. What is it used for?

bundles/framework/admin-announcements/instance.js
```diff
@@ -0,0 +1,215 @@
...
+    }, {
+        /**
+         * @property {String[]} protocol
+         * @static
+         */
+        protocol: [
+            'Oskari.bundle.BundleInstance',
+            'Oskari.mapframework.module.Module',
+            'Oskari.userinterface.Extension'
+        ]
```
ZakarFin, line 213: Most of the code here is unnecessary boilerplate that could be removed if you just would extend the DefaultExtension instead of defining these protocols:

        extend: ['Oskari.userinterface.extension.DefaultExtension']

I just realized that this is true for the new layerlist bundle as well. But "this is the way" and all that.

bundles/framework/announcements/resources/locale/fi.js
```diff
@@ -0,0 +1,15 @@
+Oskari.registerLocalization(
+    {
+        "lang": "fi",
+        "key": "announcements",
+        "value": {
+            "title": "Ilmoitukset",
+            "desc": "",
+            "valid": "Voimassa:",
+            "dontShow" : "Älä näytä uudelleen",
+            "test":"TESTT",
```
ZakarFin, line 10: The test key is probably not used?

packages/framework/bundle/admin-announcements/bundle.js
```diff
@@ -0,0 +1,104 @@
...
+            "src": "../../../../bundles/framework/admin-announcements/instance.js"
+        }, {
+            "type": "text/javascript",
+            "src": "../../../../bundles/framework/admin-announcements/Tile.js"
+        }, {
+            "type": "text/javascript",
+            "src": "../../../../bundles/framework/admin-announcements/Flyout.js"
+        }, {
+            "type": "text/css",
+            "src": "../../../../bundles/framework/admin-announcements/resources/scss/style.scss"
```
ZakarFin, line 47: This file is empty. Remove it and reference to it.

packages/framework/bundle/announcements/bundle.js
```diff
@@ -0,0 +1,107 @@
...
+            "src": "../../../../bundles/framework/announcements/Tile.js"
+        }, {
+            "type": "text/javascript",
+            "src": "../../../../bundles/framework/announcements/service/AnnouncementsService.js"
+        }, {
+            "type": "text/javascript",
+            "src": "../../../../bundles/framework/announcements/Flyout.js"
+        },{
+            "type": "text/css",
+            "src": "../../../../bundles/framework/announcements/resources/scss/style.scss"
```
ZakarFin, line 50: This file is empty. Remove it and reference to it.

bundles/framework/admin-announcements/view/DeleteButton.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const DeleteButton = ({ controller, deleteAnnouncement }) => {
+    return (
+    <Confirm
+        title={<React.Fragment>
+            <div>
+                <Message messageKey='messages.confirmDeleteAnnouncement'/>
+            </div>
+            { layerCountInGroup > 0 &&
+                <DeleteLayersCheckbox checked={deleteLayers} onChange={ evt => controller.setDeleteLayers(evt.target.checked)}>
```
ZakarFin, line 16: I think this file can be just removed. It doesn't appear to be used anywhere. In addition it references announcements but also deleting layers? Some DeleteButton component could be added to src/react/components but the confirmation text and click callback should be given as params.

bundles/framework/announcements/Flyout.js
```diff
@@ -0,0 +1,106 @@
...
+            );
+            ReactDOM.render(content, this.container);
+        }
+    }, {
+
+        /**
+         * @property {String[]} protocol
+         * @static
+         */
+        protocol: ['Oskari.userinterface.Flyout']
```
ZakarFin, line 105: Extend the DefaultFlyout instead. This will remove the boilerplate from the file and you probably only need to implement render().

bundles/framework/announcements/instance.js
```diff
@@ -0,0 +1,216 @@
...
+
+            if (this.started) {
+                return;
+            }
+
+            this.started = true;
+            this.sandbox = sandbox;
+
+            var announcementService = Oskari.clazz.create('Oskari.framework.bundle.announcements.AnnouncementsService', sandbox, this.getLocalization().service);
+            this.sandbox.registerService(announcementService);
```
ZakarFin, line 86: So both the admin-announcements and the announcement bundle create an instance of the service and register it to sandbox. This creates a conflict in the code that probably causes bugs at some point if not now. You should probably check if the service is already registered in sandbox before creating and registering a new instance.

ZakarFin: Looking at the service I'm not sure if it needs to be registered on sandbox at all as it's just a wrapper for XHR-calls.

bundles/framework/announcements/instance.js
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * @method create called automatically on construction
+     * @static
+     */
+    function () {
+        this.sandbox = null;
+        this.started = false;
+        this.plugins = {};
+        this.localization = null;
+        this.seutumaisaSearchService = null;
```
ZakarFin, line 19: Looks like there's variables defined here that are not being used and/or refer to some specific service.

bundles/framework/announcements/Tile.js
```diff
@@ -0,0 +1,54 @@
+/*
+ * @class Oskari.mapframework.bundle.layerlist.Tile
+ *
+ * Renders the "announcements" tile.
+ */
+const BasicTile = Oskari.clazz.get('Oskari.userinterface.extension.DefaultTile');
```
ZakarFin, line 6: Is there a reason for implementing a custom tile instead of just using DefaultTile?

api/framework/admin-announcements/bundle.md
osqq: Yes. I will change the title a bit and add a small description.

bundles/framework/admin-announcements/view/DeleteButton.jsx
osqq: This was the unfinished file that was supposed to become a component. I will stash it on my own computer along with this comment for future development.

api/admin/admin-announcements/bundle.md
```diff
@@ -0,0 +1,56 @@
...
+  </tr>
+  <tr>
+    <td>Active</td><td>If toggled on, the announcement will show as a pop-up when users enter the application. The pop-up can be toggled off by the users after they have seen it. (Boolean)</td>
+  </tr>
+</table>
+
+## TODO
+
+* Make Admin-announcements bundle to be a separate tab on the same tile OR make the functionality be similar to admin-layereditor.
+* Finish DeleteButton.js
```
ZakarFin, line 37: The DeleteButton.js was removed so I think this TODO can be removed as well.

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,167 @@
...
+import moment from 'moment';
+
+/*
+This file contains the form for admin-announcements.
+This is the main file for creating and editing announcements.
+*/
+
+const titleError = "Lisää otsikko!";
+const contentError = "Lisää sisältö!";
+const dateError = "Lisää aikaväli!";
```
ZakarFin, line 16: These should come from localization.

bundles/framework/admin-announcements/instance.js
```diff
@@ -0,0 +1,26 @@
...
+     * @static
+     */
+    function () {
+        var conf = this.getConfiguration();
+        conf.name = 'admin-announcements';
+        conf.flyoutClazz = 'Oskari.framework.bundle.admin-announcements.Flyout';
+    }, {
+
+        afterStart: function () {
+        }
```
ZakarFin, line 21: Do you need to define afterStart() for DefaultExtension to work? I'm not sure so if you do it's ok. If it's not required for overriding something in the DefaultExtension we can remove the empty impl.

bundles/framework/admin-announcements/view/AnnouncementsFormMapperHandler.js
```diff
@@ -0,0 +1,146 @@
...
+            updated: false,
+            activeKey: []
+        });
+    }
+
+    deleteAnnouncement (index, id) {
+        const test = { id };
+
+        // TODO: Better/different way to confirm deleting an announcement
+        if (window.confirm(this.instance.getLocalization('deleteAnnouncementConfirm'))) {
```
ZakarFin, line 72: You can wrap the delete button on an AntD confirm component. Then you don't need localization variable in the handler at all.

bundles/framework/admin-announcements/view/AnnouncementsFormMapperHandler.js
```diff
@@ -0,0 +1,146 @@
...
+*/
+
+const getMessage = (key, args) => <Message messageKey={key} messageArgs={args} bundleKey='admin-announcements' />;
+
+class ViewHandler extends StateHandler {
+    constructor (instance) {
+        super();
+        this.instance = instance;
+        this.sandbox = instance.getSandbox();
+        this.locale = instance.getLocalization();
```
ZakarFin, line 17: this.locale isn't used in this file. Can it be removed?

bundles/framework/admin-announcements/view/AnnouncementsFormMapperHandler.js
```diff
@@ -0,0 +1,146 @@
...
+Handler for admin-announcements forms.
+*/
+
+const getMessage = (key, args) => <Message messageKey={key} messageArgs={args} bundleKey='admin-announcements' />;
+
+class ViewHandler extends StateHandler {
+    constructor (instance) {
+        super();
+        this.instance = instance;
+        this.sandbox = instance.getSandbox();
```
ZakarFin, line 16: this.sandbox isn't used in this file. Can it be removed?

bundles/framework/admin-announcements/view/AnnouncementsFormMapperHandler.js
```diff
@@ -0,0 +1,146 @@
...
+import { Message } from 'oskari-ui';
+
+/*
+Handler for admin-announcements forms.
+*/
+
+const getMessage = (key, args) => <Message messageKey={key} messageArgs={args} bundleKey='admin-announcements' />;
+
+class ViewHandler extends StateHandler {
+    constructor (instance) {
```
ZakarFin, line 13: We only need instance to get access to localization and sandbox here. You should use those as params instead of instance. But it also looks like you don't need those here at all or without changes to delete confirmation you need the localization at most. Change the parameters here accordingly. Pref refactor to use no params or just send localization or even the error message since that is all you need it for.

bundles/framework/admin-announcements/Tile.js
osqq: removed tiles and used extended classes in flyout and instance

bundles/framework/admin-announcements/Tile.js
osqq: removed tiles and used extended classes in flyout and instance

bundles/framework/admin-announcements/instance.js
osqq: extend now used in instances and flyouts

bundles/framework/announcements/instance.js
osqq: Service no longer created/used

bundles/framework/announcements/view/AnnouncementsModal.jsx
```diff
@@ -0,0 +1,53 @@
...
+const StyledCheckbox = styled(Checkbox)`
+    position: absolute !important;
+    left: 0px !important;
+    padding: 4px 15px !important;
+`;
+
+const AnnouncementsModal = ({ title, content, controller, index, visible, checked }) => {
+
+    const handleOk = (index, title, content, checked) => {
+        controller.handleOk(index, title, content, checked);
```
ZakarFin, line 20: A function named handleOk is very generic on the controller. Maybe rename it as something like setAnnouncementAsSeen() or something along those lines? Also I have a bit of a problem with index being used here. Don't the announcements have an id that could be used instead? Why does the handler need title and content as well?

api/admin/admin-announcements/bundle.md
osqq: removed

bundles/framework/admin-announcements/instance.js
osqq: Yes, it appears that you need to have afterStart() for it to work. At least my bundle did not f.ex. create the tile at all of this was not present.

bundles/framework/admin-announcements/view/AnnouncementsFormMapperHandler.js
osqq: removed

bundles/framework/admin-announcements/view/AnnouncementsFormMapperHandler.js
osqq: removed

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -0,0 +1,82 @@
...
+        this.updateState({
+            panels: panels,
+            modals: modals,
+            updated: true
+        });
+    }
+
+    handleOk (index, title, content, checked) {
+        if (checked) {
+            localStorage.setItem(title, content);
```
ZakarFin, line 52: This feels very fragile approach in a way that if you have an announcement with title "Welcome" there and some other functionality that would like to store something with key "Welcome" to the localstorage these will instantly conflict. Also the content is never used? Why not just use a boolean instead so it doesn't use space on the users device as much?

Refactor this so that instead of "title" as key you use something like "oskari-announcements" and as a value store an object with title and boolean as value. Or better yet if you have an id you can use that instead of the title so different language versions don't affect each other. Instead of an object as localstorage value you could use an array with "what ever is on the array has been shown" whether it's the title or id. 

Now if the admin changes the title of the announcement it is shown again. Or if the user changes the language I suppose the announcement title changes and is shown again. Have you considered using something like the announcement add/modify/save date to detect if the announcement should be shown? Then showing the announcement again could be triggered if the admin saves any modifications to the announcement (changing the content, not the title like it is now).

bundles/framework/announcements/view/AnnouncementsModal.jsx
osqq: Title and content used to create the localStorage item. Using f.ex. id caused problems because of its general ambiguity. I thought using title as the key is unique enough and also gets reset if an announcement's title is changed. This does still however leave the possibility of changing content of an announcement but users not seeing the pop-up, I will look into that. Index is used for splicing the announcement from the modals array. This seemed to be the easiest way to do it but can be changed to something better if suggested.

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
```diff
@@ -0,0 +1,69 @@
...
+          });
+      }
+      
+      
+  return (
+    <div>
+      {modals.map((modal, index) => {
+            const visible = controller.showModal(modal.title, modal.content);
+            return (
+              <AnnouncementsModal
```
ZakarFin, line 46: What is the purpose of writing a hidden modal? Couldn't it just be skipped?

bundles/framework/admin-announcements/resources/locale/en.js
```diff
@@ -0,0 +1,41 @@
...
+        "addNewForm": "New Announcement",
+        "delete": "Delete",
+        "save": "Save",
+        "cancel": "Cancel",
+        "newAnnouncement": {
+            "title": "Title",
+            "content": "Content",
+            "date-range": "Date range",
+            "show-popup": "Show pop-up",
+            "save": "Save"
```
ZakarFin, line 23: "Save" is localized twice here. Could both functionalities requiring it use the same one?

bundles/framework/admin-announcements/view/AnnouncementsCalls.js
```diff
@@ -0,0 +1,87 @@
+export const announcementsCalls = () => {
```
ZakarFin, line 1: Maybe call this file/object helper or service instead?

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,172 @@
+import React from "react";
+import "antd/dist/antd.css";
```
ZakarFin, line 2: We have wrapped individual AntD components under own components under scr/react/components that is exposed as "oskari-ui". One of the benefits is that you don't have to link the whole css just to get a style for a single component. You should use that approach here as well.

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,172 @@
...
+    else {
+
+      const values = {
+        id: form.id,
+        title: fieldsValue["title"],
+        content: fieldsValue["content"],
+        begin_date: rangeValue[0].format('YYYY-MM-DD'), 
+        end_date: rangeValue[1].format('YYYY-MM-DD'),
+        active: fieldsValue["active"]
+      };
```
ZakarFin, line 59: Code is duplicated here. You can do a common part for the values outside the if/else and just add id on edit + the method calls inside if/else.

bundles/framework/admin-announcements/view/AnnouncementsFormMapper.jsx
```diff
@@ -0,0 +1,74 @@
...
+Maps all created announcements into a collapse.
+*/
+
+const { Panel } = Collapse;
+
+const StyledButton = styled(Button)`
+    margin-top: 10px;
+`;
+
+const AnnouncementsFormMapper = ({controller,  announcements, updated, title, activeKey }) => {
```
ZakarFin, line 20: Is this more like `AnnouncementList`? I don't understand the naming here. Can you give more info on why should this be called "FormMapper"?

bundles/framework/admin-announcements/view/AnnouncementsFormMapper.jsx
```diff
@@ -0,0 +1,74 @@
...
+      var announcements = [];
+
+        ann.data.forEach((announcement) => {
+          announcements.push(announcement);
+          
+        });
+        
+        controller.pushAnnouncements(announcements);
+      });
+  }
```
ZakarFin, line 34: What is being done here? Feels it doesn't belong in a render() method.

bundles/framework/announcements/view/AnnouncementsHandler.js
osqq: refactored

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
osqq: apparently it was not needed anyways

ZakarFin: Maybe it's imported with other parts of the app using these components. Might be best to import the individual styles to be sure.

bundles/framework/admin-announcements/view/AnnouncementsFormMapper.jsx
osqq: renamed

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
osqq: refactored

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
osqq: done!

bundles/framework/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,163 @@
+import React from 'react';
+import { StateHandler, controllerMixin, Messaging } from 'oskari-ui/util';
+import { announcementsHelper } from './AnnouncementsHelper';
```
ZakarFin, line 3: This could just be an object named AnnouncementsHelper instead of a function that returns the object.

bundles/framework/admin-announcements/view/AnnouncementsHelper.js
```diff
@@ -0,0 +1,87 @@
...
+            return;
+        }
+
+        jQuery.ajax({
+            type: 'POST',
+            dataType: 'json',
+            data: {
+                params: JSON.stringify(data)
+            },
+            url: Oskari.urls.getRoute('DeleteAnnouncement'),
```
ZakarFin, line 71: The add, update, delete and get should be in the same ActionHandler on the backend. Just use 'DELETE' instead of POST here. It will make the server side code much cleaner than having multiple controllers with very few lines each.

bundles/framework/admin-announcements/view/AnnouncementsHelper.js
```diff
@@ -0,0 +1,87 @@
+export const announcementsHelper = () => {
+    const getAdminAnnouncements = (handler) => {
+        if (typeof handler !== 'function') {
+            return;
+        }
+
+        jQuery.ajax({
+            type: 'GET',
+            dataType: 'json',
+            url: Oskari.urls.getRoute('GetAdminAnnouncements'),
```
ZakarFin, line 10: Consider just having a parameter if you want to get all the announcements for admin vs just the current ones for regular users instead of having separate action routes.

bundles/framework/admin-announcements/view/AnnouncementsHelper.js
```diff
@@ -0,0 +1,87 @@
...
+
+    const saveAnnouncement = (data, handler) => {
+        if (typeof handler !== 'function') {
+            return;
+        }
+        jQuery.ajax({
+            type: 'POST',
+            dataType: 'json',
+            data: {
+                params: JSON.stringify(data)
```
ZakarFin, line 28: This writes the "data" as a serialized JSON format to a parameter named "params"? Wouldn't it be easier and make more sense to just send the announcement as multiple parameters OR have the JSON as payload in the POST body instead of this?

bundles/framework/admin-announcements/view/AnnouncementsHelper.js
```diff
@@ -0,0 +1,87 @@
+export const announcementsHelper = () => {
```
ZakarFin, line 1: Consider making this an object instead of function since the function only returns an object.

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,157 @@
...
+                  <Button type="primary" htmlType="submit" >
+                    <Message messageKey={'save'}/>
+                  </Button>
+                </Form.Item>
+                <Form.Item>
+                  <Confirm
+                      title={<Message messageKey='messages.deleteAnnouncementConfirm'/>}
+                      onConfirm={() => controller.deleteAnnouncement(index, form.id, form.title)}
+                      okText={getMessage('yes')}
+                      cancelText={getMessage('cancel')}
```
ZakarFin, line 128: The getMessage() returns Message-tag. You use the Message-tag for title, why use the function wrapper here?

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,157 @@
...
+      type: "array",
+      required: true,
+      message: <Message messageKey='dateError' />
+    }
+  ]
+};
+
+const getMessage = (key, args) => <Message messageKey={key} messageArgs={args} bundleKey='admin-announcements' />;
+
+const dateFormat = 'YYYY-MM-DD';
```
ZakarFin, line 25: Rename in CAPS and use in the code?

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const onFinish  = fieldsValue => {
+    // Should format date value before submit.
+    const rangeValue = fieldsValue["range_picker"];
+
+    const values = {
+      title: fieldsValue["title"],
+      content: fieldsValue["content"],
+      begin_date: rangeValue[0].format('YYYY-MM-DD'), 
+      end_date: rangeValue[1].format('YYYY-MM-DD'),
```
ZakarFin, line 37: Use the constant you define for the format 10 rows earlier?

bundles/framework/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+    // Cancel creating a new announcement or editing one. Close all panels.
+    cancel (index, id) {
+        if (id !== undefined) {
+            this.updateState({
+                activeKey: []
+            });
+        } else {
+            const newList = [...this.state.announcements];
+            newList.splice(index, 1);
```
ZakarFin, line 114: Do you add a new announcement to the list before it's saved? Or why do you need to splice here?

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
```diff
@@ -0,0 +1,70 @@
...
+                          <p>{announcement.content}</p>
+                          <Divider />
+                          <b><Message messageKey={'valid'} /></b>
+                          <p>{announcement.begin_date} - {announcement.end_date}</p>
+                      </Panel>
+                );
+            });
+            controller.updatePanelsModals(panels,modals);
+          });
+      }
```
ZakarFin, line 38: What's happening here? What does "updated" mean? Instead of updated variable why not just pass the announcements and create the panels on render based on that? This seems weirdly complex way of triggering loading of announcements if that is what this is.

osqq: Could not figure out how to do it without creating an error. The render happened every time before the announcements were fetched so there was no content and it lead to an error. Suggestions on how to render only after the announcements are fetched?

bundles/framework/admin-announcements/view/AnnouncementsListHandler.js
osqq: Yes, addForm() pushes a new object into the announcements list when user creates a new announcement (aka panel)

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
osqq: refactored

bundles/framework/admin-announcements/view/AnnouncementsForm.jsx
osqq: moved to localization

bundles/admin/admin-announcements/view/AnnouncementsHelper.js
```diff
@@ -0,0 +1,83 @@
...
+    const deleteAnnouncement = (data, handler) => {
+        if (typeof handler !== 'function') {
+            return;
+        }
+
+        jQuery.ajax({
+            type: 'DELETE',
+            dataType: 'json',
+            data: JSON.stringify(data),
+            url: Oskari.urls.getRoute('Announcements'),
```
ZakarFin, line 67: This makes the id in data-object to be sent as parameter in the url so you can use the getRequiredParam() on the server to validate it:
```suggestion
            url: Oskari.urls.getRoute('Announcements', { id: data.id }),
```

You could also modify the caller to just sent the id as param and not the data object to make it easier to understand/more maintainable.

bundles/admin/admin-announcements/Flyout.js
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * @method create called automatically on construction
+     * @static
+     * @param {Oskari.admin.bundle.admin-announcements.Flyout} instance
+     *    reference to component that created the flyout
+     */
+    function (instance) {
+        this.instance = instance;
+        this.container = null;
+        this.AnnouncementsListHandler = new AnnouncementsListHandler(this.instance);
```
ZakarFin, line 22: The AnnouncementsListHandler contructor doesn't do anything with the parameter. Remove the instance parameter here to reduce complexity.
```suggestion
        this.AnnouncementsListHandler = new AnnouncementsListHandler();
```

bundles/admin/admin-announcements/Flyout.js
```diff
@@ -0,0 +1,68 @@
...
+         */
+        render: function () {
+            if (!this.container) {
+                return;
+            }
+            const content = (
+                <LocaleProvider value={{ bundleKey: this.instance.getName() }}>
+                    <AnnouncementsList
+                        {...this.AnnouncementsListHandler.getState()}
+                        controller={this.AnnouncementsListHandler.getController()}
```
ZakarFin, line 60:
```suggestion
                        {...this.announcementsListHandler.getState()}
                        controller={this.announcementsListHandler.getController()}
```

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+    // Save back to localStorage
+    localStorage.setItem(name, existing.toString());
+};
+
+class ViewHandler extends StateHandler {
+    constructor () {
+        super();
+        this.announcementsHelper = announcementsHelper();
+        this.newTitle = 'Uusi ilmoitus';
```
ZakarFin, line 36: Hard-coded localization here

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,161 @@
...
+        this.state = {
+            update: false,
+            announcements: [],
+            title: 'Uusi Ilmoitus',
```
ZakarFin, line 40: Hard-coded localization here.

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,161 @@
...
+
+    getAdminAnnouncements (callback) {
+        this.announcementsHelper.getAdminAnnouncements(function (err, data) {
+            if (err) {
+                Messaging.error(getMessage('messages.getAdminAnnouncementsFailed'));
+                return;
+            }
+            callback(data);
+        });
+    }
```
ZakarFin, line 56: If you just update the state with the results you can remove that hack from the component that triggers the loading on render and pushes it to state afterwards and probably get rid of that updated-variable in state with this as well: 
```suggestion
    fetchAdminAnnouncements () {
        this.announcementsHelper.getAdminAnnouncements(function (err, data) {
            if (err) {
                Messaging.error(getMessage('messages.getAdminAnnouncementsFailed'));
                return;
            }
            this.updateState({
                announcements: data
            });
        });
    }
```

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,161 @@
...
+        this.newTitle = 'Uusi ilmoitus';
+        this.state = {
+            update: false,
+            announcements: [],
+            title: 'Uusi Ilmoitus',
+            updated: false,
+            active: true,
+            activeKey: []
+
+        };
```
ZakarFin, line 45:
```suggestion
            this.fetchAdminAnnouncements()
        };
```

bundles/admin/admin-announcements/view/AnnouncementsList.jsx
```diff
@@ -0,0 +1,71 @@
...
+  if (!updated) {
+    controller.getAdminAnnouncements(function(ann){
+      var announcements = [];
+
+        ann.data.forEach((announcement) => {
+          announcements.push(announcement);
+        });
+        controller.pushAnnouncements(announcements);
+      });
+  }
```
ZakarFin, line 31: Check the suggestion about triggering the fetch when the handler is initialized. With that change you should be able to remove this (and the pushAnnouncements() from handler).

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -0,0 +1,101 @@
...
+            dataType: 'json',
+            url: Oskari.urls.getRoute('Announcements'),
+            success: function (pResp) {
+                callback(pResp);
+            },
+            error: function (jqXHR, textStatus) {
+                Messaging.error(getMessage('messages.getFailed'));
+            }
+        });
+    }
```
ZakarFin, line 37: Trigger loading on startup, not on the component at render time:
```suggestion
        };
        this.fetchAnnouncements();
    }

    fetchAnnouncements () {
        jQuery.ajax({
            type: 'GET',
            dataType: 'json',
            url: Oskari.urls.getRoute('Announcements'),
            success: (pResp) => {
                this.updateState({
                    announcements: pResp.data
                });
            },
            error: function (jqXHR, textStatus) {
                Messaging.error(getMessage('messages.getFailed'));
            }
        });
    }
```

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
```diff
@@ -0,0 +1,70 @@
...
+                          <p>{announcement.content}</p>
+                          <Divider />
+                          <b><Message messageKey={'valid'} /></b>
+                          <p>{announcement.begin_date} - {announcement.end_date}</p>
+                      </CollapsePanel>
+                );
+            });
+            controller.updatePanelsModals(panels,modals);
+          });
+      }
```
ZakarFin, line 40: See the suggestion on the handler for fetching announcements at start and not on render. Refactor so you don't have to push modals and panels to the state, but map them on the render from announcements.

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -0,0 +1,101 @@
...
+        this.updateState({
+            panels: panels,
+            modals: modals,
+            updated: true
+        });
+    }
+
+    setAnnouncementAsSeen (index, checked, id) {
+        if (checked) {
+            this.addToLocalStorageArray('oskari-announcements', id);
```
ZakarFin, line 49: Move the name of the localstorage variable to a constants file for announcements. Import it here and in the admin and use the name from that constant so the two bundles don't go out-of-sync if one is updated (making it more maintainable).

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
```diff
@@ -0,0 +1,70 @@
...
+      
+      
+  return (
+    <div>
+      {modals.map((modal, index) => {
```
ZakarFin, line 45:
```suggestion
      {announcements.filter(a => a.active).map((modal, index) => {
```

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
```diff
@@ -0,0 +1,70 @@
...
+        controller.getAnnouncements(function(ann){
+          console.log("get announcements collapse");
+          console.log(ann);
+
+        
+          var panels = [];
+          var modals = [];
+            ann.data.forEach((announcement) => {
+
+              if (announcement.active && controller.showModal(announcement.id)) {
```
ZakarFin, line 24: Toggle the active flag in correct state on the controller before we hit render here so you don't have to double-check with controller.showModal()

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
```diff
@@ -0,0 +1,70 @@
...
+              title={modal.title}
+              content={modal.content}
+              key={modal.id}
+              index={index}
+              checked={checked}
+              />
+            );
+          })}
+      <div>
+        <Collapse accordion>{panels}</Collapse>
```
ZakarFin, line 59: Loop the announcements here so you don't need to push the panels in the state and have all of the render inside here without checking what is {panels} from props:
```suggestion
        <Collapse accordion>
          { announcements.map((announcement) => {
              return {
              <CollapsePanel header={announcement.title} key={announcement.id}>
                  <h3>{announcement.title}</h3>
                  <p>{announcement.content}</p>
                  <Divider />
                  <b><Message messageKey={'valid'} /></b>
                  <p>{announcement.begin_date} - {announcement.end_date}</p>
              </CollapsePanel>
            })}
        </Collapse>
```

osqq: Will push these on Friday!

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,156 @@
...
+                this.updateState({
+                    activeKey: []
+                });
+                this.fetchAdminAnnouncements();
+            }
+        }.bind(this));
+    }
+
+    deleteAnnouncement (index, id) {
+        this.id = { id };
```
ZakarFin, line 91: Why do you save id as this.id here?

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
ZakarFin: Aren't modals just announcements with `active:true`? Is there some reason they should be in two different arrays instead of using the same one for both?

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -0,0 +1,105 @@
...
+                // if announcement is active, then show pop-up of the content
+                modals.push(announcement);
+            }
+        });
+        this.updateState({
+            modals: modals
+        });
+    }
+
+    setAnnouncementAsSeen (index, checked, id) {
```
ZakarFin, line 51: `checked` is "Don't show again"? Rename it that way to make this easier to maintain. Also index isn't really needed since you can find the announcement from the list based on the id.

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -0,0 +1,105 @@
...
+            });
+        }
+        const newList = [...this.state.modals];
+        newList.splice(index, 1);
+        this.updateState({
+            modals: newList
+        });
+    }
+
+    showModal (id) {
```
ZakarFin, line 65: I would expect a modal to appear on the screen when calling a function with name `showModal()`. Rename to something like `isAnnouncementMarkedWithDontShow()`. You could make it better by checking if the announcement has active: true. Then it would be obvious why some are shown and others are not instead of the logic spread between functions. Then this could be called something like `isAnnouncementShown()`.

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -0,0 +1,105 @@
...
+        // is the modal stored in the localstorage aka has it been set to not show again
+        if (announcements && announcements.includes(id)) {
+            return false;
+        }
+        else {
+            return true;
+        }
+    }
+
+    onCheckboxChange (checked) {
```
ZakarFin, line 76: So, what does this do? What does checking a checkbox here mean? Would benefit from some comment at least or preferrably renaming to something that makes sense in this context :)

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -0,0 +1,105 @@
...
+        // Add new data to localStorage Array
+        existing.push(value);
+
+        // Save back to localStorage
+        localStorage.setItem(name, existing.toString());
+    }
+}
+
+export const AnnouncementsHandler = controllerMixin(ViewHandler, [
+    'getAnnouncements', 'updatePanelsModals', 'setAnnouncementAsSeen', 'onCheckboxChange', 'showModal'
```
ZakarFin, line 104: Remove getAnnouncements.

ZakarFin: I think updatePanelsModals() shouldn't be here either but I'm not sure about that one.

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,156 @@
...
+
+    updateActiveKey (key) {
+        this.updateState({
+            activeKey: key
+        });
+    }
+}
+
+export const AnnouncementsListHandler = controllerMixin(ViewHandler, [
+    'pushAnnouncements', 'getAdminAnnouncements', 'addForm', 'deleteAnnouncement', 'updateTitle', 'toggleActive', 'saveAnnouncement', 'updateAnnouncement', 'cancel', 'updateActiveKey'
```
ZakarFin, line 155: Can you check this list to only include existing functions that are not getters.

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
ZakarFin: Also you could get the index based on the id. Having both means you could end up removing the wrong thing due to some js error. Also it doesn't really make sense to both splice the list locally and fetch a new list entirely. Maybe just do one or the other?

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,156 @@
...
+                    announcements: newList,
+                    activeKey: [],
+                    updated: false
+                });
+                this.fetchAdminAnnouncements();
+            }
+        }.bind(this));
+    }
+    // Cancel creating a new announcement or editing one. Close all panels.
+    cancel (index, id) {
```
ZakarFin, line 110: You could split this to cancelEditing() and cancelAdding() to make it more readable. That way you probably don't need either of the params since adding is always the last? and editing just removes the activeKey from state no matter the what the params are.

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,156 @@
...
+        newList.push({
+            title: this.newTitle
+        });
+        this.updateState({
+            announcements: newList,
+            title: this.newTitle
+        });
+    }
+
+    updateTitle (value) {
```
ZakarFin, line 136: Why is title updated on it's own? The first announcement to add will have the "newTitle" as initial value but the second one will have the first ones title as initial value?

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,156 @@
...
+            title: value
+        });
+    }
+    toggleActive () {
+        this.updateState({
+            active: !this.state.active
+        });
+    }
+
+    updateActiveKey (key) {
```
ZakarFin, line 147: This makes a collapse open? Rename it to something like `openCollapse(key)`?

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
```diff
@@ -0,0 +1,156 @@
...
+            title: this.newTitle
+        });
+    }
+
+    updateTitle (value) {
+        this.updateState({
+            title: value
+        });
+    }
+    toggleActive () {
```
ZakarFin, line 141: Is this similar to updateTitle? The announcement has an active flag that is inherited from the last announcement added to the next one? Consider having a tmpAnnouncementForAdding object in state that triggers the showing of the form and holds the current values. Then it would be obvious what these means and could be reseted between adding announcements by just removing the key from state.

bundles/admin/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,163 @@
...
+};
+
+AnnouncementsForm.propTypes = {
+  title: PropTypes.string,
+  key: PropTypes.number,
+  controller: PropTypes.instanceOf(Controller).isRequired
+};
+
+const contextWrap = LocaleConsumer(AnnouncementsForm);
+export { contextWrap as AnnouncementsForm };
```
ZakarFin, line 163: Should work without wrapping to LocaleConsumer as this only uses Message-tag that is a LocaleConsumer itself.
```suggestion
export AnnouncementsForm;
```

bundles/admin/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,163 @@
...
+                label={<Message messageKey='newAnnouncement.title' />}
+                rules={[
+                  {
+                    required: true,
+                    message: <Message messageKey='titleError' />,
+                    whitespace: true
+                  }
+                ]}
+                value={title}
+                onChange={ evt => controller.updateTitle(evt.target.value)}
```
ZakarFin, line 88: I'm not sure you need to do this onChange() with the antd form component. Maybe try if removing it will function the same? And if so just remove it from the controller.

bundles/admin/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,163 @@
...
+              </Form.Item>
+              <Form.Item
+                name="range_picker"
+                label={<Message messageKey='newAnnouncement.date-range' />}
+                {...rangeConfig}
+              >
+                <DateRange popupStyle={{zIndex: '999999'}} />
+              </Form.Item>
+              <Form.Item name="active" label={<Message messageKey='newAnnouncement.show-popup' />} valuePropName="checked">
+                <Switch onClick={() => controller.toggleActive()}/>
```
ZakarFin, line 113: Same here. With antd form you can just handle the onFinish() and not call controller from fields.

bundles/admin/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -0,0 +1,163 @@
...
+                      onConfirm={() => controller.deleteAnnouncement(index, form.id, form.title)}
+                      okText={<Message messageKey='yes'/>}
+                      cancelText={<Message messageKey='cancel'/>}
+                      placement='top'
+                      popupStyle={{zIndex: '999999'}}
+                  >
+                      <Button
+                    key={key}>
+                          <Message messageKey='delete'/>
+                      </Button>
```
ZakarFin, line 133:
```suggestion
                      <Button key={key}>
                          <Message messageKey='delete'/>
                      </Button>
```

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
ZakarFin: Actually you can probably remove this function and just use the antd forms onFinish() like you already do now.

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
ZakarFin: Same as title again you might be able to remove this and just use the antd form onFinish()?

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
osqq: Modals array is an array of announcements that have been set active (to show as pop-up) AND have not been checked as 'don't show' by user ie. they are not in the localStorage array.

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
osqq: Have to check showModal to see is the announcement added to localStorage ie. user checked it as 'don't show', or am I understanding this wrong?

bundles/framework/announcements/view/AnnouncementsCollapse.jsx
osqq: This is outdated but to clarify, modals array is array of announcements that have been clicked "active" and have not been added to localStorage array.

bundles/admin/admin-announcements/view/AnnouncementsListHandler.js
osqq: This was supposed to update the title as title is written but it doesn't seem work and also it's not necessary so I will just remove it.

bundles/framework/announcements/view/AnnouncementsModal.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+        <Modal
+              mask={false}
+              centered
+              title={title}
+              visible={true}
+              onCancel={() => handleOk(index)}
+              footer={[
+                <StyledCheckbox key="checkbox" onChange={onCheckboxChange} ><Message messageKey={'dontShow'}/></StyledCheckbox>,
```
ZakarFin, line 34: This onCheckboxChange still feels weird to me since the checked value is not used to toggle the checkbox. But if it works I guess its ok. I have a feeling we might change this at some point anyway so merging for now.

ZakarFin:
There's a conflict with a couple of files. Can you check those? I removed most of the TODOs from the description since I feel they are no longer valid. The checkbox handling still seems a bit weird to me but I guess it's ok if it works.

PR #1380, opened 2020-11-17
OskariStyle like operator

okauppinen:
Changed like operator to have asterisk (end, start, both). 
Fixed non-case sensitive in-operator.
Changed caseSensitivity to default to false.
Added tests for filterOptionalStyle.

### DIFF
--- a/bundles/mapping/mapmodule/oskariStyle/filter.js
+++ b/bundles/mapping/mapmodule/oskariStyle/filter.js
@@ -1,6 +1,6 @@
 import { FILTER } from './constants';
 
-const { GREATER, LESS, LESS_EQUAL, GREATER_EQUAL, NOT_IN, IN, EQUAL, NOT_LIKE, LIKE, REG_EXP } = FILTER;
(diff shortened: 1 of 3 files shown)


bundles/mapping/mapmodule/oskariStyle/filter.js
```diff
@@ -113,3 +107,20 @@ const filterProperty = (property, featValue) => {
...
+    let val;
+    if (endsWildCard && startsWildCard) {
+        val = like.substr(1, like.length - 2);
+    } else if (endsWildCard) {
+        val = like.substr(0, like.length - 2);
+    } else if (startsWildCard) {
+        val = like.substr(1);
+    } else {
+        Oskari.log('Oskari.mapping.mapmodule.oskariStyle').info('Optional style\'s filter like operator does not have wild card, use value operator instead');
+        return false;
```
ZakarFin, line 123: Should we still compare value instead of always returning false if there is no wildcard?
```suggestion
        return featValue == like;
```

ZakarFin: "Working as intended"

PR #1383, opened 2020-11-27
Add Preview -component

kasparia:


ZakarFin:
I'm not sure what the purpose of this component is other than wrapping an existing SVG to a viewport. Feels like it might be specific to the styleform (visualization form rewrite based on React) so we might want to move it to a subfolder called styleform or something along those lines. Let's see what the other plans around the styleform are before making that call.

ZakarFin:
A later iteration of preview component was merged in #1388 so closing this one.

PR #1384, opened 2020-12-01
Add admin time series fields to layer editor admin UI

mingfeng:
The timeseries fields are used for configuring the ui widget for WMS-T layers.

Available ui options are: player, range and none. When range ui is selected,
admin users can further choose a WFS metadata layer, the timeline attribute
name and the zoom level to toggle between WMS-T/WFS.

time-series-fields

### DIFF
--- a/bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
+++ b/bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
@@ -180,6 +180,36 @@ class UIHandler extends StateHandler {
         }
         this.updateState({ layer });
     }
+    setTimeSeriesUI (ui) {
+        const layer = { ...this.getState().layer };
+        const timeseries = { ...layer.options.timeseries, ui };
+        layer.options.timeseries = timeseries;
+        this.updateState({ layer });
+    }
+    setTimeSeriesMetadataLayer (layerId) {
+        const layer = { ...this.getState().layer };
+        const timeseries = { ...layer.options.timeseries };
+        const metadata = { ...timeseries.metadata, layer: layerId };
+        timeseries.metadata = metadata;
+        layer.options.timeseries = timeseries;
+        this.updateState({ layer });
+    }
+    setTimeSeriesMetadataAttribute (attribute) {
+        const layer = { ...this.getState().layer };
+        const timeseries = { ...layer.options.timeseries };
+        const metadata = { ...timeseries.metadata, attribute };
+        timeseries.metadata = metadata;
+        layer.options.timeseries = timeseries;
+        this.updateState({ layer });
+    }
+    setTimeSeriesMetadataToggleLevel (toggleLevel) {
+        const layer = { ...this.getState().layer };
+        const timeseries = { ...layer.options.timeseries };
+        const metadata = { ...timeseries.metadata, toggleLevel };
+        timeseries.metadata = metadata;
+        layer.options.timeseries = timeseries;
+        this.updateState({ layer });
+    }
     setRefreshRate (refreshRate) {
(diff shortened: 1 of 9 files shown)


bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/VisualizationTabPane.jsx
```diff
@@ -10,6 +10,8 @@ import { Scale } from './Scale';
 import { ClusteringDistance } from './ClusteringDistance';
 import { WfsRenderMode } from './WfsRenderMode';
 import { StyledColumn } from './styled';
+import { TimeSeries } from './TimeSeries';
+import { scale } from 'ol/size';
```
ZakarFin, line 14: ol/size/scale isn't used

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/VisualizationTabPane.jsx
```diff
@@ -28,6 +31,9 @@ export const VisualizationTabPane = ({ layer, capabilities, scales, propertyFiel
             { propertyFields.includes(OPACITY) &&
                 <Opacity layer={layer} controller={controller} />
             }
+            { propertyFields.includes(TIMES) &&
```
ZakarFin, line 34: This checks if the functionality _can_ be configured based on the layer type. I think we should also check if the layer we are editing has a timeseries before showing this. However if a layer loses it's timeseries the metadata can't be unlinked in such case.

Check where the timeseries times are stored on the layer. I think they should be on layer.times or layer.capabilities.times. If the layer has a timeseries (==has the times array) OR if the layer.options.timeseries object is present -> show the control. Otherwise it can be left out of the UI.

mingfeng: Makes sense. Thanks for detailed explanation.

bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
```diff
@@ -180,6 +180,36 @@ class UIHandler extends StateHandler {
         }
         this.updateState({ layer });
     }
+    setTimeSeriesUI (ui) {
```
ZakarFin, line 183: Add the options object for these to get the designed structure:
```
layer: {
  options: {
    timeseries: {
      ui: 'player' | 'range' | 'none',
      metadata: {
        layer: [id for metadata layer],
        attribute: "alkuvuosi",
        toggleLevel: 8
      }
    },
    ...,
  },
  ...
}
```

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/TimeSeries.jsx
```diff
@@ -0,0 +1,173 @@
...
+const TimeSeriesContainer = styled('div')`
+  border: 1px solid #d9d9d9;
+  padding: 10px;
+  margin-top: 5px;
+  margin-bottom: 10px;
+`
+
+export const TimeSeries = ({ scales, controller }) => {
+    const [timeSeriesUI, setTimeSeriesUI] = useState(TIME_SERIES_UI.PLAYER);
+    const [metadataLayer, setMetadataLayer] = useState('');
```
ZakarFin, line 126: You shouldn't need to use state here as these are passed through controller -> the component is re-rendered with new props because of that. Instead take the options.timeseries as additional prop-value and render based on that.

ZakarFin:
Looks good, thanks!

PR #1387, opened 2020-12-03
Layerlist fix stashed state

okauppinen:
LayerVewTabsHandler's state contains also selected layers and using stashed state with request + filter (analysis, publisher) caused that SelectedLayersHandler.layers !== Oskari.getSandbox().findAllSelectedMapLayers() if layer is added after stash (stashed when publisher/analysis is activated) and flyout is closed.

setActiveFilter changes tab and filter but I think that filter only may confuse user if isn't returned. Returning other states from stash may confuse user more and create weird bugs. Changed to stash only filter handler state. Another solution is to override useStashedState in LayerViewTabsHandler and set only tab from stashed state.


ZakarFin:
Fixes an issue where adding layers while in publisher didn't show up correctly in selected layers after existing the publisher functionality.

PR #1388, opened 2020-12-08
New visual form

kasparia:


src/react/components/GenericForm.jsx
```diff
@@ -79,10 +79,7 @@ const StyledButton = styled(Button)`
 export class GenericForm extends React.Component {
     constructor (props) {
         super(props);
-
-        this.state = {
-            disabledButtons: this.props.formSettings.disabledButtons
-        };
+        console.log(props);
```
ZakarFin, line 82: Remove console.log()

src/react/components/ColorPicker.jsx
```diff
@@ -0,0 +1,29 @@
...
+    constructor (props) {
+        super(props);
+    }
+
+    render () {
+        return (
+            <StyledColorPicker type='color' { ...this.props } />
+        );
+    }
+}
```
ZakarFin, line 29: Guess this could just be a functional component?

```suggestion
export const ColorPicker = (props) => {
        return (
            <StyledColorPicker type='color' { ...this.props } />
        );
}
```

src/react/components/SvgRadioButton.jsx
```diff
@@ -0,0 +1,48 @@
...
+ * @calssdesc <SvgRadioButton>
+ * @memberof module:oskari-ui
+ * @see {@link module:oskari-ui/util.LocaleProvider|LocaleProvider}
+ * @param {Object} props - { }
+ *
+ * @example <caption>Basic usage</caption>
+ * <SvgRadioButton props={{ ...exampleProps }}/>
+ */
+
+export class SvgRadioButton extends React.Component {
```
ZakarFin, line 30: Make this a function component instead.

src/react/components/SvgRadioButton.jsx
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Button, Select, Tooltip } from 'oskari-ui';
+import { Form, Card, Space, Input, Row, Radio } from 'antd';
+import styled from 'styled-components';
+
+const formLayout = {
+    labelCol: { span: 24 },
```
ZakarFin, line 8: What does 24 mean here? Should it be related to the content of the options shown?

src/react/components/SvgRadioButton.jsx
```diff
@@ -0,0 +1,48 @@
...
+    wrapperCol: { span: 24 }
+}
+
+const RadioIcon = styled(Radio.Button)`
+    height: 34px;
```
ZakarFin, line 13: Feels like this should be given as a prop for the component if this is meant to be a generic component. You could just move this to a folder named after StyleForm and give it more specific name based on that.

ZakarFin: Size probably assumes something about the content of the options as well?

src/react/components/SvgRadioButton.jsx
ZakarFin: Also naming could be improved. Since this gets options as props this is a selection between those options rather than a single radiobutton option?

src/react/components/styleform/AreaTab.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { SvgRadioButton } from 'oskari-ui';
+import { Form, Row } from 'antd';
+
+export const AreaTab = (props) => {
+    return (
+        <>
```
ZakarFin, line 7: Use React.Fragment instead to be consistent in the codebase.

src/react/components/styleform/AreaTab.jsx
```diff
@@ -0,0 +1,30 @@
...
+            <Row>
+                <Form.Item name='lineStyle' label='Line dash' { ...props.formLayout }>
```
ZakarFin, line 9: Why does this have name and the Fill style one doesn't? Does it matter/affect something?

src/react/components/styleform/LineTab.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { SvgRadioButton } from 'oskari-ui';
+import { Form, Row } from 'antd';
+
+export const LineTab = (props) => {
+    return (
+        <>
```
ZakarFin, line 7: Use React.Fragment instead for consistency.

src/react/components/styleform/LineTab.jsx
```diff
@@ -0,0 +1,37 @@
...
+            <Row>
+                <Form.Item name='lineStyle' label='Line dash' { ...props.formLayout }>
```
ZakarFin, line 9: What does name mean here? It's repeated for all Form.Items with the same value. Is it needed/can it be removed?

src/react/components/SvgRadioButton.jsx
kasparia: AntD divided all components to 24 sections so `span: 24` here means that label column of this component is full width.

https://ant.design/components/grid/

src/react/components/styleform/LineTab.jsx
```diff
@@ -0,0 +1,37 @@
...
+                    <SvgRadioButton
+                        options={ props.lineIcons.lineDash }
```
ZakarFin, line 11: We can bake in the options for all of these in this component. No need to pass them from the parent.

src/react/components/styleform/LineTab.jsx
```diff
@@ -0,0 +1,37 @@
...
+                        onChange={ (event) => props.onChangeCallback('stroke.lineDash', event.target.value) }
+                        defaultValue={ props.styleSettings.stroke.lineDash }
```
ZakarFin, line 13: These need a bunch of checks to make sure we have the whole structure available.

ZakarFin: Also you might be able to just pass the line-settings from the whole style JSON to this component to make it easier for everyone to see what's happening.

src/react/components/styleform/PointTab.jsx
```diff
@@ -0,0 +1,20 @@
...
+import { Form, Row } from 'antd';
+
+export const PointTab = (props) => {
+    return (
+        <>
+            <Row>
+                <Form.Item
+                    { ...props.formLayout }
+                    name='image'
+                    label='Ikoni'
```
ZakarFin, line 12: It's work in progress and all but labels should come from the localization/Message component.

src/react/components/styleform/PointTab.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { SvgRadioButton } from 'oskari-ui';
+import { Form, Row } from 'antd';
+
+export const PointTab = (props) => {
+    return (
+        <>
```
ZakarFin, line 7: React.Fragment here as well.

src/react/components/styleform/PointTab.jsx
```diff
@@ -0,0 +1,20 @@
...
+            <Row>
+                <Form.Item
+                    { ...props.formLayout }
+                    name='image'
+                    label='Ikoni'
+                    onChange={ (event) => props.onChangeCallback('image.shape', event.target.value) }
+                >
+                    <SvgRadioButton options={ props.markers } defaultValue={ props.styleSettings.image.shape } />
```
ZakarFin, line 15: Here we do want to pass the markers as props so don't change this one (line styles and area fill styles can be baked in the form component).

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+import { Button, Select, ColorPicker, SvgRadioButton, Preview } from 'oskari-ui';
+import { Form, Card, Space, Input, Row, Radio, InputNumber } from 'antd';
+import styled from 'styled-components';
+
+import { LineTab } from './styleform/LineTab';
+import { AreaTab } from './styleform/AreaTab';
+import { PointTab } from './styleform/PointTab';
+
+const formLayout = {
+    labelCol: { span: 24 },
```
ZakarFin, line 12: What does 24 mean here?

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+    wrapperCol: { span: 24 }
+}
+
+const lineIcons = {
```
ZakarFin, line 16: This file seems to do a bit too much but if you move these to the forms that actually need them it's going to reduce the amount of code here.

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+
+const StaticForm = styled(Form)`
+    width: 400px;
+
+    & > .ant-space {
+        width: 100%;
+    }
+`;
+
+const sizeFormatter = (number) => Math.abs(number);
```
ZakarFin, line 96: This should not be in the main form component but components that handle size changes.

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+        case 'line':
+            return (<svg viewBox="0 0 80 80" width="80" height="80" xmlns="http://www.w3.org/2000/svg"><path fill="none" stroke={ strokeColor } d="M10,15L20,35L40,25" strokeWidth={strokeWidth} strokeLinejoin="miter" strokeLinecap="butt" strokeDasharray="0"></path></svg>);
+        case 'area':
+            return(<svg viewBox="0 0 80 80" width="80" height="80" xmlns="http://www.w3.org/2000/svg"><defs><pattern id="checker" x="0" y="0" width="20" height="20" patternUnits="userSpaceOnUse"><rect fill="#eee" x="0" width="10" height="10" y="0"><rect fill="#eee" x="10" width="10" height="10" y="10"></rect></rect></pattern></defs><rect x="0" y="0" width="80" height="80" fill="url(#checker)"></rect><path fill={ fillColor } stroke={ strokeColor } d="M10,17L40,12L29,40Z" stroke-width={ strokeWidth } stroke-linejoin="miter" stroke-linecap="butt" stroke-dasharray="0"></path></svg>);
+        default:
+            return;
+    }
+}
```
ZakarFin, line 116: This probably belongs in that preview component?

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+ * @calssdesc <StyleForm>
+ * @memberof module:oskari-ui
+ * @see {@link module:oskari-ui/util.LocaleProvider|LocaleProvider}
+ * @param {Object} props - { }
+ *
+ * @example <caption>Basic usage</caption>
+ * <StyleForm props={{ ...exampleProps }}/>
+ */
+
+export class StyleForm extends React.Component {
```
ZakarFin, line 129: Does this need to be a React.Component class or could it be a function component?

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+    constructor (props) {
+        super(props);
+
+        this.ref = React.createRef();
```
ZakarFin, line 133: Is ref really required here? I'd try to find a way to just use props.

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+        );
+    }
+
+    /**
+     * @method _getColorPickers
+     * @description Compose fill and stroke color picker elements for style form
+     * 
+     * @returns {React.Component} Only stroke color picker or both stroke and fill 
+     */
+    _getColorPickers () {
```
ZakarFin, line 259: Each form can init their own color pickers so it shouldn't be done on the main form level

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+                return false;
+        }
+    }
+
+    /**
+     * @method _createPreview
+     * @description Compose Preview -component
+     * @returns {React.Component} Preview component with provided style
+     */
+    _createPreview () {
```
ZakarFin, line 241: Each form can init their own preview so it shouldn't be done on the main form level

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+                            defaultValue={ this.state.fill.color }
+                        />
+                    </Form.Item>
+                    : false
+                }
+            </Row>
+        );
+    }
+
+    _getSizeControl () {
```
ZakarFin, line 281: Each form can init their own size controls so it shouldn't be done on the main form level

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+                            this.updateState('stroke.width', value);
+                            this.updateState('stroke.area.width', value);
+                            this.updateState('image.size', value);
+                        } } />
+                </Form.Item>
+            </Row>
+        );
+    }
+
+    _buildStyleList (options) {
```
ZakarFin, line 300: I'm not sure what is this used for?

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+                return this.parseThru(targetObject, targetString.split('.'), value); // first cycle of recursion converts targetString into array
+            } else if (targetString.length == 1 && value !== undefined) {
+                return targetObject[targetString[0]] = value; // We reach end of the recursion
+            } else if (targetString.length === 0) {
+                return targetObject; // target is already on level 0 so no need for recursion
+            } else {
+                return this.parseThru(targetObject[targetString[0]], targetString.slice(1), value); // recursive call and remove first element from array
+            }
+        }
+    }
```
ZakarFin, line 199: The tab selection can probably be handled as a state of the form component but all of this other stuff should be moved to a controller/uihandler type file. Similar to the one in admin-layereditor for example: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js

src/react/components/StyleForm.jsx
```diff
@@ -0,0 +1,348 @@
...
+                        </Form.Item>
+
+                        <Card>
+                            { this._getColorPickers() }
+
+                            { this._getCurrentTab( this.state.format ) }
+
+                            { this._getSizeControl() }
+
+                            { this._createPreview() }
```
ZakarFin, line 341: Color pickers, size controls and previews should all be in the "tab" components. The current tab can then be shown somewhat like in the layer adding wizard instead of calling a method. It's easier to see what's inside the `<Card>` that way: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/admin/admin-layereditor/view/LayerWizard/LayerWizard.jsx

src/react/components/StyleForm.stories.js
```diff
@@ -0,0 +1,158 @@
...
+        format: 'area',
+    }
+];
+
+
+storiesOf('StyleForm', module)
+    .add('Default', () => (
+        <StyleForm
+            icons={ testOptions }
+            styleList={ styleList }
```
ZakarFin, line 153: I don't see style list being a prop here. I think it should be handled outside the form with another component. Selecting an existing style from that list would then populate the props for StyleForm based on that selected style.

src/react/components/StyleForm.stories.js
```diff
@@ -0,0 +1,158 @@
...
+        value: 'storybooktyyli',
+        format: 'area',
+    }
+];
+
+
+storiesOf('StyleForm', module)
+    .add('Default', () => (
+        <StyleForm
+            icons={ testOptions }
```
ZakarFin, line 152: Not sure what these icons do or why they need to be passed. I think we only need to pass:
- style json
- markers (so the user can select which point symbol to use for points)
- a controller that will have setters for the individual controls on the form to call OR an onChange callback that will get the updated style when the user changes anything on the form (that should trigger a new render on the storybook that uses that updated style as prop, initializing a new version of the form)

src/react/components/styleform/LineTab.jsx
kasparia: Name of Form.Item is used for populating values into form.

kasparia: Fixed them now.

src/react/components/styleform/AreaTab.jsx
kasparia: Fix'd. Name is used for populating.

src/react/components/StyleForm.jsx
kasparia: Moved.

src/react/components/StyleForm.jsx
kasparia: Changed to function component.

src/react/components/StyleForm.jsx
kasparia: Changed so.

src/react/components/StyleForm.stories.js
kasparia: Fixed.

src/react/components/StyleForm.stories.js
kasparia: Moved style selector to its own component.

src/react/components/StyleForm.jsx
kasparia: Changed form into function component and now referencing to Form is done with: `let [form] = Form.useForm();`

This reference is needed to set field values into form.

src/react/components/StyleForm.jsx
kasparia: Fixed.

src/react/components/styleform/PointTab.jsx
kasparia: Added localisations to all components with `Message`.

src/react/components/StyleForm.jsx
kasparia: FormHandler added to `src/react/util/extras/StyleFormHandler.js`. Should this be somewhere else?

bundles/framework/oskariui/resources/locale/en.js
```diff
@@ -2,6 +2,117 @@ Oskari.registerLocalization({
...
+            },
+            "tooltips": {
+                "dot": "Edit point feature style",
+                "area": "Edit area feature style",
+                "line": "Edit line feature style"
+            },
+            "validation": {
+                "title": "The given data are invalid:",
+                "lineSize": "The line width is invalid. It must be between 1 and 50.",
+                "borderWidth": "The line width is invalid. It must be between 1 and 50."
```
ZakarFin, line 20: Since all the fields now seem to have an object with key label, would it make sense to have that fields validation message beside that label instead of this common validation key?

src/react/components/styleform/AreaTab.jsx
```diff
@@ -0,0 +1,101 @@
...
+];
+
+export const AreaTab = (props) => {
+    return (
+        <React.Fragment>
+            <Row>
+                <Form.Item
+                    { ...props.formLayout }
+                    name='stroke.color'
+                    label={ <Message bundleKey={ props.locSettings.localeKey } messageKey='VisualizationForm.area.linecolor.label' /> }
```
ZakarFin, line 36: Maybe replace VisualizationForm in localization key with StyleEditor so we know it's the new one?

ZakarFin: Also the linecolor name would better map to the style if it would be 'stroke.color' (based on the name of the field). This would simplify doing some generic validation stuff and make it easier to debug.

src/react/components/styleform/SizeControl.jsx
```diff
@@ -0,0 +1,42 @@
...
+import { Form, InputNumber } from 'antd';
+
+const sizeFormatter = (number) => Math.abs(number); 
+
+
+export const SizeControl = (props) => {
+    const locKey =
+        props.format === 'point' ? 'VisualizationForm.point.size.label'
+        : props.format === 'line' ? 'VisualizationForm.line.width.label'
+        : 'VisualizationForm.area.linewidth.label'
```
ZakarFin, line 13: Make SizeControl take the label as child component or prop so you don't have to tie the label to the style logic:
```
<SizeControl ...props>
   <Message bundleKey={ props.locSettings.localeKey } messageKey={ locKey }
</SizeControl>
```

bundles/framework/oskariui/resources/locale/en.js
kasparia: I'll add validation messages to localisation when validations are done.

PR #1391, opened 2020-12-16
Implement the logic to switch between different UI widgets for timeseries layers

mingfeng:
This PR does not implement the actual range UI widget for timeseries layers, but instead just implement the logic to switch between different UI widgets depending on the layer timeseries metadata.

switch-time-series-ui-2

### DIFF
--- a/bundles/framework/timeseries/instance.js
+++ b/bundles/framework/timeseries/instance.js
@@ -103,23 +103,52 @@ Oskari.clazz.define('Oskari.mapframework.bundle.timeseries.TimeseriesToolBundleI
             if (active) {
                 var conf = jQuery.extend(true, {}, this._controlPluginConf || {}, active.conf);
                 if (typeof conf.showControl === 'undefined' || conf.showControl) {
-                    this._createControlPlugin(active.delegate, conf);
-                    return;
+                    const controlClass = this._getControlPluginClazz(active.delegate);
+                    if (controlClass !== null) {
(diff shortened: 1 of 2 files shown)


bundles/framework/timeseries/instance.js
```diff
@@ -103,23 +103,52 @@ Oskari.clazz.define('Oskari.mapframework.bundle.timeseries.TimeseriesToolBundleI
             if (active) {
                 var conf = jQuery.extend(true, {}, this._controlPluginConf || {}, active.conf);
                 if (typeof conf.showControl === 'undefined' || conf.showControl) {
-                    this._createControlPlugin(active.delegate, conf);
+                    const controlClass = this._getControlPluginClazz(active.delegate);
+                    if (controlClass !== null) {
+                        this._createControlPlugin(controlClass, active.delegate, conf);
+                    }
                     return;
```
ZakarFin, line 110: If you move the return inside the if-statement that creates the plugin -> any possible existing ui will be removed if the new control wasn't created (`_removeControlPlugin()` is called at the end of the method). This should prevent some confusing side-effects where an existing UI might be used to control a timeseries layer that was configured with `ui="none"`
```suggestion
                    if (controlClass !== null) {
                        this._createControlPlugin(controlClass, active.delegate, conf);
                        return;
                    }
```

ZakarFin:
Thanks! For future reference we'll need to make some changes for the warning popup and the timeseries icon so they shouldn't be shown for timeseries layers with `ui=none` selected.

PR #1393, opened 2020-12-22
Add an initial version of timeseries range widget

mingfeng:
The widget gets the year range from the timeseries layer, but currently does not get the timeseries photo available years from metadata layer.

time-range-widget


bundles/framework/timeseries/view/TimeSeriesRange/TimeSeriesRangeControl.jsx
```diff
@@ -0,0 +1,66 @@
...
+    const [startValue, endValue] = value;
+    const marks = {
+        [start]: start,
+        [end]: end,
+    };
+    for (let i = start + 1; i < end; i++) {
+        if (i % 10 === 0) {
+            marks[i] = i;
+        }
+    }
```
ZakarFin, line 16: This (marks) could be an impl detail that is implemented in YearRangeSlider, but I'm really nitpicking here. This looks really nice for a first version.

PR #1396, opened 2020-12-23
Update timeseries data years from metadata layer for current map view

mingfeng:
data-years-from-wfs

### DIFF
--- a/bundles/framework/timeseries/view/TimeSeriesRangeControlPlugin.js
+++ b/bundles/framework/timeseries/view/TimeSeriesRangeControlPlugin.js
@@ -1,3 +1,4 @@
+import moment from 'moment';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { LocaleProvider } from 'oskari-ui/util';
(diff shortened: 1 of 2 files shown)


bundles/framework/timeseries/view/TimeSeriesRangeControlPlugin.js
```diff
@@ -68,6 +71,49 @@ class TimeSeriesRangeControlPlugin extends BasicMapModulePlugin {
...
+            srs: sandbox.getMap().getSrsName()
+        });
+        fetch(url, {
+            method: 'GET',
+            headers: {
+                'Accept': 'application/json'
+            }
+        }).then(response => {
+            if (!response.ok) {
+                this._loc('messages.errorFetchTimeSeriesMetadata');
```
ZakarFin, line 104: `this._loc()` returns a localized string so this doesn't really do anything.

bundles/framework/timeseries/view/TimeSeriesRangeControlPlugin.js
```diff
@@ -68,6 +71,49 @@ class TimeSeriesRangeControlPlugin extends BasicMapModulePlugin {
...
+            }
+            return response.json();
+        }).then(json => {
+            const yearSet = new Set();
+            const { features } = json;
+            features.forEach(feature => {
+                const year = moment(feature.properties.time).year();
```
ZakarFin, line 111: The attribute name to use for `time` can be configured by the admin as part of the metadata. `time` can be used as a default value but you should check the metadata what the admin has configured for the attribute name.

PR #1397, opened 2020-12-30
Implement new 3D layer settings

mr-lev1:
Implemented some new 3D layer settings to the Visualization tab of the layer editor according to the issue:
https://github.com/oskariorg/oskari-docs/issues/235


bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
```diff
@@ -536,6 +536,13 @@ class UIHandler extends StateHandler {
         }
         // Take a copy
         const layer = { ...this.getState().layer };
+
+        // If layer options contain default values, remove them.
+        layer.options.hasOwnProperty('dynamicScreenSpaceError') && layer.options.dynamicScreenSpaceError === false && delete layer.options['dynamicScreenSpaceError'];
+        layer.options.hasOwnProperty('dynamicScreenSpaceErrorDensity') && layer.options.dynamicScreenSpaceErrorDensity === 0.00278 && delete layer.options['dynamicScreenSpaceErrorDensity'];
+        layer.options.hasOwnProperty('dynamicScreenSpaceErrorFactor') && layer.options.dynamicScreenSpaceErrorFactor === 4.0 && delete layer.options['dynamicScreenSpaceErrorFactor'];
+        layer.options.hasOwnProperty('dynamicScreenSpaceErrorHeightFalloff') && layer.options.dynamicScreenSpaceErrorHeightFalloff === 0.25 && delete layer.options['dynamicScreenSpaceErrorHeightFalloff'];
```
ZakarFin, line 544: Move this logic to the components instead. If the selected value is the default -> remove the key from the options. Otherwise add it to options. That way the default value is in a single location and weird bugs don't happen if it ever changes.

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/DynamicScreensSpaceErrorOptions.jsx
```diff
@@ -0,0 +1,148 @@
...
+the dynamic screen space error,
+similar to fog density.`;
+
+const dynamicScreenSpaceErrorFactorTemplate = `A factor used to increase
+the computed dynamic
+screen space error.`;
+
+const dynamicScreenSpaceErrorHeightFalloffTemplate = `A ratio of the tileset's
+height at which the density
+starts to falloff.`;
```
ZakarFin, line 43: These are not localized but I'm not sure if they need to be. Perhaps it's best to keep it this way.

ZakarFin:
Looks good. If you can move the default value handling to the components so it's easier to maintain in the future this looks good for merging.

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/DynamicScreensSpaceErrorOptions.jsx
```diff
@@ -0,0 +1,157 @@
...
+    layer.options.hasOwnProperty('dynamicScreenSpaceErrorHeightFalloff') && layer.options.dynamicScreenSpaceErrorHeightFalloff === defaultDynamicScreenSpaceErrorHeightFalloff && delete layer.options['dynamicScreenSpaceErrorHeightFalloff'];
+
+    return <StyledFormField>
+        <StyledDynamicScreenSpaceErrorOptionsContainer>
+            <DynamicScreenSpaceError layer={layer} controller={controller} defaultValue={defaultDynamicScreenSpaceError}/>
+            <DynamicScreenSpaceErrorDensity layer={layer} controller={controller} defaultValue={defaultDynamicScreenSpaceErrorDensity}/>
+            <DynamicScreenSpaceErrorFactor layer={layer} controller={controller} defaultValue={defaultDynamicScreenSpaceErrorFactor}/>
+            <DynamicScreenSpaceErrorHeightFalloff layer={layer} controller={controller} defaultValue={defaultDynamicScreenSpaceErrorHeightFalloff}/>
+        </StyledDynamicScreenSpaceErrorOptionsContainer>
+    </StyledFormField>
```
ZakarFin, line 152: This approach only removes the value for rendering, not saving. How about something like this:
```suggestion
    const defaults = {
        dynamicScreenSpaceError: false,
        dynamicScreenSpaceErrorDensity: 0.00278,
        dynamicScreenSpaceErrorFactor: 4.0,
        dynamicScreenSpaceErrorHeightFalloff: 0.25
    };

    const getDefaultRemovingController = (key) => {
        return {
            setOptions = (opts) => {
                if (defaults[key] === opts[key]) {
                    // setting default value -> remove it
                    delete opts[key];
                }
                controller.setOptions(opts);
            }
        };
    };
    const getAttributes = (fieldName) => {
        return {
            layer,
            controller: getDefaultRemovingController(fieldName),
            defaultValue: defaults[fieldName]
        };
    }

    return <StyledFormField>
        <StyledDynamicScreenSpaceErrorOptionsContainer>
            <DynamicScreenSpaceError ...getAttributes('dynamicScreenSpaceError') />
            <DynamicScreenSpaceErrorDensity  ...getAttributes('dynamicScreenSpaceErrorDensity') />
            <DynamicScreenSpaceErrorFactor  ...getAttributes('dynamicScreenSpaceErrorFactor') />
            <DynamicScreenSpaceErrorHeightFalloff  ...getAttributes('dynamicScreenSpaceErrorHeightFalloff') />
        </StyledDynamicScreenSpaceErrorOptionsContainer>
    </StyledFormField>
```
Though I would refactor the child components to have a onChange() callback that could be used to call the setOptions() of the actual controller without needing to fake it, but this was easier to link as an example. Also the defaults could be moved out of the component to a constant variable in the file.

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/DynamicScreensSpaceErrorOptions.jsx
ZakarFin: Add a comment here that these descriptions have been copied from CesiumJS docs: https://cesium.com/docs/cesiumjs-ref-doc/Cesium3DTileset.html

ZakarFin:
I wonder if it would benefit the user to have some "reset to defaults" button on these or if the default value would be shown to the user for each field so it can be "manually reset". But that can be some future development. It's already good that you don't have to use SQL to update the values :)

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/DynamicScreensSpaceErrorOptions.jsx
tuomo-hautala: Just tested this again and seems to work and options are saved to database if they are not default values and save button is pressed.
setOptions() is called from each component when inputs are changed.
Layer options is compared to react component state on each render and removed if layer option matches default value.

https://github.com/oskariorg/oskari-frontend/pull/1397/files/7c1105ccbfd8397ffbea344179fd4241898c7a6f#diff-222746e53136bc07b49bd3b0ba18f6dc1c4c4078cedaad4e3d78d99f731a991bR46-R130

Or did I miss something?

ZakarFin: I think it makes more sense to _not_ set the value and remove the value when it it's changed to the default one _when it's being modified_  since we don't want to save it. Your solution always sets the value and then tries to correct the data on render assuming it is re-rendered and that the state is mutable. The component might not be rendered again due to logic change or js error. That's the difference on our approaches: the proposed one resets the value in the state when it's changed to a default value, your approach writes an unwanted state and removes the value when the component is rendered. Changing state while rendering will lead to non-trivial bugs later on.

In practice there's currently little difference to either one but if we ever move to immutable state this won't work. Also if we update the code to save the changes without the user clicking the save button the approach where data is changed while being rendered doesn't work and it won't be obvious why the saved layer isn't the same the user sees on the screen.

tuomo-hautala: Oh yes I get your point and your approach is probably more suitable.
Unnecessary re-renders are indeed not very good to have.
I will do the changes like you proposed.

ZakarFin:
Thanks!

mr-lev1:
Thank you very much for the support @ZakarFin and for the excellent work by @tuomo-hautala!

PR #1398, opened 2020-12-30
Switch between wms/wfs for wms timeseries layer

mingfeng:
This commit adds the functionality to toggle between wms and wfs layer
if a WFS metadata layer and toggle level is configured for timeseries
layer. The styles for the WFS features can be configured in WFS layer
style json under `timeseriesStyle`

wms-wfs-switch

With custom styles:

```
{
  "timeseriesStyle": {
    "stroke": {
      "color": "rgba(0, 0, 255, 1)",
      "width": 5
    },
    "fill": {
      "color": "rgba(255, 0, 0, 0.5)"
    }
  }
}
```

wms-wfs-switch-style

### DIFF
--- a/bundles/mapping/mapmodule/plugin/wmslayer/WmsLayerPlugin.ol.js
+++ b/bundles/mapping/mapmodule/plugin/wmslayer/WmsLayerPlugin.ol.js
@@ -21,6 +21,16 @@ Oskari.clazz.define(
      */
     function () {
         this._log = Oskari.log(this.getName());
+        this._timeSeriesMetadataLayer = null;
+        this._timeSeriesMetadataLayerStyleName = 'timeseriesStyle';
+        this._defaultMetadataLayerStyleDef = {
+            stroke: {
+                width: 1
+            },
+            fill: {
+                color: 'rgba(24, 219, 218, 0.5)'
+            }
+        };
     },
     {
         __name: 'WmsLayerPlugin',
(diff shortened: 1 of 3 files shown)


bundles/mapping/mapmodule/plugin/wmslayer/WmsLayerPlugin.ol.js
```diff
@@ -153,16 +174,85 @@ Oskari.clazz.define(
...
+                style,
+                maxZoom: toggleLevel
+            });
+            return layer;
+        },
+
+        _getMetadataLayerStyle: function (layerId) {
+            const service = Oskari.getSandbox().getService('Oskari.mapframework.service.MapLayerService');
+            const layer = service.findMapLayer(layerId);
+            const styles = layer.getOptions().styles;
```
okauppinen, line 235: Could use:
const layerStyleDef = layer.getStyleDef(this._timeSeriesMetadataLayerStyleName) || {};
const featureStyleDef = layerStyleDef.featureStyle || {};

With custom styles:
"timeseriesStyle": {
  "featureStyle" : {
     "stroke":...

https://oskari.org/documentation/examples/oskari-style#vector-layer-styling

bundles/mapping/mapmodule/plugin/wmslayer/WmsLayerPlugin.ol.js
```diff
@@ -153,16 +174,85 @@ Oskari.clazz.define(
...
+            const layerStyleDef = styles[this._timeSeriesMetadataLayerStyleName] || {};
+            const styleDef = {
+                ...this._defaultMetadataLayerStyleDef,
+                ...layerStyleDef
+            };
+            const { stroke, fill } = styleDef;
+            return new olStyle({
```
okauppinen, line 242: return this.getMapModule().getStyle(styleDef);

okauppinen:
I don't know if filtered features could be added to map by adding new olVectorLayer and should vector layer be handled by TimeSeriesRangeControlPlugin instead of WMSLayerPlugin.
"the timeseries bundle/functionality adds the features used to create the timeline highlights on the map. The timeline range selection is still used to limit which features are shown (only show features that have the attribute value in range of the selection)"

Alternative solution would go something like this:
As far as I understand TimeSeriesRangeControlPlugin fetches features AfterMapMoveEvent so maybe we can store range filtered features from here:
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/framework/timeseries/view/TimeSeriesRangeControlPlugin.js#L110
At first step we could filter feature array and when filtered OAPIF queries is implemented then we can add range filter to GetWFSFeatures request.
Then we can use filtered features for GFI popup, feature data table and adding features to map. 

When the user zooms out the WMS-layer (toggleLevel) then we can add olVectorLayer (via mapmodule?) or use VectorLayerRequest and AddFeaturesToMapRequest:
https://oskari.org/api/requests#unreleased/mapping/mapmodule/request/addfeaturestomaprequest.md
https://oskari.org/api/requests#unreleased/mapping/mapmodule/request/vectorlayerrequest.md
Requests supports geojson and oskari style definition and ol layer is handled by VectorLayerPlugin (featuerevent, hover,..) but I don't know if performance is better with olVectorLayer (at least if stored features can be used directly as source).

TimeSeriesRangeControlPlugin AfterMapMoveEvent listener could call toggleVectorLayer(event.getZoom()) which removes/adds/hides olVectorLayer/VectorLayer
or layer would be added with zoom limits olVectorLayer (minZoom/maxZoom), VectorLayerRequest (minZoomLevel/maxZoomLevel).
https://github.com/oskariorg/oskari-frontend/pull/1385

mingfeng:
Updates:

- Add WFS feature filtering with year range selector (move the code logic inside `TimeSeriesRangeControlPlugin.js`)
- update the WFS layer style creation as suggested

wms-wfs-switch-2

ZakarFin:
Removed some unused vars and refactored WFS-metadata handling to its own file + using the request we have for features: https://github.com/oskariorg/oskari-frontend/compare/develop...ZakarFin:mingfeng-feature/wms-wfs-switch?expand=1

Still needs some work but lets do it this way.

PR #1401, opened 2021-01-15
Fix padding on tabs and sort map layers alphabetically  

kasparia:


bundles/framework/myplaces3/CategoryHandler.js
```diff
@@ -65,7 +65,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.CategoryHandler',
         },
         getAllCategories: function () {
             return this.mapLayerService.getAllLayersByMetaType(this.metaType)
-                .map(layer => this._parseLayerToCategory(layer));
+                .map(layer => this._parseLayerToCategory(layer))
+                .sort((a, b) => a.name.localeCompare(b.name));
```
ZakarFin, line 69: Use the naturalSort() instead:
```suggestion
                .sort((a, b) =>  Oskari.util.naturalSort(a.name, b.name));
```
It handles numbers better without additional flags:
```
// by default, "2" > "10"
console.log("2".localeCompare("10")); // 1

// numeric using options:
console.log("2".localeCompare("10", undefined, {numeric: true})); // -1
```

ZakarFin: https://github.com/oskariorg/oskari-frontend/blob/master/src/util.js#L298-L305

PR #1403, opened 2021-01-19
Fix localisation keys and localisation file

kasparia:


bundles/framework/oskariui/resources/locale/en.js
```diff
@@ -1,116 +1,35 @@
...
+    "key": 'oskariui',
+    "value": {
+        "otherLanguages": 'Other languages',
+        "StyleEditor": {
             "subheaders" : {
-                styleFormat: 'Style format',
-                name: "Style name",
-                style: "Style"
+                "styleFormat": 'Style format',
+                "name": "Style name",
```
ZakarFin, line 9: There's both `' ` and `"` used here. Change these to `' ` since that's the ESLint rules we have currently. Also the keys don't need the `"` and ESLint will probably let you know about that. Other than that this looks good. You can include the fi & sv files as well with this PR.

PR #1408, opened 2021-01-26
Update map libs/resolve build issues

ZakarFin:
Updates map libs and solves oskariorg/oskari-docs#241


ZakarFin:
Build issue fixed in #1409. Will update libs in a different pr.

PR #1413, opened 2021-02-02
Layerlist hierarchical features

mr-lev1:
Frontend changes related to the hierarchical features of layerlist excluding the drag 'n drop ordering of groups. 
There is some code implemented related to the drag 'n drop ordering, but it still needs work.

Related issue: https://github.com/oskariorg/oskari-docs/issues/234


ZakarFin:
Any chance to revert the formatting changes to get a usable diff on this? 
image

bundles/admin/admin-layereditor/view/AdminLayerForm/GeneralTabPane/Groups.jsx
```diff
@@ -3,24 +3,54 @@ import PropTypes from 'prop-types';
...
     );
-    const renderItem = item => <ListItem>{item}</ListItem>;
     return (
         <Fragment>
             <Message messageKey='fields.groups' />
             <StyledFormField>
                 <Collapse>
                     <CollapsePanel header={<Message messageKey='selectMapLayerGroupsButton'/>}>
-                        <List dataSource={dataSource} renderItem={renderItem} />
+                        <List dataSource={dataSource} renderItem={item => item} />
```
ZakarFin, line 53: Guess renderItem isn't really needed if it doesn't do anything?

bundles/admin/admin-layereditor/view/AdminLayerForm/GeneralTabPane/Groups.jsx
```diff
@@ -3,24 +3,54 @@ import PropTypes from 'prop-types';
...
+        <div>
+            <StyledListItem>
+                <Checkbox key={group.id}
+                    onChange={evt => controller.setGroup(evt.target.checked, group)}
+                    checked={!!layer.groups.find(cur => cur === group.id)}
+                >
+                    {group.name[Oskari.getLang()]}
+                </Checkbox>
+            </StyledListItem> 
+            {group.groups.length > 0 && <SubGroupsListItem group={group} layer={layer} controller={controller}/>}
```
ZakarFin, line 44: Could this just use itself like SubGroupsListItem does? Then we didn't need the SubGroupsListItem?
```suggestion
            {group.groups.length > 0 && <Groups groups={group.groups} layer={layer} controller={controller}/>}
```

bundles/admin/admin-layereditor/view/LocalizingFlyout.jsx
```diff
@@ -9,7 +9,7 @@ const ExtraFlyout = Oskari.clazz.get('Oskari.userinterface.extension.ExtraFlyout
 
 // NOTE: Everything is in the single file for demonstrating purposes.
 export class LocalizingFlyout extends ExtraFlyout {
-    constructor (instance, title, options = {}, deleteMapLayersText) {
+    constructor(instance, title, options = {}, deleteMapLayersText) {
```
ZakarFin, line 12: This file has ONLY formatting changes? No added/modified/removed code?

bundles/framework/layerlist/resources/scss/style.scss
```diff
@@ -1,3 +1,6 @@
+$editIcon: url("../images/glyphicons-151-edit.png");
+$addSubThemeIcon: url("../images/glyphicons-191-plus-sign.png");
```
ZakarFin, line 2: Could you use AntD icons instead?

bundles/admin/admin-layereditor/instance.js
```diff
@@ -341,16 +364,24 @@ Oskari.clazz.defineES('Oskari.admin.admin-layereditor.instance',
             });
             this.themeFlyout.setSaveAction((value, id) => {
                 const httpMethod = id ? 'POST' : 'PUT';
-                const payload = id ? { locales: value, id: id } : { locales: value };
+                const getpayload = () => {
+                    if (id) {
+                        return parentId ? { locales: value, id: id, parentId: parentId } : { locales: value, id: id };
+                    } else {
+                        return { locales: value };
+                    }
```
ZakarFin, line 372: Saving a new group (group without id) will never send the parent id? So you can only add groups to root? But then you have another saveAction for subgroups? What does parent id mean here? Why do group and subgroup need different handling? Aren't they just groups that either have a parent or don't?

mr-lev1:
Thank you @ZakarFin for the comments! @osqq will check those out and fix what is necessary.

bundles/admin/admin-layereditor/view/LocalizingFlyout.jsx
osqq: Something happened with a developer's IDE that formatted everything weirdly but I think I got to revert them all in the lastest push.

bundles/admin/admin-layereditor/view/AdminLayerForm/GeneralTabPane/Groups.jsx
osqq: removed

bundles/admin/admin-layereditor/view/AdminLayerForm/GeneralTabPane/Groups.jsx
osqq: SubgroupsListItem has StyledDiv in it. That lets us indent the subgroups and also having the div in the Groups' "datasource" broke the list I think.

bundles/framework/layerlist/resources/scss/style.scss
osqq: These are what admin-layerlist already used, so either leave these in place or we need to change the icons for admin-layerlist too

bundles/admin/admin-layereditor/instance.js
swTuomo: To my understanding payload needs to have root id (id) and parent id (parentId) when saving "subsubgroups".
Parent id points to id at one step backwards on hierarchy.
This might be wrong and parentId here is unnecessary.

bundles/framework/layerlist/resources/scss/style.scss
ZakarFin: I don't understand this. What is this admin-layerlist? And why would you need to change them there if you use AntD icons here?

bundles/admin/admin-layereditor/instance.js
ZakarFin: Sure, looks like it needs parentId for update and insert for subgroups. I'm wondering why it's never sent for new groups (groups without id).

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -82,16 +85,82 @@ class ViewHandler extends StateHandler {
...
         const tools = Object.values(this.toolingService.getTools()).filter(tool => tool.getTypes().includes('layergroup'));
         const isUserAdmin = tools.length > 0;
         // For admin users all groups and all data providers are provided to groupLayers function to include possible empty groups to layerlist.
         // For non admin users empty arrays are provided and with this empty groups are not included to layerlist.
-        const allGroups = isUserAdmin ? this.mapLayerService.getAllLayerGroups() : [];
-        const allDataProviders = isUserAdmin ? this.mapLayerService.getDataProviders() : [];
-        let groups = groupLayers([...layers], this.groupingMethod, tools, allGroups, allDataProviders, this.loc.grouping.noGroup);
+        const allGroups = this.mapLayerService.getAllLayerGroups();
+        const allDataProviders = this.mapLayerService.getDataProviders();
+        let groups = isUserAdmin ? groupLayersAdmin([...layers], this.groupingMethod, tools, allGroups, allDataProviders, this.loc.grouping.noGroup) : groupLayers([...layers], this.groupingMethod, tools, allGroups, [], this.loc.grouping.noGroup);
```
ZakarFin, line 163: This is too complex one-liner to read. Simplify it. Also why dataproviders are sent as an empty array for non-admins?

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -82,16 +85,82 @@ class ViewHandler extends StateHandler {
...
+        if (index === -1) {
+            return;
+        }
+        const showWarn = [...this.state.showWarn];
+        showWarn.splice(index, 1);
+        this.updateState({ showWarn });
+    }
+
+    // deactivate all layers in group
+    removeGroup (group) {
```
ZakarFin, line 134: Rename to something like removeGroupLayersFromMap(group) or something descriptive like it.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -82,16 +85,82 @@ class ViewHandler extends StateHandler {
...
+        for (var i in group.layers) {
```
ZakarFin, line 135: Is group.layers an array or object?

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -82,16 +85,82 @@ class ViewHandler extends StateHandler {
...
+            const index = this.state.selectedLayerIds.indexOf(group.layers[i]._id);
+            if (index === -1) {
+                continue;
+            }
+            const selectedLayerIds = [...this.state.selectedLayerIds];
+            selectedLayerIds.splice(index, 1);
+            this.updateState({ selectedLayerIds });
+            setTimeout(() => this.sandbox.postRequestByName('RemoveMapLayerRequest', [group.layers[i]._id]), ANIMATION_TIMEOUT);
```
ZakarFin, line 143: Do you need to modify the state in addition to sending the request? The request should trigger AfterMapLayerRemoveEvent that should update the state.

ZakarFin: Maybe gather the layer ids that are removed and and sent them after single timeout instead of generating multiple timeouts in a loop?

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -82,16 +85,82 @@ class ViewHandler extends StateHandler {
...
+                return;
+            }
+            const showWarn = [...this.state.showWarn];
+            showWarn.splice(index, 1);
+            this.updateState({ showWarn });
+        }
+    }
+
+    // active all layers in selected group
+    addGroup (group) {
```
ZakarFin, line 115: Rename to addGroupLayersToMap(group) or something descriptive like it.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -82,16 +85,82 @@ class ViewHandler extends StateHandler {
...
+        if (this.state.showWarn.includes(group.id)) {
+            const index = this.state.showWarn.indexOf(group.id);
+            if (index === -1) {
+                return;
+            }
+            const showWarn = [...this.state.showWarn];
+            showWarn.splice(index, 1);
+            this.updateState({ showWarn });
+        }
+    }
```
ZakarFin, line 112: I don't see a reason why you should need to track the warning like this instead of just wrapping the checkbox with a confirm tag if there are too many layers that will be added to the map. This makes the API more difficult to understand than it needs to be and you need to do bookkeeping that you don't need to do.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/util.js
```diff
@@ -45,26 +45,66 @@ export const groupLayers = (layers, method, tools, allGroups = [], allDataProvid
         return typeof groupId === 'number' ? groupId : -Math.random();
     };
 
+    const mapGroupsAndLayers = function (group) {
+        // group has subgroups
+        if (group.hasSubgroups()) {
+            // check that group has layers
+            if (group.layers && group.layers.length > 0) {
```
ZakarFin, line 52: Both if statements are wrapped in the same ` if (group.layers && group.layers.length > 0) {` making already complicated method more nested than is should be.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/util.js
```diff
@@ -45,26 +45,66 @@ export const groupLayers = (layers, method, tools, allGroups = [], allDataProvid
...
+                        .forEach(layer => {
+                            if (group.layers[j].id == layer._id) {
+                                newGroup.addLayer(layer);
+                            }
+                        });
+                }
+                return newGroup;
+            }
+        }
+    };
```
ZakarFin, line 100: Actually most of the method is duplicated for some reason. The only thing that seems to be different in the ifs is recursion that's moved to the end of the method in this impl:
```suggestion
    const mapGroupsAndLayers = function (group) {
        if (!group.layers || group.layers.length === 0) {
            return;
        }
        const lang = Oskari.getLang();
        const name = group.getName();
        let newGroup = Oskari.clazz.create(
            'Oskari.mapframework.bundle.layerselector2.model.LayerGroup',
            group.id, method, name[lang]
        );
        newGroup.setTools(tools);
        // add layer(s) to group
        for (var i in group.layers) {
            layers.sort((a, b) => comparator(a, b, method))
                .filter(layer => !layer.getMetaType || layer.getMetaType() !== 'published')
                .forEach(layer => {
                    if (group.layers[i].id == layer._id) {
                        newGroup.addLayer(layer);
                    }
                });
        }
        if (!group.hasSubgroups()) {
            return newGroup;
        }
        // group has subgroups
        // check possible subgroups
        const mappedSubgroups = group.getGroups().map(subgroup => {
            return mapGroupsAndLayers(subgroup);
        });
        // filter out subgroups that don't have a layer nor a subgroup with a layer
        newGroup.setGroups(mappedSubgroups.filter(g => g !== undefined));
        return newGroup;
    };
```

ZakarFin: What is the layer object here though where you refer to its id with an underscore "private" variable `layer._id`? Is it an instance of an AbstractLayer? You should use getter instead.

bundles/mapping/mapmodule/domain/MaplayerGroup.js
```diff
@@ -102,12 +104,24 @@ Oskari.clazz.define('Oskari.mapframework.domain.MaplayerGroup',
...
         },
         setOrderNumber: function (orderNumber) {
             this.orderNumber = orderNumber;
         },
+        getRootId: function () {
+            return this.rootId;
+        },
+        setRootId: function (rootId) {
+            this.rootId = rootId;
+        },
```
ZakarFin, line 124: Looks like there is no code that calls set or get for rootId. Is there a reason for these?

bundles/framework/layerselector2/model/LayerGroup.class.js
```diff
@@ -1,12 +1,34 @@
...
         this.searchIndex = {};
         this.tools = [];
     }
+
+    getRootId () {
+        return this.rootId;
+    }
+    setRootId (rootId) {
+        this.rootId = rootId;
+    }
```
ZakarFin, line 19: Looks like there is no code that calls set or get for rootId. Is there a reason for these?

bundles/admin/admin-layereditor/instance.js
ZakarFin: Also there's another setSaveAction() that uses parent id differently so can they be combined?

ZakarFin: I feel like there's some hundred lines of copy pasted code here that isn't really needed and makes maintaining this stuff harder than it should be.

bundles/framework/layerlist/resources/scss/style.scss
osqq: *admin-layereditor, aka the admin functionalities of layerlist. These icons were already in use in there.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/util.js
```diff
@@ -76,19 +116,130 @@ export const groupLayers = (layers, method, tools, allGroups = [], allDataProvid
...
+            }
+        } else {
+            groupId = layerAdmin ? layerAdmin.organizationId : undefined;
+        }
+        // My map layers, my places, own analysis and 'orphan' groups don't have id so use negated random number
+        // as unique Id (with positive id group is interpret as editable and group tools are shown in layer list).
+        return typeof groupId === 'number' ? groupId : -Math.random();
+    };
+
+    const mapGroupsAndLayers = function (group) {
```
ZakarFin, line 173: Why is this a copy of mapGroupsAndLayers() from under groupLayers()? Why isn't the implementation shared?

bundles/framework/layerlist/resources/scss/style.scss
ZakarFin: Can you link where? I can't find any reference to png-files in admin-layereditor.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
osqq: Not completely sure how this request works, this was how it was before the changes minus the loop. Can we send an array of the removed "selectedLayerIds" with the request (so have the loop stay the same but gather ids in an array and move the request out of the loop) or how should this be done?

bundles/framework/layerlist/resources/scss/style.scss
osqq: Actually it was in admin-hierarchical-layerlist, but admin-leyereditor apparently uses the same icons anyways: https://demo.oskari.org/Oskari/dist/1.6.0-dev/geoportal/icons.png

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
osqq: The dataprovider array was that way originally, also the allGroups was sent as an empty array but that I needed to change. I'm not sure about the logic on the dataproviders.

bundles/framework/layerlist/resources/scss/style.scss
ZakarFin: Thats the generic sprite that is generated on build. I'm pretty sure it doesn't include pngs referenced by bundle scss-files but only the icons in oskari-frontend/resources/icons. But unless you can point to where they are used in the current admin (admin-layereditor) I'd like them replaced by an SVG icon from the AntD icons.

osqq: https://github.com/oskariorg/oskari-frontend/blob/9aa98238fbd49d4da2854ef7b71891420c69bd96/resources/icons.css#L137

https://github.com/oskariorg/oskari-frontend/blob/9aa98238fbd49d4da2854ef7b71891420c69bd96/bundles/admin/admin-layereditor/instance.js#L129

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
ZakarFin: You need to send multiple requests. I'm just thinking that they could probably send them in a single setTimeout() instead of generating multiple calls to setTimeout()

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
ZakarFin: Could the logic be changed so that the checkbox is _not_ wrapped inside the confirm tag if there's less than x layers that would be added?

bundles/framework/layerlist/resources/scss/style.scss
osqq: Apparently the 'editIcon' was different icon than I originally thought but hasn't worked, the plus sign has. The admin-layereditor icons for editing the layer and theme (group) is an icon from the generated sprite and this 'addSubThemeIcon' icon is also an icon that is present in the sprite , just not taken from there in this instance. So changing these would mean we need to change the icon for 'edit-layer' too to keep consistency (?)

https://github.com/oskariorg/oskari-frontend/blob/9aa98238fbd49d4da2854ef7b71891420c69bd96/resources/icons.css#L137-L143

https://github.com/oskariorg/oskari-frontend/blob/9aa98238fbd49d4da2854ef7b71891420c69bd96/bundles/admin/admin-layereditor/instance.js#L129

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapsePanel.jsx
```diff
@@ -30,31 +50,132 @@ const StyledEditGroup = styled.span`
...
+    return (
+        <StyledSubCollapse>
+            <StyledSubCollapsePanel {...propsNeededForPanel}
+                header={group.getTitle()}
+                showArrow
+                extra={
+                    <React.Fragment>
+                            <Confirm
+                                title={<Message messageKey='grouping.manyLayersWarn'/>}
+                                visible={warnActive}
```
ZakarFin, line 130: The API in https://ant.design/components/popconfirm/ doesn't mention visible. What does it do? It has a disabled prop that you could use to toggle showing/not showing the confirmation. Also instead of passing warnActive through the handler could you just use `disabled={layers.length < 10}` or something like that?

bundles/admin/admin-layereditor/instance.js
```diff
@@ -362,28 +364,26 @@ Oskari.clazz.defineES('Oskari.admin.admin-layereditor.instance',
...
-                const getpayload = () => {
-                    if (id) {
-                        return parentId ? { locales: value, id: id, parentId: parentId } : { locales: value, id: id };
-                    } else {
-                        return { locales: value };
-                    }
-                };
+                const httpMethod = newSubgroup ? 'PUT' : id ? 'POST' : 'PUT';
+
+                const payload = newSubgroup ? { locales: value, parentId: id } : id ? { locales: value, id: id, parentId: parentId } : { locales: value };
```
ZakarFin, line 369: Simplify these lines. It's easier to maintain if you don't try one-lining this.

bundles/admin/admin-layereditor/instance.js
```diff
@@ -345,15 +346,16 @@ Oskari.clazz.defineES('Oskari.admin.admin-layereditor.instance',
...
                     }
                 });
             };
 
-            const loc = id ? this.loc('editTheme') : this.loc('addTheme');
+            const loc = newSubgroup ? this.loc('addTheme') : id ? this.loc('editTheme') : this.loc('addTheme');
             this.themeFlyout = new LocalizingFlyout(this, loc, {
                 headerMessageKey: 'themeName',
-                id: id,
+                id: id || newSubgroup && id,
```
ZakarFin, line 358: How does this work? I read it as id is either id if it's truthy OR id(?) if newSubgroup is truthy resulting in id = id whatever the case. If it does something different than that you should add some parenthesis to make it easier to read.

swTuomo: When user will edit theme or add new sub theme, data needs to be fetched. This will basically just check if id needs to be set or not. When new root theme is added, id is not set at all. But yes this could indeed look better. I will fix it.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapsePanel.jsx
osqq: https://ant.design/components/popconfirm/ "Conditional trigger" example. I tried to do this in several other ways but it never worked properly with the switch.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
osqq: Not sure how to implement that, could you suggest one? Researching for loops with setTimeout I only encountered a solution that mentioned to switch var to let, I made that change.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
osqq: Object

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -825,38 +844,42 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
         /**
          * @method getAllLayerGroups
          * Returns an array of layer groups added to the service
          * @param {String|Integer} id if defined and not equal -1 return only wanted group
-         * @return {Oskari.mapframework.domain.AbstractLayer[]}
+         * @return {Oskari.clazz.define.getGroups[]}
          */
         getAllLayerGroups: function (id) {
             var layerGroups = null;
             if (id && id !== -1) {
```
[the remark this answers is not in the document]
osqq: Do this server side

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/Layer/Layer.jsx
```diff
@@ -55,7 +56,9 @@ const Layer = ({ model, even, selected, controller }) => {
                         .filter(tool => tool.getTypes().includes('layerList'))
                         .map((tool, i) =>
                             <Tooltip key={`${tool.getName()}_${i}`} title={tool.getTooltip()}>
-                                <div className={tool.getIconCls()} onClick={() => onToolClick(tool)}/>
+                                <div onClick={() => onToolClick(tool)}>
+                                    <EditFilled />
```
ZakarFin, line 60: Use tool.getIconComponent() here as well so we don't need to bundle the SVG if it's not used for non-admin users and it allows multiple tools with different icons to be added.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapsePanel.jsx
```diff
@@ -23,71 +43,203 @@ const StyledListItem = styled(ListItem)`
...
+    event.stopPropagation();
+};
+
+const onCancel = (event, setVisible) => {
+    setVisible(false);
+    event.stopPropagation();
+}
+
+
+const SubCollapsePanel = ({ active, group, selectedLayerIds, selectedGroupIds, controller, propsNeededForPanel }) => {
```
ZakarFin, line 101: I don't find where `selectedGroupIds` is used. Can it be removed? Also I don't see a reason for passing `active` here since selectedLayerIds and group that are used to calculate the value is passed here. So move the logic of calculating "active" here instead and remove it from props. What is in `propsNeededForPanel`? Is it needed or can it be removed as well?

bundles/framework/layerlist/resources/scss/style.scss
ZakarFin: Note that the binaries for icons are still here with the style.scss changes that shouldn't be required anymore with the SVG icons. Remove these.

ZakarFin:
I'm not sure what the status of this change set is currently. I added some comments but it feels like the requested changes have been half-made with at least cleanup missing. Can you let me know when I should review it again/when you think it's ready?

mr-lev1:
Thanks for your feedback, @ZakarFin !
When we talked about this pull request last week with our developers I got the expression that all requested fixes were done. If not, we will do the necessary fixes since we're aiming for the 2.2.0 version (hopefully it's still possible!).

@osqq and @swTuomo, can you make sure that everything is fixed according to the comments and let @ZakarFin know if he can review? Or ask if there is still something that needs discussing?

ZakarFin:
Thanks for letting me know! We're planning on doing 2.2.0 this week. I don't feel too comfortable merging this so near the release since we haven't been able to test it ourselves at all yet. However we will probably do 2.3.0 or 2.2.1 in the very near future (I'm thinking early April at this point) and I could see this being a part of that release if we can merge it right after the 2.2.0.

bundles/admin/admin-layereditor/instance.js
swTuomo: This has been fixed.

bundles/admin/admin-layereditor/instance.js
swTuomo: It has been a bit hard to overcome the fact that so much functionality has been packed to getThemeFlyout method so additional getSubthemeFlyout method was created (copy pasted and modified). For now this has been fixed and previously added code has been removed and only getThemeFlyout is used now with some additional props letting code know if user has pressed "Add theme", "Modify Theme" or "Add subtheme". Generally I feel like this would be so much simpler if all of these three methods would be separately done when this method was implemented at first place.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/Layer/Layer.jsx
osqq: Done

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapsePanel.jsx
osqq: Active is needed as it defines if toggle is on based on if all layers are selected. Others have now been cleaned up.

bundles/admin/admin-layereditor/instance.js
ZakarFin: Sure, though in my opinion adding is basically the same thing as modifying but you either have an id for modification or you don't hence you are adding. In the same sense subtheme doesn't really have to be any different than a "root" theme. A theme either has a parent id (making it a "subtheme") or it doesn't  (making it a "root" theme) and that can be also used to determine if you want to show a different title for the form or not. The theme/group details are loaded from the same endpoint whether its a root or subgroup and clicking save or delete are also handled by the same endpoint regardless. I'm not sure that having that separation adds any value for the end-user either but maybe it does.

This is an open source project so if you feel too much code is packed to a function you have the ability to make it better. That is also what I'm trying to do here. Instead of blindly accepting and merging PRs I try to guide the code towards a better future. Of course I'm not always right but it is my goal to make the code better and more maintainable since I'm one of the devs that actually needs to maintain it. If you have multiple places doing essentially the same thing it's very easy to miss some of those places when a modification is required. That leads to bugs that are not easily spotted. If code needs to be copy-pasted in large amounts it needs some comments on the code to document WHY that code was copy-pasted. Otherwise when you need to do something with that code again spotting that _one line difference_ in a 100 lines of a function that has been copy-pasted to four different functions is really painful. It usually leads to the function being copied again and again for fear of breaking something making it worse and worse for the future. I agree that it's much easier to keep adding code than making sure that a function does what it is supposed to do based with different parameters. However it's much easier to maintain code where you only need to handle a change on one spot than copy-pasting that same change in four more spots.

In this example only things that change between adding/editing, root/subtheme is that the HTTP-method changes and the id and parent id are either known and sent or unknown and as such can't be sent. Copy pasting some 100 lines of code to a parallel functions just for those changes feels wrong to me:

- If we have id -> use HTTP POST, otherwise use PUT
- If we have an id -> pass it to the backend
- If we have a parent id -> pass it to the backend

You could even argue that you don't need a control-flow for the last two and just send them to the backend as received in the payload and as there's no value the backend will handle accordingly whether you send them or not.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapse.jsx
```diff
@@ -21,13 +21,20 @@ const StyledLayerCollapsePanel = styled(LayerCollapsePanel)`
 `;
 
 const LayerCollapse = ({ groups, openGroupTitles, selectedLayerIds, controller }) => {
+
+    const [groupsData, setGroupsData] = useState([]);
+
+    useEffect(() => {
+        setGroupsData(groups);
+    },[groups]);
```
ZakarFin, line 29: I think the proper way to do this is with React.memo() and this feels very hackish :) Did you try if it makes a difference?

ZakarFin: I mean if it doesn't we can remove it. But if it does then I'm not sure if it's good for merging or if it should be refactored. I'll probably try removing it in local testing but I'm curious if you tested if this is something that is required for performance.

swTuomo: I did not try react memo yet. It would probably be more suited for this one indeed. useState and useEffect can be removed from here.

ZakarFin: https://github.com/mr-lev1/oskari-frontend/pull/10

mr-lev1: Changes merged!

bundles/framework/layerlist/view/LayerViewTabs/SelectedLayers/SelectedTab.jsx
```diff
@@ -21,11 +21,10 @@ const StyledBadge = styled.div`
     background: #ffd400;
     border-radius: 4px;
     text-align: center;
-    padding: 2px 10px;
```
ZakarFin, line 24: Is there a reason padding was removed? Now it looks like 
image

While with the padding it looks like 
image

swTuomo: Oh no there is no reason. This was my mistake. I will correct it. Thanks!

PR #1416, opened 2021-02-02
Admin layereditor legend urls

okauppinen:
Add RasterStyle which can be used to manage styles and legends. Show RasterStyle for layers that have styles in capabilities (wms, wmts). Moved LegendImage to RasterStyle. LegendImage is shown if there's no styles available. Not sure if it useful for wms or wmts layers but maybe other raster layers could use it.

Removed LayerComposingModel.STYLE from WMS and WMTS layers. Style.jsx is used to select layers default style and RasterStyle has default style checkbox.


ZakarFin:
Looks good but could RasterStyle be refactored as function-component with React.useState() for consistency since all other components are function-components? Also consider breaking RasterStyle.jsx to multiple files.

PR #1428, opened 2021-02-12
Initialize capabilities as empty object

kasparia:


ZakarFin:
This fixes an issue where some inputs on the admin rely on capabilities object being present (like timeseries controls).

PR #1432, opened 2021-02-18
Make preview component more clear

kasparia:


src/react/components/StyleEditor/Preview.jsx
```diff
@@ -24,109 +25,37 @@ const areaPreviewSVG = '<svg viewBox="0 0 80 80" width="80" height="80" xmlns="h
...
+    let size = 3;
+
+    let previewAttributes = {
+        strokeColor: PREVIEW_DEFAULTS.stroke.color,
+        strokeWidth: PREVIEW_DEFAULTS.stroke.width,
+        strokeLineCap: PREVIEW_DEFAULTS.stroke.lineCap,
+        fill: PREVIEW_DEFAULTS.defaultFill, // empty string here?
+        fillColor: PREVIEW_DEFAULTS.fill.color,
+        pattern: PREVIEW_DEFAULTS.fill.area.pattern
+    };
```
ZakarFin, line 58: Most of these helper functions and variables don't need to be inside the render function. Move them out.

src/react/components/StyleEditor/Preview.jsx
```diff
@@ -190,47 +115,126 @@ export class Preview extends React.Component {
...
+        path.setAttribute('stroke', previewAttributes.strokeColor);
+        path.setAttribute('stroke-width', previewAttributes.strokeWidth);
+        path.setAttribute('stroke-linecap', previewAttributes.strokeLineCap);
+        path.setAttribute('stroke-dasharray', previewAttributes.strokeDashArray);
+        path.setAttribute('stroke-linejoin', previewAttributes.strokeLineJoin);
+
+        size = PREVIEW_DEFAULTS.image.size;
+
+        return path.outerHTML;
+    }
```
ZakarFin, line 184: If you add oskariStyle as param to _composeLinePath() it doesn't need to be defined inside the render function. If it's not inside the render function it's easier to debug and test by itself helping with maintenance etc.

src/react/components/StyleEditor/constants.js
```diff
@@ -39,3 +39,38 @@ export const LINE_STYLES = {
         }
     ]
 };
+
+
+// Default values for every style settings value
+export const PREVIEW_DEFAULTS = {
```
ZakarFin, line 45: How does preview defaults differ from the default style that is used as template for StyleEditor? Do we really need these or could we use that default style here as well?

kasparia: Removed `PREVIEW_DEFAULTS` and replaced with `OSKARI_BLANK_STYLE`

src/react/components/StyleEditor/Preview.jsx
kasparia: Moved.

src/react/components/StyleEditor/Preview.jsx
kasparia: Moved along all the other helper functions outside of render.

PR #1433, opened 2021-02-23
Fix check for style amount on layer

kasparia:


bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -71,25 +71,19 @@ Oskari.clazz.define(
...
-                                style.setName(styleId);
-                                style.setTitle(options.styles[styleId].title || styleId);
-                                layer.addStyle(style);
-                            }
-                        });
-                    }
-                    layer.setHoverOptions(options.hover);
-                    layer.selectStyle(defaultStyle.getName());
-                }
+                layer.setHoverOptions(styles.hover);
```
ZakarFin, line 86: Was `options.hover` -> `styles.hover` an intended change? Since this will break any previously saved hover styles I think it needs some discussion or migrations etc. but I'd rather not change it lightly.

kasparia: Whoopsie! Fixed this one back to ` layer.setHoverOptions(mapLayerJson.options.hover);`

PR #1434, opened 2021-02-25
Fix email check for the add new user dialog

mr-lev1:
Currently, if you try to add a new user to Oskari via the admin-users bundle, you're unable to do so if the email is missing. Instead you get the following (a bit confusing) error:

user_add_confusing_notification

To fix this, I added the required-attribute for the email input (similar to the other fields) to let the user know that the email is required, e.g.

user_add_please_fill_notification


ZakarFin:
Hmm, it looks like email is not required but username must be unique based on:
- https://oskari.org/db/tables/oskari_users.html
- https://github.com/oskariorg/oskari-server/blob/master/content-resources/src/main/resources/flyway/oskari/V2_0_1__tables_for_empty_db.sql#L1613-L1614

But it can be required some other part of the software. What breaks if it's empty?

ZakarFin:
Answering to myself, this requires it: https://github.com/oskariorg/oskari-server/blob/master/control-admin/src/main/java/fi/nls/oskari/control/admin/UsersHandler.java#L175

I'm not sure how I feel about having the constraint only on the UI and the handler that is saving the user from this UI. But I guess it's an improvement for this UI anyway so I'll merge this.

PR #1437, opened 2021-02-26
Stop coverage delimiting on tab change

kasparia:
When changing tab on search flyout we want to stop coverage delimiting. Added missing event handler that was notified at search bundle but wasn't handled in the metadata_catalogue end-point.

Changing tab could also reset search state completely in my opinion.

### DIFF
--- a/bundles/catalogue/metadatacatalogue/instance.js
+++ b/bundles/catalogue/metadatacatalogue/instance.js
@@ -360,44 +360,33 @@ Oskari.clazz.define(
                     return;
                 }
 
-                if (!isShown && me.drawCoverage === false) {
-                    me._stopCoverage();
-                    if (me.coverageButton) {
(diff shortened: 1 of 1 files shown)


bundles/catalogue/metadatacatalogue/instance.js
```diff
@@ -377,6 +378,23 @@ Oskari.clazz.define(
...
+                if (event.getNewTabId() !== this.id) {
+                    me._stopCoverage();
+                    if (me.coverageButton) {
+                        me.coverageButton.val(me.getLocalization('delimitArea'));
+                    }
+                    me.drawCoverage = true;
+
+                    if (me.coverageButton) {
+                        me.coverageButton[0].data = '';
+                    }
```
ZakarFin, line 396: We should do some teardown() function that does this and call it instead of copypasting the code.

PR #1440, opened 2021-03-02
Remove printing link from metadata actions

kasparia:


ZakarFin:
This was some paikkatietoikkuna.fi specific stuff and the server side implementation has been removed years ago. The link is removed since there's nothing on the server that handles this link anymore. Not in Oskari and not in Paikkatietoikkuna.

PR #1442, opened 2021-03-03
Remove "old" hierarchical layerlist and admin functionalities

mr-lev1:
The "old" hierarchical layerlist along with its admin functionalities are removed since they will become obsolete when the hierarchical features of the new layerlist will be released.


mr-lev1:
Maybe it would be best to instead replace the glyphicons anyway, as @ZakarFin has suggested.

ZakarFin:
I'll craft some email to the mailing list about the bundles we want to remove after 2.2.0 and lets see if we get some responses to those.

ZakarFin:
Added an issue for the clean up: oskariorg/oskari-docs#245 and email requesting action if these are to be kept in oskari-frontend

ZakarFin:
These bundles have been removed in other pull requests so closing this one (for some reason it isn't closed automatically by GitHub).

PR #1445, opened 2021-03-04
Replace glyphicons with Iconicons

osqq:
Replace Glyphicons with iconicons, remove all glyphicon references and some little css changes to adjust the iconicons


ZakarFin:
Can you remove all the icons that are not used and link the license for these icons?

osqq:
All now removed, license at https://github.com/ionic-team/ionicons/blob/master/LICENSE , should it be included in the folder?

ZakarFin:
Thanks! Based on "Replace glyphicons with Iconicons" I actually thought this was another lib since there's an additional `c` in there :D I think this is good for now 👍 I'll merge this on monday.

osqq:
> Thanks! Based on "Replace glyphicons with Iconicons" I actually thought this was another lib since there's an additional `c` in there :D I think this is good for now 👍 I'll merge this on monday.

Ah no just my dyslexia 😄

PR #1448, opened 2021-03-09
Update layer without refreshing

kasparia:


bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
```diff
@@ -14,7 +14,8 @@ const __VALIDATOR_CACHE = {};
 class UIHandler extends StateHandler {
     constructor (consumer) {
         super();
-        const mapmodule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
+        this.sandbox = Oskari.getSandbox();
+        this.mapmodule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
```
ZakarFin, line 18: It doesn't look like you would need to store a reference to mapmodule in `this` as it's not used by any of these changes?

bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
```diff
@@ -667,14 +668,41 @@ class UIHandler extends StateHandler {
         });
     }
 
+    /**
+     *
+     * @param {Number} layerId id for layer affected
+     * @param {Object} layer affected layer as WFSlayer object
+     * @param {Object} layerData new fetched layer data
+     */
+    refreshLayerOnMap (layerId, layer, layerData) {
```
ZakarFin, line 677: The layer parameter isn't used?

bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
```diff
@@ -667,14 +668,41 @@ class UIHandler extends StateHandler {
...
+        this.mapLayerService.removeLayer(layerId); // remove existing layer and supress event
+
+        // if layer was found from the selected layers remove and re-add it
+        if (originalLayerIndex !== -1) {
+            this.sandbox.postRequestByName('RemoveMapLayerRequest', [layerId]);
+
+            this.mapLayerService.addLayer(modifiedLayer); // add layer but dont supress event
+            this.sandbox.postRequestByName('AddMapLayerRequest', [layerId]);
+            this.sandbox.postRequestByName('RearrangeSelectedMapLayerRequest', [layerId, originalLayerIndex]);
+        }
```
ZakarFin, line 694: Sending RemoveMapLayerRequest when the layer is not registered on the service might lead to unexpected problems since it triggers an event that can't be fully populated with proper data (== layer details). Also with your code the layer is not added to the service/register either if it was NOT on the map. That needs to be done regardless.

```suggestion
        // if layer was found from the selected layers remove and re-add it
        if (originalLayerIndex !== -1) {
            this.sandbox.postRequestByName('RemoveMapLayerRequest', [layerId]);
        }
        // remove the previous version replace with the new layer data without sending events
        // this is a more secure way of updating all of the layer data for the frontend instead of calling updateLayer that does only partial update
        this.mapLayerService.removeLayer(layerId, true);
        this.mapLayerService.addLayer(modifiedLayer, true);
        // TODO: send maplayer event with "update" operation here

        // if layer was found from the selected layers remove it from map and re-add it
        // this handles everything that needs to be updated on the map without separate code to update and potentially changed data separately
        if (originalLayerIndex !== -1) {
            this.sandbox.postRequestByName('RemoveMapLayerRequest', [layerId]);

            this.sandbox.postRequestByName('AddMapLayerRequest', [layerId]);
            this.sandbox.postRequestByName('RearrangeSelectedMapLayerRequest', [layerId, originalLayerIndex]);
        }
```

ZakarFin: Sending update event for maplayer is done like this:
```
                // notify components of layer update
                sandbox.notifyAll(Oskari.eventBuilder('MapLayerEvent')(layerId, 'update'));
```

It will probably do less work on most functionalities leading to better performance than having both remove and add events sent.

bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
kasparia: Changed back to `const mapmodule ...`

bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
kasparia: Removed this.

PR #1457, opened 2021-03-15
Improve the icon usage for timeseries layers in layer list and selected layers

mingfeng:
In layer list:
image

In selected layers:
image

### DIFF
--- a/bundles/framework/layerlist/view/LayerViewTabs/LayerIcon.jsx
+++ b/bundles/framework/layerlist/view/LayerViewTabs/LayerIcon.jsx
@@ -1,9 +1,12 @@
-import React from 'react';
 import PropTypes from 'prop-types';
-import { ImageLayerIcon, DataLayerIcon, UserDataIcon, ThreeDIcon } from './CustomIcons';
+import React from 'react';
+import { DataLayerIcon, ImageLayerIcon, ThreeDIcon, TimeSerieIcon, UserDataIcon } from './CustomIcons';
 
 // TODO replace logic with layer plugins registering their icons.
-export const LayerIcon = ({ type, ...rest }) => {
+export const LayerIcon = ({ type, hasTimeseries = false, ...rest }) => {
+    if (hasTimeseries) {
+        return <TimeSerieIcon {...rest} />;
+    }
     if (['wmts', 'wms', 'arcgis93', 'arcgis'].includes(type)) {
         return <ImageLayerIcon {...rest} />;
     }
@@ -20,5 +23,6 @@ export const LayerIcon = ({ type, ...rest }) => {
 };
 
 LayerIcon.propTypes = {
-    type: PropTypes.string
+    type: PropTypes.string,
+    hasTimeseries: PropTypes.bool
 };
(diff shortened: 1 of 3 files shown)


bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/Layer/LayerTools.jsx
```diff
@@ -62,22 +61,25 @@ const LayerTools = ({ model, controller }) => {
...
     const reasons = !map.isLayerSupported(model) ? map.getUnsupportedLayerReasons(model) : undefined;
     const reason = reasons ? map.getMostSevereUnsupportedLayerReason(reasons) : undefined;
-    const statusOnClick = backendStatus.status !== 'UNKNOWN' ? () => controller.showLayerBackendStatus(model.getId()) : undefined;
+    let layerIcon;
+    if (model.hasTimeseries()) {
+        layerIcon = (
+            <Tooltip title={<Message messageKey="layer.tooltip.timeseries" />}>
+                <TimeSerieIcon />
+            </Tooltip>
+        );
```
ZakarFin, line 70: I'll think on this a bit before merging. I think timeseries shouldn't have a special handling here but should be part of the LayerStatus tag, but it might mean some other changes a bit. I'll get back to this tomorrow probably.

mingfeng: Yeah, I had the same feeling too

bundles/framework/layerlist/view/LayerViewTabs/LayerIcon.jsx
```diff
@@ -1,9 +1,17 @@
...
-import { ImageLayerIcon, DataLayerIcon, UserDataIcon, ThreeDIcon } from './CustomIcons';
+import React from 'react';
+import { DataLayerIcon, ImageLayerIcon, ThreeDIcon, TimeSerieIcon, UserDataIcon } from './CustomIcons';
 
 // TODO replace logic with layer plugins registering their icons.
-export const LayerIcon = ({ type, ...rest }) => {
+export const LayerIcon = ({ type, hasTimeseries = false, ...rest }) => {
+    if (hasTimeseries) {
+        return (
+            <Tooltip title={<Message messageKey="layer.tooltip.timeseries" />}>
```
ZakarFin, line 10: Let's remove the tooltip from here and just use the TimeSerieIcon and pass the ...rest for it like the other icons.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/Layer/LayerTools.jsx
ZakarFin: Let's remove the timeseries special handling from here. Then the icon comes from LayerIcon.jsx which handles timeseries with the other change you made. You'll need to dig out that hasTimeSeries() prop inside LayerStatus so it's passed to LayerIcon though.

PR #1459, opened 2021-03-16
Admin layereditor metadata UUID

okauppinen:
Show both service metadataUuid (capabilities json) and overridden (metadataId) in layereditor.

https://github.com/oskariorg/oskari-server/pull/693

### DIFF
--- a/bundles/admin/admin-layereditor/view/AdminLayerForm/AdditionalTabPane/ServiceMetadata.jsx
+++ b/bundles/admin/admin-layereditor/view/AdminLayerForm/AdditionalTabPane/ServiceMetadata.jsx
@@ -1,0 +1,29 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { Message, Tooltip } from 'oskari-ui';
+import { Controller } from 'oskari-ui/util';
+import { MetadataButton } from './styled';
+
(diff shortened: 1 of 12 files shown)


bundles/admin/admin-layereditor/view/AdminLayerForm/AdditionalTabPane/ServiceMetadata.jsx
```diff
@@ -0,0 +1,29 @@
...
+                <Message messageKey='metadata.service' />
+                <span>:&nbsp;</span>
+                <Message messageKey='serviceNotAvailable' />
+            </Fragment>
+        );
+    }
+    return (
+        <Tooltip title={metadataUuid}>
+            <Message messageKey='metadata.service' />
+            <MetadataButton onClick={() => controller.showLayerMetadata(metadataUuid)}/>
```
ZakarFin, line 21: We should check if the metadata request is available in the app and not show the button if it's not.

bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -1,3 +1,5 @@
+import { ConstantPositionProperty } from "cesium";
```
ZakarFin, line 1: Accidental import?

PR #1464, opened 2021-03-19
Remove the fill color of timeseries metadata boxes

mingfeng:
image

### DIFF
--- a/bundles/framework/timeseries/view/TimeSeriesRange/TimeseriesMetadataService.js
+++ b/bundles/framework/timeseries/view/TimeSeriesRange/TimeseriesMetadataService.js
@@ -7,7 +7,7 @@ const DEFAULT_STYLE = {
         color: 'rgba(0, 0, 0, 0.8)'
     },
     fill: {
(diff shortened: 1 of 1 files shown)


bundles/framework/timeseries/view/TimeSeriesRange/TimeseriesMetadataService.js
```diff
@@ -7,7 +7,7 @@ const DEFAULT_STYLE = {
         color: 'rgba(0, 0, 0, 0.8)'
     },
     fill: {
-        color: 'rgba(24, 219, 218, 0.5)'
+        color: 'rgba(24, 219, 218, 0)'
```
ZakarFin, line 10: Maybe just remove the fill object from style instead of making it transparent?

mingfeng: Removing the fill will use a default fill color (a purplish color) defined somewhere else.

PR #1468, opened 2021-03-23
MyPlaces load features as geojson

okauppinen:
Add download link to MyPlaces tab to allow user download MyPlaces layer features as geojson file.

I don't know if we have to use fetch instead of using window.location.href. Now there is both solutions.

https://github.com/oskariorg/oskari-server/pull/697


ZakarFin:
Let's use window.location.href and remove the fetch-version.

PR #1469, opened 2021-03-23
Wfs selected features

okauppinen:
Don't store viewport feature properties arrays (field values) to WFSLayer on map move. Only FeatureData table and WFSSetPropertyFilter needs them so it's  lighter to use WFS layer plugin getLayerFeaturePropertiesInViewport to gain array of processed  properties objects on map move when FeatureData table is opened or filter set. 

Changed updateLayerProperties to load properties info via GetWFSLayerFields only. Removed storing fields and activeFeatures.

Remove unused methods from WFSLayer.

Update FilterDialog to use propertyTypes object and getLayerFeaturePropertiesInViewport.

Update GFI popup to use property objects.

FeatureData update: https://github.com/oskariorg/oskari-frontend/pull/1581


bundles/framework/divmanazer/component/FilterDialog.js
```diff
@@ -627,24 +628,18 @@ Oskari.clazz.define('Oskari.userinterface.component.FilterDialog',
...
-                        id: fields[i],
-                        name: (locales[i] || fields[i])
+            const localizedNames = layer.getPropertyLabels();
+            if (Object.keys(localizedNames).length) {
+                // map name: locale to array containing localized names
+                return Object.entries(localizedNames)
+                    .map(([id, locale]) => {
+                        return { id, name: locale };
                     });
-                }
```
ZakarFin, line 644: Should we loop propertyTypes() instead and attach the matching localized name for it when found. With this approach we only show properties that have a localized name and skip any that don't.

bundles/mapping/mapwfs2/domain/WFSLayer.js
```diff
@@ -11,23 +11,36 @@ export class WFSLayer extends VectorTileLayer {
...
+    getLegendImage () {
+        return null;
+    }
+
+    isSupportedSrs () {
+        return true;
+    }
+
+    getLayerUrl () {
+        return Oskari.urls.getRoute('GetWFSVectorTile') + `&id=${this.getId()}&srs={epsg}&z={z}&x={x}&y={y}`;
```
ZakarFin, line 42:
```suggestion
        return Oskari.urls.getRoute('GetWFSVectorTile', {
          id: this.getId(),
          srs: epsg,
          z,
          x,
          y
        });
```

ZakarFin: ^That should do the same thing

PR #1471, opened 2021-03-23
Allow ampersand in MyPlaces description

kasparia:


bundles/framework/myplaces3/view/PlaceForm.js
```diff
@@ -326,6 +310,11 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.view.PlaceForm',
 
             this.dialog.moveTo('#toolbar', 'right');
         },
+
+        _decodeFormFieldInput: function (text) {
+            const placeholderElement = new DOMParser().parseFromString(text, 'text/html');
+            return placeholderElement.documentElement.textContent;
+        },
```
ZakarFin, line 317: This is not used.

ZakarFin:
When we remove sanitizing we need to handle the value as text instead of HTML at least in:
- object data table (it's handled as HTML atleast in here: https://github.com/oskariorg/oskari-frontend/blob/2.2.0/bundles/framework/divmanazer/component/Grid.js#L717)
- in "My data" ^same grid component is used there
- in GFI popup  (it's handled as HTML atleast in here: https://github.com/oskariorg/oskari-frontend/blob/2.2.0/bundles/mapping/mapmodule/plugin/getinfo/GetFeatureInfoFormatter.js#L235)

So while removing the sanitize when the feature is added but we need to use it when displaying the content or handle the content as text instead of html.

PR #1473, opened 2021-03-24
Remove name field from admin editor

kasparia:


ZakarFin:
Nope. we need to have the name field in admin. We only want it gone from certain type of layers. Layers of type vectortilelayer. So the LayerComposingModel for vectortilelayers shouldn't include the name field: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileLayerPlugin.js#L40

I would check if it's feasible to add some removeField() function to LayerComposingModel that we could use to take out the "common" field "name" that is included by default. If we remove it from common fields we need to add it manually to all other layertypes.

PR #1478, opened 2021-03-26
Remove unique name on MVT layers

kasparia:


ZakarFin:
Removes the field from admin UI as it's not required or even used for them.

PR #1485, opened 2021-04-01
RPC functions to new service

kuosman:
This PR change RPC functions creations and bindings to new `RpcService`.  

Moved RPC supported function creations to bundles/plugins.
RPC functions creations are separated into the following bundles/plugins:
- `RPC`: getSupportedEvents, getSupportedFunctions, getSupportedRequests, getInfo, resetState, getCurrentState, useState, sendUIEvent
- `AbstractMapModule`: getAllLayers, getMapBbox, getMapPosition, getZoomRange, zoomIn, zoomOut, zoomTo, getPixelMeasuresInScale, setCursorStyle
- `MapModuleClass.ol`: getScreenshot
- `VectorLayerPlugin`: getFeatures

Only RPC added functions are defaults (if not configured `allowedfunctions`), other are optional and exists only when view contains bundle/plugin.

# How to add new function to RPC

This allows all bundles to add new function(s) to RPC (also application-specific) and use them in published map.

For example this adds example function to RPC:
```
// listen to application started event and register new RPC function.
Oskari.on('app.start', function (details) {
    const rpcService = Oskari.getSandbox().getService('Oskari.mapframework.bundle.rpc.service.RpcService');

    if (!rpcService) {
        return;
    }

    rpcService.addFunction('example', function () {
        console.log('New added RPC function');
        return 'my example result';
    });

    // async functions can return a promise and signal failure through reject
    rpcService.addFunction('example2', function () {
            return new Promise((resolve, reject) => {
                setTimeout(() => resolve('my async example result'), 500);
              });
    });
});
```

Now you can call created example function from your published map:
```
channel.example(function(data) {
    console.log(data);
});

channel.example2(function(data) {
    console.log(data);
});
```


bundles/framework/rpc/service/RpcService.js
```diff
@@ -0,0 +1,122 @@
...
+         * @method addFunction
+         * Add function to RPC
+         *
+         * @param func function that should be added
+         */
+        addFunction: function (func) {
+            if (!func && typeof func !== 'function') {
+                return;
+            }
+
```
ZakarFin, line 51: Check that the function has a name and is not called with `addFunction(() => { /* TODO impl */})` or `addFunction(function () { /* TODO impl */ });`
```suggestion

            if (!func.name || !func.name.trim()) {
                me._log.warn('Trying add anonymous RPC function, skipping. Provide name for function.');
                return;
            }
```

bundles/framework/rpc/service/RpcService.js
```diff
@@ -0,0 +1,122 @@
...
+            const me = this;
```
ZakarFin, line 52: We don't really need `me` in this function

bundles/framework/rpc/service/RpcService.js
```diff
@@ -0,0 +1,122 @@
...
+        },
+
+        /**
+         * @public @method getAllowedFunctions
+         * Get allowed functions
+         *
+         * @returns {Object} allowed functions
+         */
+        getAllowedFunctions: function () {
+            return this.instance.__arrayToObject(this._allowedFunctions);
```
ZakarFin, line 72: Consider moving __arrayToObject() to the service or having it in it's own helper.js

bundles/framework/rpc/service/RpcService.js
ZakarFin: Then again processors and transpilers might change the name of a function or remove it. Should we pass the name as string with additional param to this function?

bundles/framework/rpc/service/RpcService.js
```diff
@@ -0,0 +1,122 @@
...
+        },
+
+        /**
+         * @private @method __bindFunction
+         * Bind allowed function to channel
+         *
+         * @param {Function} func
+         */
+        __bindFunction: function (func) {
```
ZakarFin, line 81: `func` is only used to get the name of the function here. Replace it with string parameter for name instead.

bundles/framework/rpc/service/RpcService.js
```diff
@@ -0,0 +1,122 @@
...
+                    };
+                }
+                params = params || [];
+                params.unshift(trans);
+                var value = me._availableFunctions[func.name].apply(me, params);
+                if (typeof value === 'undefined') {
+                    trans.delayReturn(true);
+                    return;
+                }
+                return value;
```
ZakarFin, line 103: I think we should abstract JSChannel from registered functions. This replaces transaction object from functions that _declare_ having parameters and attaches a "done" callback function instead of it to the parameters. Also try/catch the call and use transaction.error() to pass the error to the RPC-caller.

```suggestion
                const funcParams = params || [];
                const requestedFunction = me._availableFunctions[func.name];
                if (requestedFunction.length) {
                    // attach "done" function as first function parameter if it declares having any parameters
                    funcParams.unshift((response) => trans.complete(response));
                }
                try {
                    const value = requestedFunction.apply(me, funcParams);
                    if (typeof value === 'undefined') {
                        // assume it's an async function -> wait for "done()" to be called
                        trans.delayReturn(true);
                        return;
                    }
                    return value;
                } catch (err) {
                    trans.error(err, 'Error calling RPC-function: ' + func.name + 'with params: ' + params);
                }
```

This would mean that for example getScreenshot() on MapModuleClass.ol.js would need to change from:
```
        rpcService.addFunction(function getScreenshot (transaction) {
            getScreenshot(function (image) {
                transaction.complete(image);
            });
        });
```
To this:
```
        rpcService.addFunction(function getScreenshot (done) {
            getScreenshot(done);
        });
```

bundles/framework/rpc/service/RpcService.js
```diff
@@ -0,0 +1,122 @@
...
+        init: function () {
+        },
+
+        /**
+         * @method addFunction
+         * Add function to RPC
+         *
+         * @param func function that should be added
+         */
+        addFunction: function (func) {
```
ZakarFin, line 47: Let's change this so we give the function name as string. Just to make sure the build doesn't mangle it up (if it doesn't now it might after some future update).
```suggestion
        addFunction: function (name, func) {
```
This would change the example to:
```
    rpcService.addFunction('example', function () {
        console.log('New added RPC function');
        return 'example';
    });

File: bundles/framework/rpc/service/RpcService.js (hunk 5 above)
Comment (ZakarFin) [reply bundles/framework/rpc/service/RpcService.js:103]: Any thoughts on this?^

File: api/framework/rpc/bundle.md [hunk 7]
```diff
@@ -13,30 +13,67 @@ Examples of all the RPC functionalities can be found [here](https://www.oskari.o
...
+    });
+});
+
+// now you can call created example function from your published map
+channel.example(function(data) {
+    console.log(data);
+});
+
+```
+
```
ZakarFin, line 41:
```suggestion
    rpcService.addFunction('example', function () {
        console.log('New added RPC function');
        return 'my example result';
    });
    
    // async functions can return a promise and signal failure through reject
    rpcService.addFunction('example2', function () {
            return new Promise((resolve, reject) => {
                setTimeout(() => resolve('my async example result'), 500);
              });
    });
});

// now you can call created example function from your published map
channel.example(function(data) {
    console.log(data);
});

```

```

**Comment by ZakarFin** [comment]
This is awesome! Added a few suggestions to make this clearer for bundles registering functions:
- the registered function name should be given as string as additional param so build-scripts won't change the names
- the registered function should _always_ return something. Async functions can return a promise which allows them to signal failure as well
- functions not returning either a value or a Promise should be considered faulty and trigger the error handling

File: bundles/framework/rpc/service/RpcService.js (hunk 1 above)
Comment (kuosman) [reply bundles/framework/rpc/service/RpcService.js:54]: added name param

File: bundles/framework/rpc/service/RpcService.js (hunk 6 above)
Comment (kuosman) [reply bundles/framework/rpc/service/RpcService.js]: Changed

File: bundles/framework/rpc/service/RpcService.js [hunk 8]
```diff
@@ -0,0 +1,140 @@
...
+         * @method addFunction
+         * Add function to RPC
+         *
+         * @param {String} name function name
+         * @param {Function} func function that should be added
+         */
+        addFunction: function (name, func) {
+            if (!func && typeof func !== 'function') {
+                return;
+            }
```
ZakarFin, line 53: We still need to check that the function has a name. I think it's best to throw an error to make it more noticeable.
```suggestion
            if (!name || typeof func !== 'function') {
                throw new Error('Tried registering function without name or implementation!');
            }
```

bundles/framework/rpc/service/RpcService.js
```diff
@@ -0,0 +1,140 @@
...
+            me.channel.bind(name, function (trans, params) {
+                if (!domainMatch(trans.origin, me.conf.domain)) {
+                    // eslint-disable-next-line no-throw-literal
+                    throw {
+                        error: 'invalid_origin',
+                        message: 'Invalid origin: ' + trans.origin
+                    };
+                }
+
+                const requestedFunction = me._availableFunctions[name];
```
ZakarFin, line 99: We can store reference to the function at the beginning so we don't have to reference _availableFunctions twice.
```suggestion

            const requestedFunction = me._availableFunctions[name];
            if (!requestedFunction) {
                return;
            }

            me.channel.bind(name, function (trans, params) {
                if (!domainMatch(trans.origin, me.conf.domain)) {
                    // eslint-disable-next-line no-throw-literal
                    throw {
                        error: 'invalid_origin',
                        message: 'Invalid origin: ' + trans.origin
                    };
                }

```

bundles/framework/rpc/instance.js
```diff
@@ -499,17 +237,17 @@ Oskari.clazz.define(
...
+                if (Object.prototype.hasOwnProperty.call(me.eventHandlers, p)) {
                     sandbox.unregisterFromEventByName(me, p);
                 }
             }
             for (p in me.requestHandlers) {
-                if (me.requestHandlers.hasOwnProperty(p)) {
+                if (Object.prototype.hasOwnProperty.call(me.requestHandlers, p)) {
                     sandbox.removeRequestHandler(p, this);
                 }
             }
```
ZakarFin, line 253: Maybe just do:
```
Object.keys(me.eventHandlers).forEach(eventName => sandbox.unregisterFromEventByName(me, eventName);
Object.keys(me.requestHandlers).forEach(requestName => sandbox.removeRequestHandler(requestName, me);
```
Funky that the instance is the first param for events but second for requests :) Haven't noticed that before.

ZakarFin:
Thanks!

PR #1486, opened 2021-04-01
Implement a layer swipe tool

mingfeng:
Created a new layer swipe bundle that provides the layer swipe map tool.
The button to activated the tool is inserted to basic toolbar when the
bundle is started. When activated, the topmost layer will be the swipe
layer and appear on left side while the rest of the layers will be on
the right side. Currently only non-vector layers can be the swipe layer.

When there are layers added to or removed from the map, the tool will
update the swipe layer accordingly.

TODO:
[] The new icon needs to be built into the `icons.png` file before it can be used for button

layer-swipe


resources/icons.css
```diff
@@ -1638,6 +1638,14 @@
     background-position: -5092px 0px !important;
 }
 
+.tool-layer-swipe {
+    width: 32px;
+    height: 32px;
+    background-image: url('icons.png') !important;
+    background-repeat: no-repeat !important;
+    background-position: -5060px 0px !important;
+}
```
ZakarFin, line 1647: This is generated when the icons.png sprite is created.

bundles/mapping/layerswipe/instance.js
```diff
@@ -0,0 +1,143 @@
...
+            });
+            this.eventListenerKeys.push(prerenderKey);
+            const postrenderKey = layer.on('postrender', (event) => {
+                event.context.restore();
+            });
+            this.eventListenerKeys.push(postrenderKey);
+        },
+
+        unregisterEventListeners: function () {
+            this.eventListenerKeys.forEach(key => unByKey(key));
```
ZakarFin, line 96: I'm not sure what this does but if it unregisters all listeners for pre/postrender (also registered by other bundles) it will break stuff.

bundles/mapping/layerswipe/instance.js
```diff
@@ -0,0 +1,143 @@
...
+            this.eventListenerKeys = [];
+        },
+
+        getSplitterElement: function () {
+            if (!this.splitter) {
+                this.splitter = jQuery('<div class="layer-swipe-splitter"></div>');
```
ZakarFin, line 102: We are moving away from jQuery. Is there some React-based code that could be used here?

bundles/mapping/layerswipe/instance.js
```diff
@@ -0,0 +1,143 @@
...
+        updateSwipeLayer: function () {
+            this.unregisterEventListeners();
+            const layer = this.getTopmostLayer();
+            if (layer) {
+                this.registerEventListeners(layer);
+            }
+        },
+
+        getTopmostLayer: function () {
+            const layers = this.map.getLayers().getArray().filter(layer => layer.getVisible() && !(layer instanceof VectorLayer));
```
ZakarFin, line 61: Should the localized tooltip say that vectorlayers are not counted as layers? Or why do we filter them out? There might be some layers that we do want to filter out like a draw layer or marker layer but probably not all VectorLayers?

PR #1487, opened 2021-04-01
Move and improve updated and created dates in admin view

kasparia:


bundles/admin/admin-layereditor/view/AdminLayerForm/AdditionalTabPane/LayerGeneralInfo.jsx
```diff
@@ -0,0 +1,93 @@
...
+    const localeDateOptions = {
+        year: 'numeric',
+        month: '2-digit',
+        day: '2-digit'
+    };
+
+    const formattedCreatedTime = typeof layer.created !== 'undefined' ? createdDate.toLocaleTimeString(dateLocale).replace(/\./g, ':') : '--:--:--';
+    const formattedCreatedDate = typeof layer.created !== 'undefined' ? createdDate.toLocaleDateString(dateLocale, localeDateOptions) : '--.--.----';
+    const formattedUpdatedTime = typeof layer.updated !== 'undefined' ? updatedDate.toLocaleTimeString(dateLocale).replace(/\./g, ':') : '--:--:--';
+    const formattedUpdatedDate = typeof layer.updated !== 'undefined' ? updatedDate.toLocaleDateString(dateLocale, localeDateOptions) : '--.--.----';
```
ZakarFin, line 55: Refactor to a function that takes layer.created | updated as param and returns an object with formatted time and date. Then we don't call new Date() for undefined values etc.

```suggestion

const formatTimestamp = (timestamp) => {
    let date;
    if (typeof timestamp !== 'undefined') {
        date = new Date(timestamp);
    }
    return {
        date: formatDate(date),
        time: formatTime(time)
    };
};

const formatDate = (date) => {
    if (typeof date !== 'undefined') {
        return '--.--.----';
    }
    return date.toLocaleDateString(dateLocale, localeDateOptions);
};

const formatTime = (date) => {
    if (typeof date !== 'undefined') {
        return '--:--:--';
    }
    return date.toLocaleTimeString(dateLocale).replace(/\./g, ':');
};

export const LayerGeneralInfo = ({ layer }) => {
    const createdDate = new Date(layer.created);
    const updatedDate = new Date(layer.updated);
    const dateLocale = 'fi-FI'; // we are not using here other locales than Finnish so we can hard core it into constant variable
    const localeDateOptions = {
        year: 'numeric',
        month: '2-digit',
        day: '2-digit'
    };

    const created = formatTimestamp(layer.created);
    const updated = formatTimestamp(layer.updated);
    // probably get rid of these and just use created/updated on JSX
    const formattedCreatedTime = created.time;
    const formattedCreatedDate = created.date
    const formattedUpdatedTime = updated.time;
    const formattedUpdatedDate = updated.date;
```

PR #1508, opened 2021-04-16
Update data year hover text style

mingfeng:
image


mingfeng:
Updated the font for the hover text to use the same font as slider mark label

image

PR #1517, opened 2021-04-20
Layerswipe visible layers

okauppinen:
Get topmost layer from selected Oskari layers instead of ol map layers. Swipe layer is topmost visible layer (raster or vector).

Added isInGeometry (same than in LayersPlugin). I don't know if it should be in (Abstract)MapModule. First tried to handle scale and geomMatch with MapLayerVisibilityChangedEvent but it's seemed to be simpler to check them in getTopmostLayer.

### DIFF
--- a/bundles/mapping/layerswipe/instance.js
+++ b/bundles/mapping/layerswipe/instance.js
@@ -118,10 +118,7 @@ Oskari.clazz.define(
             }
             var viewBounds = this.mapModule.getCurrentExtent();
             var olExtent = [viewBounds.left, viewBounds.bottom, viewBounds.right, viewBounds.top];
-            if (geometries[0].intersectsExtent(olExtent)) {
-                return true;
-            }
-            return false;
+            return geometries[0].intersectsExtent(olExtent);
         },
 
         getTopmostLayer: function () {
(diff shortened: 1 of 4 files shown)


bundles/mapping/layerswipe/instance.js
```diff
@@ -114,12 +110,32 @@ Oskari.clazz.define(
...
+            var geometries = layer.getGeometry();
+            if (!geometries || geometries.length === 0) {
+                return true;
+            }
+            var viewBounds = this.mapModule.getCurrentExtent();
+            var olExtent = [viewBounds.left, viewBounds.bottom, viewBounds.right, viewBounds.top];
+            if (geometries[0].intersectsExtent(olExtent)) {
+                return true;
+            }
+            return false;
```
ZakarFin, line 123:
```suggestion
            return geometries[0].intersectsExtent(olExtent);
```

bundles/mapping/layerswipe/instance.js
```diff
@@ -114,12 +110,32 @@ Oskari.clazz.define(
             popup.show(title, message, [closeBtn]);
         },
 
+        isInGeometry: function (layer) {
+            var geometries = layer.getGeometry();
+            if (!geometries || geometries.length === 0) {
```
ZakarFin, line 115:
```suggestion
            if (!geometries || geometries.length === 0) {
                // we might not have the coverage geometry so assume all is good if we don't know for sure
```

PR #1522, opened 2021-04-21
Add check for empty indicators

kasparia:


ZakarFin:
The statistical maps opens the indicator search window if the region set layer is added to the map when there's no indicators selected (to guide users better).

PR #1550, opened 2021-04-30
Fix "did you mean React.forwardRef()" errors

ZakarFin:
AntD confirm that has a direct child of a styled-component 'tag' gives this error. It works but the dev-console shows an error. Wrapping the `<StyledX` to an empty span is not clean or elegant but it removes the error from being generated. Maybe we should create a wrapper component just for this purpose instead?

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `styled__StyledButton`.
    in Button (created by styled__StyledButton)
    in styled__StyledButton (created by Trigger)
    in Trigger (created by ForwardRef(Tooltip))
    in ForwardRef(Tooltip) (created by Tooltip)
    in Tooltip (created by ForwardRef)
    in ForwardRef (created by Confirm)
    in Confirm (created by AdminLayerForm)
    in div (created by styled__StyledRoot)

### DIFF
--- a/bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerForm.jsx
+++ b/bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerForm.jsx
@@ -106,7 +106,7 @@ const AdminLayerForm = ({
                     cancelText={getMessage('cancel')}
                     placement='bottomLeft'
                 >
-                    <StyledButton>
+                    <StyledButton danger>
                         <Message messageKey='delete'/>
                     </StyledButton>
                 </Confirm>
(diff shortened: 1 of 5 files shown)


ZakarFin:
I think this approach is fragile since it's very likely developers will clean out the extra span when they look at the code not knowing that the `span` was added to fix this kind of issue (I know I would after I'd forgotten about it). Maybe better approach would be to add the wrapper span on these components so we can document _why_ it's there on just these two components:
- https://github.com/oskariorg/oskari-frontend/blob/master/src/react/components/Confirm.jsx#L10 - https://github.com/oskariorg/oskari-frontend/blob/master/src/react/components/Tooltip.jsx 

It would add an extra element to the DOM even for the tooltips/confirms that don't need it but even then I think it would be a better approach.

ZakarFin:
TODO: Check if we can remove the additional span added in #1544 after this.

PR #1564, opened 2021-05-06
Popup react container

okauppinen:


ZakarFin:
We now have a new way of using popups with React (https://github.com/oskariorg/oskari-frontend/blob/2.6.0/src/react/components/window/index.js#L76) so closing this as ~deprecated

PR #1565, opened 2021-05-06
Popup show fadeout fix

okauppinen:
Revert fadeout on popup.show() and remove click-close without actions  because there could be other close handlers and fadeout and click-close caused unwanted behavior.

Add fadeout to admin showMessage.


ZakarFin:
Fixes issues created by #1561 since it turns out we have a bunch of existing popups that don't have "actions" or close icon that we still don't want to close automatically.

PR #1569, opened 2021-05-07
Layer description style fixes

kasparia:


bundles/framework/maplegend/Flyout.js
```diff
@@ -127,10 +127,27 @@ Oskari.clazz.define('Oskari.mapframework.bundle.maplegend.Flyout',
...
+                    layerTitle.find('div.layer-description').hide(); // no functionality -> hide
+                } else {
+                    layerTitle.find('div.icon-info').on('click', function (event) {
+                        event.stopPropagation();
+
+                        sandbox.postRequestByName('catalogue.ShowMetadataRequest', [{
+                            uuid: uuid
+                        }]);
+                    });
+                }
```
ZakarFin, line 145: It's cleaner to add the tool container when needed instead of always adding it and removing it if there's no uuid:
```suggestion
                const uuid = layer.getMetadataIdentifier();
                if (uuid) {
                    layerTitle.append(me.templateTools.clone());
                    layerTitle.find('div.icon-info').on('click', function (event) {
                        event.stopPropagation();
                        sandbox.postRequestByName('catalogue.ShowMetadataRequest', [{
                            uuid: uuid
                        }]);
                    });
                }
```

PR #1579, opened 2021-05-19
SearchPlugin published mobile search area width

okauppinen:
Include margins in search area width calculation

### DIFF
--- a/bundles/mapping/mapmodule/plugin/search/SearchPlugin.js
+++ b/bundles/mapping/mapmodule/plugin/search/SearchPlugin.js
@@ -570,16 +570,16 @@ Oskari.clazz.define(
                 'background-image': 'url("' + bgRight + '")',
                 'width': style.widthRight + 'px'
             });
-
+            const searchArea = div.find('.search-area-div');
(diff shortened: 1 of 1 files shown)


bundles/mapping/mapmodule/plugin/search/SearchPlugin.js
```diff
@@ -570,16 +570,16 @@ Oskari.clazz.define(
                 'background-image': 'url("' + bgRight + '")',
                 'width': style.widthRight + 'px'
             });
-
+            const searchArea = div.find('.search-area-div');
             // calculate the width for the middle container of the search
-            var middleWidth = parseInt(jQuery('.search-area-div').css('width')) - style.widthLeft - style.widthRight;
+            var middleWidth = parseInt(searchArea.css('width')) - style.widthLeft - style.widthRight;
```
ZakarFin, line 575: middleWidth is not used and could be removed?

ZakarFin: Nevermind, it's used on the next row for middle.css()

PR #1580, opened 2021-05-19
Admin LayerEditor raster legends

okauppinen:
Add legends without style to select options to allow edit/delete (marked with ( ! )).
Remove LegendImage.jsx and show style unavailable instead of style select.
Show warning notifications if default/selected style doesn't exist, there is additional overridden legends (doesn't have style) or global/general legend (legendImage) is used even layer has styles.


bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/RasterStyle.jsx
```diff
@@ -1,68 +1,86 @@
...
+        warnings.push('globalWithStyles');
+    }
+    return warnings;
+};
+
+const legendsWithoutStyle = (styles, legends) => {
+    const styleNames = styles.map(s => s.name);
+    return Object.keys(legends)
+        .filter(name => !styleNames.includes(name));
+};
```
ZakarFin, line 38: As `rasterStyleCapabilities()` is only used from handler (outside the component) and `legendsWithoutStyle()` is used by both rasterStyleCapabilities() and the component I would move both of them to some RasterStyle/helper or util.js file. As the functions don't need React at all I think they would benefit from being in a file of their own.

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/RasterStyle.jsx
```diff
@@ -1,68 +1,86 @@
...
+
+const additionalLegendsToStyles = (styles, legends, globalTitle) => {
```
ZakarFin, line 40: Maybe add a comment to clarify this?
```suggestion
// Used to generate style-objects for styles that have been removed from the service
// so we can show the override legends urls that we have saved for the layer
// and notify admin that such styles don't exist any more on the service
const additionalLegendsToStyles = (styles, legends, globalTitle) => {
```

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/RasterStyle.jsx
```diff
@@ -1,68 +1,86 @@
...
-        return label;
-    };
-    const onDefaultStyleChange = (styleName, selected) => {
-        const defaultStyle = selected ? styleName : '';
-        controller.setStyle(defaultStyle);
-    };
+    const firstOption = styleOptions.length > 0 ? styleOptions[0].name : '';
+    const [selected, setSelected] = useState(defaultName || firstOption);
+
+    const style = styles.find(s => s.name === selected);
```
ZakarFin, line 63: Shouldn't `layer.style` be used somewhere here?

ZakarFin: Nevermind, just spotted `defaultName` is destructured from `layer.style` at the start of the component.

bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
```diff
@@ -831,6 +833,20 @@ class UIHandler extends StateHandler {
...
+    validateCapabilities () {
+        const { layer, propertyFields } = this.getState();
+        if (propertyFields.includes(LayerComposingModel.CAPABILITIES_STYLES)) {
+            rasterStyleCapabilities(layer).forEach(field => {
+                const options = {
+                    duration: null,
+                    title: getMessage('capabilities.validate'),
+                    content: getMessage(`capabilities.rasterStyle.${field}`)
+                };
+                Messaging.warn(options);
```
ZakarFin, line 845: This shows a warning popup for each style that has been removed and has an override legend? Maybe we could just show one and gather up the messages in content to a list? Also renaming rasterStyleCapabilities() to something like getWarningsForRemovedStyles() or something like that might help read it better :)

bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/RasterStyle.jsx
```diff
@@ -1,68 +1,86 @@
...
+    const { legends = {} } = options;
+    const selectedExists = styles.some(s => s.name === style);
+    const warnings = [];
+    if (style && !selectedExists) {
+        warnings.push('defaultStyle');
+    }
+    const additionals = legendsWithoutStyle(styles, legends);
+    const hasGlobal = additionals.find(name => name === GLOBAL_LEGEND);
+    // don't notify if global is only additional legend
+    if (additionals.length > 1 || (additionals.lenth === 1 && !hasGlobal)) {
```
ZakarFin, line 25:
```suggestion
    if (additionals.length > 1 || (additionals.length === 1 && !hasGlobal)) {
```

bundles/admin/admin-layereditor/view/AdminLayerForm/AdminLayerFormHandler.js
ZakarFin: Then again it looks like it's only max two notifications so not for each style. I think that's ok.

PR #1581, opened 2021-05-20
Featuredata refactor propeties handling

okauppinen:
Refactor featuredata to handle feature properties after: https://github.com/oskariorg/oskari-frontend/pull/1469

Added statehandler to trigger rendering. Flyout content and Grid uses jQuery so had to make some ugly jQuery optimization (some state changes doesn't trigger full render).

Splitted _prepareData to smaller functions.

Force GFI to be disabled until Flyout is closed e.g. reset state doesn't enable GFI if Flyout is attached.

Note! locateOnMap doesn't add icons to grid but it did't work properly before these changes. If manualRefresh layers aren't used then locateOnMap could be removed from Featuredata.


ZakarFin:
It's likely that something breaks between changes in this and #1469 but we should have plenty of time to fix the issues before the next release.

PR #1587, opened 2021-05-23
Announcements bundle tool and plugin

osqq:
Changes contain announcements bundle's tool and plugin for map publishing. Tool allows selecting of valid announcements to be published with map, plugin is just a simple collapse to show these selected announcements. Plugin also contains a function that RPC can use to fetch announcements.


bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+            me.data = data;
+            me.selectedAnnouncements = [];
+            me.annTitles = [];
+
+            if (data.configuration && data.configuration.mapfull && data.configuration.mapfull.conf && data.configuration.mapfull.conf.plugins) {
+                _.each(data.configuration.mapfull.conf.plugins, function (plugin) {
+                    if (me.getTool().id === plugin.id) {
+                        me.setEnabled(true);
+                    }
+                });
```
ZakarFin, line 44:
```suggestion
                const myId = this.getTool().id;
                me.setEnabled(data.configuration.mapfull.conf.plugins.some(plugin => myId === plugin.id));
```

bundles/framework/publisher2/resources/locale/fi.js
```diff
@@ -280,6 +281,13 @@ Oskari.registerLocalization(
                 "publishable": "Näytä vain julkaistavissa olevat karttatasot"
             }
         },
+        "AnnouncementsTool": {
+            "buttonLabel": "Valitse Ilmoitukset",
+            "popup": {
+                "title": "Ilmoitukset",
+                "close": "Sulje"
+            },
+        },
```
ZakarFin, line 290: You already have these in the announcement bundle localizations. Why add them again?

bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+        selectedAnnouncements: [],
+        annTitles: [],
+        announcements: {},
+
+        isAnnouncementsDialogOpen: false,
+
+        groupedSiblings: true,
+
+        templates: {
+            announcements: jQuery('<div id="publisher-layout-announcements" class="tool-options">' + '<div id="publisher-layout-announcementsSelector">' + '<input type="text" name="publisher-announcements" id="publisher-announcements" disabled />' + '<button id="publisher-announcements-button"></button>' + '</div>' + '</div>'),
```
ZakarFin, line 18: Do you really need an id for _all_ of the elements?

bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+            const toolPluginMapfullConf = this._getToolPluginMapfullConf();
+            if (toolPluginMapfullConf != null) {
+                me._sendAnnouncementsChangedEvent(toolPluginMapfullConf.config.announcements);
+                toolPluginMapfullConf.config.announcements.forEach(announcement => {
+                    me.annTitles.push(announcement.title);
+                    me.selectedAnnouncements.push(announcement);
+                });
+            }
+            this.fetchAnnouncements();
+            jQuery('div.basic_publisher').find('input[name=publisher-announcements]').val(me.annTitles.toString()).attr('announcement-name', me.annTitles.toString());
```
ZakarFin, line 61: Whatever this does, find another way of doing it.

bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+                type: 'GET',
+                dataType: 'json',
+                url: Oskari.urls.getRoute('Announcements'),
+                success: (pResp) => {
+                    this.announcements = pResp.data;
+                },
+                error: function (jqXHR, textStatus) {
+                }
+            });
+        },
```
ZakarFin, line 254: Aren't announcements already loaded to the frontend if the appsetup has announcements bundle included? Maybe refactor to a service and get the data from there?

bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+            if (me.state.enabled) {
+                var pluginConfig = { id: this.getTool().id, config: this.getPlugin().getConfig() };
+                var announcementsSelection = me._getAnnouncementsSelection();
+
+                if (announcementsSelection && !jQuery.isEmptyObject(announcementsSelection)) {
+                    pluginConfig.config.announcements = announcementsSelection.announcements;
+                }
+                return {
+                    configuration: {
+                        mapfull: {
```
ZakarFin, line 285: Reconsider having this announcements bundle as part of mapfull plugins. A weird dependency is created for mapfull here if it's part of these plugins.

bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+        },
+
+        /**
+        * Sends an event to notify interested parties that the announcements have changed.
+        *
+        * @method _sendAnnouncementsChangedEvent
+        * @param {Object} announcements the changed announcement
+        */
+        _sendAnnouncementsChangedEvent: function (announcements) {
+            this._sendEvent('Announcements.AnnouncementsChangedEvent', announcements);
```
ZakarFin, line 321: I can see this event useful as part of the announcements functionality. I don't see it as a good practice if it's only used to transfer information between this publisher tool and the plugin. You already have and use a reference to the plugin on this tool so you could just call a function on that plugin instead.

bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+        _getItemByCode: function (code, list) {
+            var listLen = list.length,
+                i;
+
+            for (i = 0; i < listLen; ++i) {
+                if (list[i].title === code) {
+                    return list[i];
+                }
+            }
+            return null;
```
ZakarFin, line 342:
```suggestion
     return list.find(item => item.title === code)
```

ZakarFin: I don't see why this would need it's own function. It only obfuscates reading the code.

bundles/framework/announcements/tool/AnnouncementsTool.js
```diff
@@ -0,0 +1,367 @@
...
+                // check if announcement is already checked, if is, add/remove accordingly
+                if (!this.checked) {
+                    me.selectedAnnouncements = me.selectedAnnouncements.filter(function (ann) {
+                        return ann.title !== announcement.title;
+                    });
+                    me.annTitles = me.annTitles.filter(function (e) { return e !== announcement.title; });
+                    me._sendAnnouncementsChangedEvent(me.selectedAnnouncements);
+                } else {
+                    me.selectedAnnouncements.push(announcement);
+                    me.annTitles.push(announcement.title);
```
ZakarFin, line 183: Why would you keep track of this in 2 separate arrays?

bundles/framework/announcements/event/AnnouncementsChangedEvent.js
```diff
@@ -0,0 +1,40 @@
...
+        /**
+         * @method getName
+         * Returns event name
+         * @return {String}
+         */
+        getName: function () {
+            return this.__name;
+        },
+        /**
+         * @method getStyle
```
ZakarFin, line 27: Update

bundles/framework/announcements/event/AnnouncementsChangedEvent.js
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @method create called automatically on construction
+     * @static
+     * @param {Object} announcements new list of announcements
+     */
+    function (announcements) {
+        this._announcements = announcements;
+    }, {
+        /** @static @property __name event name */
+        __name: 'Announcements.AnnouncementsChangedEvent',
```
ZakarFin, line 17: Maybe we can do it without the `Announcements.` prefix?

bundles/framework/publisher2/resources/scss/style.scss
```diff
@@ -23,6 +23,10 @@
         }
     }
 }
+      
+#publisher-announcements {
```
ZakarFin, line 27: Don't use ids when you don't need to. And shouldn't this be inside `div.basic_publisher`?

bundles/framework/publisher2/resources/scss/style.scss
```diff
@@ -51,6 +55,10 @@ div.basic_publisher {
         width: 125px;
     }
 
+    div#publisher-layout-announcements input[name=publisher-announcements] {
```
ZakarFin, line 58: See above

bundles/framework/publisher2/resources/scss/style.scss
```diff
@@ -59,6 +67,10 @@ div.basic_publisher {
         float: right;
     }
 
+    button#publisher-announcements-button {
```
ZakarFin, line 70: See above

bundles/framework/publisher2/resources/scss/style.scss
```diff
@@ -361,4 +373,8 @@ div#publisher-custom-colours {
 
 #publisher-layout-colours {
     padding-left: 28px;
+}
+
+#publisher-layout-announcements {
```
ZakarFin, line 378: See above

bundles/mapping/mapmodule/resources/locale/bg.js
```diff
@@ -20,6 +20,12 @@ Oskari.registerLocalization(
             "zoomLvl-12": ""
         },
         "plugin": {
+            "AnnouncementsPlugin": {
+                "terms": "Условия на ползване",
+                "dataSources": "",
+                "layersHeader": "",
+                "indicatorsHeader": ""
+            },
```
ZakarFin, line 28: ?

packages/mapping/ol/mapmodule/bundle.js
```diff
@@ -144,6 +144,14 @@ Oskari.clazz.define(
...
                     'src': '../../../../bundles/mapping/mapmodule/resources/scss/indexmap.ol.scss'
                 },
+                // AnnouncementsPlugin
+                {
+                    'type': 'text/javascript',
+                    'src': '../../../../bundles/mapping/mapmodule/plugin/announcements/AnnouncementsPlugin.js'
+                }, {
+                    'type': 'text/css',
+                    'src': '../../../../bundles/mapping/mapmodule/resources/scss/announcementsplugin.ol.scss'
+                },
```
ZakarFin, line 154: This should be part of `announcements` bundle or something but not part of mapmodule.

ZakarFin:
So this lets end-users choose from announcements which ones they would like to copy to their embedded map? Looks like the copied announcements don't expire like the usual announcements and new ones aren't shown either. They can only select from the set they see when they enter the page when publishing a map.

So is there an actual benefit for the end-user here? Feels that the publisher tool should be part of the announcements-admin bundle at best and I really don't like the display part to be part of mapmodule. Also, all new UI code should be made with React instead of jQuery (at least the plugin, the publisher tool might make more sense to use jQuery instead).

bundles/mapping/mapmodule/resources/locale/bg.js
osqq: No idea, most likely accidental

ZakarFin: Ok, please remove :)

bundles/framework/announcements/tool/AnnouncementsTool.js
ZakarFin: I heard this comment caused some grief so I'll try to rephrase: Can you tell me what this line does and why is it done. The problem I have with it is that it tries to find a single input from the whole publisher panel and set a string value and attribute as string but as far as I can see the variables are lists so I have no idea why this is done. Can you walk me through it?

bundles/framework/announcements/tool/AnnouncementsTool.js
ZakarFin: In my opinion the plugin should be part of the announcements bundle and not part of the "mapmodule" plugins.

bundles/framework/announcements/tool/AnnouncementsTool.js
osqq: Just for ease and not needing to filter trough all the announcement data every time we need just the title.

bundles/framework/announcements/tool/AnnouncementsTool.js
osqq: I just took it from another plugin/tool code, I’ll refactor accordingly

bundles/framework/announcements/tool/AnnouncementsTool.js
osqq: The input field is bit misleading, but basically we just get the (not editable) input field and set the selected announcements’ titles there. There could be a better way to do this, but it’s just implemented in the same style as in the ”styling” tool if I remember correctly.

bundles/framework/announcements/tool/AnnouncementsTool.js
ZakarFin: Reason being that if it's part of mapmodule plugins _everyone_ will need to have that code loaded to the browser even if the Oskari instance isn't using the announcements functionality. So it will increase the size of the js-file end-users need to load to get the application up and running even when the functionality isn't used.

bundles/framework/announcements/tool/AnnouncementsTool.js
ZakarFin: But isn't this already done in getExtraOptions() where it makes more sense? And what happens with fetchAnnouncements() above? It's just fired to load announcements to a variable which will complete at future time but nothing is triggered from them being loaded.

ZakarFin: I'm not sure if I understand what it is that is the expected behavior for this whole functionality. Since copies of announcements are saved to the plugin config and they are restored as choices here. But then the "actual" announcements are loaded as well but nothing seems to handle that really. So should the extra options list the currently available announcements in the instance. Or should it list the previously selected announcements. And what happens if the previously selected announcements are not longer part of the instances announcements? That's what doesn't make sense to me on this functionality as a whole.

Can you explain what the intended use is so I can make better evaluations for these changes. Making copies of announcements to the embedded maps is bad practice since you lose all control over the announcements. To me it would make more sense to save the announcement ids and fetch them by id on the embedded map. Then if they have expired -> no announcement is shown. And if they are edited the edited version is shown. As it is, it seems like a weird CMS to provide some popup notice on an embedded map BUT it requires both the announcements functionality and admin permissions for the user that is creating the embedded map to make it work in any sensible way. However since the publisher stuff for the announcements is shown for all users it feels a bit broken to me.

bundles/framework/announcements/tool/AnnouncementsTool.js
ZakarFin: Just to be clear, there was nothing specifically wrong here. But we are trying to get rid of lodash since we now have most of the functions we use it for natively. Hence suggesting that we use native functions instead of an additional library when we don't need it. Sorry for not being clear on that before.

bundles/framework/announcements/event/AnnouncementsChangedEvent.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @class Oskari.framework.announcements.event.AnnouncementsChangedEvent
+ *
+ * Used to notify getinfo plugin that the announcements have changed.
```
ZakarFin, line 4: I don't think this is used for notifying getinfo plugin of anything? If I read the code correctly it's used to only communicate changes in selections between the publisher tool and the plugin while publishing an embedded map. As such I would remove this event entirely.

bundles/framework/announcements/tool/AnnouncementsTool.js
ZakarFin: I'm not sure what you mean. The other array has the announcements as is while the other has just the titles? Wouldn't it make more sense to have just one array and instead of getting the title with `arrayItem` you would get `arrayItem.title` when you need just the title?

osqq:
Closed and new PR made to avoid confusion because of a lot of changes.

PR #1594, opened 2021-06-04
Update AntD to 4.16.1

kasparia:


ZakarFin:
Considering if we want this for 2.4.0 (do we have enough time to test this) or should we update after the 2.4.0. Btw there's a 4.16.2 available for AntD.

kasparia:
Maybe we should wait and update AntD stright to 4.16.2

PR #1606, opened 2021-06-18
Handle WMTS source opt 'wrapX'

ZakarFin:
Make sure we pass wrapX from layer options to WMTSSource. For some reason the OL function merging the config with WMTS-capabilities loses this flag. It's required for example in Arctic-SDI geoportal for showing WMTS with EPSG:4326 matrix on the polar projections/looking at north pole (EPSG:3575 etc).


ZakarFin:
Looks like with ol 6.4.3 (our current version) and 6.5.0 OpenLayers just returns `false` for wrapX always: https://github.com/openlayers/openlayers/blob/v6.4.3/src/ol/source/WMTS.js#L451

On the main branch there's more logic on it: https://github.com/openlayers/openlayers/blob/45f88eb107605056db80576837c3a4229fe52ecd/src/ol/source/WMTS.js#L500-L503

PR #1608, opened 2021-06-23
Metadatacatalogue localization

ZakarFin:
Something funky going on in metadatacatalogue localization variables. This should be fixed but only doing a draft to remember this.

### DIFF
--- a/bundles/catalogue/metadatacatalogue/instance.js
+++ b/bundles/catalogue/metadatacatalogue/instance.js
@@ -217,7 +217,7 @@ Oskari.clazz.define(
             if (key && this._localization[key]) {
                 return this._localization[key];
             }
-            if (!this.localization) {
+            if (!this._localization) {
                 return {};
             }
             return this._localization;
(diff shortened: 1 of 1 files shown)


ZakarFin:
Same problem in https://github.com/oskariorg/oskari-frontend/blob/master/bundles/catalogue/metadatacatalogue/instance.ol2.js and we could remove the ol2 impl. This needs to be looked at more closely.

ZakarFin:
Similar issue here: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/admin-layerrights/instance.js#L96 <> https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/admin-layerrights/instance.js#L75

ZakarFin:
And here:
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/maplegend/instance.js#L64
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/maplegend/instance.js#L104

ZakarFin:
And here:
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/printout/instance.js#L79
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/printout/instance.js#L98

ZakarFin:
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/search/instance.js#L81
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/search/instance.js#L103

ZakarFin:
Most of these have been rewritten with React so there's no need to take a look at the old ones anymore.

PR #1610, opened 2021-06-24
Add empty fill option

kasparia:


ZakarFin:
We need to add the logic for "no fill" as well. Otherwise it's easy to forget and then we have an option that doesn't really work.

PR #1613, opened 2021-06-29
VectorFeatureService common hover handler

okauppinen:
Handle feature hover with common hover handler. Hover handler handles tooltip and feature style except VectorLayer which haves feature based styling. With common handler we get rid of isHovered in style function and doesn't have to call layer.changed() which runs style function for all features and layers. Also hover state is in one place (except VectorLayer plugin has own hover state for feature styling). Added throttle for forEachFeatureAtPixel which is quite heavy operation if there is lot of layers and features (subsequent hover events might be triggered < 10ms and even twice to same px).

Now HoverHandler is under mapmodule/service as VectorFeatureService. Maybe could be moved to better place.

Next step is to try to add only one layer for WFS layers and when layer is changed then style, opacity and visibility is updated. Also selected feature styles could be handled same common way as hover to simplify style functions.

Didn't have time to move getStyleForGeometry from wfs2. Now hoverhandler imports function from mapwfs2/plugin/WfsVectorLayerPlugin/util/style. Maybe we could try to improve wfs2 to add geom typed styles only if geometry type is unknown or collection. Also tests have some issues with imports.


bundles/mapping/mapmodule/service/HoverHandler.js
```diff
@@ -0,0 +1,265 @@
...
+                return feature => {
+                    if (this.state.renderFeatureId === feature.get(FTR_PROPERTY_ID)) {
+                        return getStyleForGeometry(feature.getType(), olStyles);
+                    }
+                };
+            }
+            return feature => {
+                return getStyleForGeometry(feature.getGeometry(), olStyles);
+            };
+        }
```
ZakarFin, line 134: Reverse this so don't get as much indentation:
```suggestion
        if (!hoverDef) {
            return null;
        }
        if (hoverDef.inherit === true) {
            let layerStyleDef = layer.getCurrentStyleDef() || {};
            if (!layerStyleDef.featureStyle) {
                // Bypass possible layer definitions
                Object.values(layerStyleDef).find(obj => {
                    if (obj.hasOwnProperty('featureStyle')) {
                        layerStyleDef = obj;
                        return true;
                    }
                });
            }
            hoverDef = jQuery.extend(true, {}, defaultFeatureStyle, layerStyleDef.featureStyle, hoverDef);
        }
        // TODO: if layer contains only one geometry type return olStyle (hoverDef) instead of function
        const olStyles = this.styleFactory(hoverDef);
        if (layer.getLayerType() === VECTOR_TILE_TYPE) {
            return feature => {
                if (this.state.renderFeatureId === feature.get(FTR_PROPERTY_ID)) {
                    return getStyleForGeometry(feature.getType(), olStyles);
                }
            };
        }
        return feature => {
            return getStyleForGeometry(feature.getGeometry(), olStyles);
        };
```

ZakarFin:
There's a lot of changes but I'm fine with merging this after the ESLint issue is fixed (+ I added a suggestion for removing indentation). We should be able to test it works properly before the next release.

PR #1614, opened 2021-07-01
Pre defined colors

kasparia:


src/react/components/StyleEditor/constants.js
```diff
@@ -39,3 +39,67 @@ export const LINE_STYLES = {
         }
     ]
 };
+
+export const PRE_DEFINED_COLORS = [
+        {
+            "name": "#ffffff",
+            "data": "<svg viewBox=\"0 0 32 32\" width=\"32\" height=\"32\" xmlns=\"http://www.w3.org/2000/svg\"><rect x=\"0\" y=\"0\" width=\"32\" height=\"32\" fill=\"#ffffff\"/></svg>"
```
ZakarFin, line 46: Looks like nothing in the SVG changes other than the color. Would be much cleaner to take it out and just have it once where it's being rendered. So PRE_DEFINED_COLORS would be just the name without data and you could add some "color" component that has the SVG and takes the color as a prop?

src/react/components/StyleEditor.jsx
```diff
@@ -22,7 +22,7 @@ const TabSelector = styled(Radio.Group)`
 `;
 
 const StaticForm = styled(Form)`
-    width: 400px;
+    width: 522px;
```
ZakarFin, line 25: Do we need both: the pixel width here AND `width={ 620 }` in AdminLayerForm/VisualizationTabPane/VectorStyle.jsx? Could we make it work with one or the other?

kasparia: Removed this.

PR #1615, opened 2021-07-27
UserStyle bundle

okauppinen:
Move userstyle functionality from mapwfs2 to own bundle. Refactor request and handler names. Rename isCreateNew to showStyle. showStyle indicates if list or visuform should be shown. If styleName is missing then new style will be created. Title is hooked to visuform options to get empty name input (setOskariStyleValues rejects empty title/name). Could also handle this by creating always new visuform and pass title in options but since visuform will be replaced by react implementation didn't pay much attention to handling visuform.


ZakarFin:
The request name change `ShowOwnStyleRequest` -> `ShowUserStylesRequest` should be documented in oskari-frontend/api/CHANGELOG.md. It doesn't look like it's documented at all in https://github.com/oskariorg/oskari-frontend/tree/master/api/mapping/mapmodule/request but I guess we should add a note about it.

bundles/mapping/heatmap/domain/HeatMapLayerModelBuilder.js
```diff
@@ -28,7 +28,7 @@ Oskari.clazz.define(
             toolHeatmap.setTitle(locLabel);
             toolHeatmap.setTooltip(locLabel);
             toolHeatmap.setCallback(function () {
-                me.sandbox.postRequestByName('ShowOwnStyleRequest', [layer.getId()]);
+                me.sandbox.postRequestByName('ShowUserStylesRequest', [layer.getId()]);
             });
             layer.addTool(toolHeatmap);
```
ZakarFin, line 33: It doesn't do so now since we assume the style editor is always present but I think we should check if the `ShowUserStylesRequest` request is available in the application and only then create the tool for layers.

bundles/mapping/mapwfs2/domain/WfsLayerModelBuilder.js
```diff
@@ -50,7 +50,7 @@ Oskari.clazz.define(
                 toolOwnStyle.setIconCls('show-own-style-tool');
                 toolOwnStyle.setTooltip(locOwnStyle);
                 toolOwnStyle.setCallback(function () {
-                    me.sandbox.postRequestByName('ShowOwnStyleRequest', [layer.getId(), undefined, false]);
+                    me.sandbox.postRequestByName('ShowUserStylesRequest', [layer.getId(), false]);
                 });
                 layer.addTool(toolOwnStyle);
```
ZakarFin, line 55: Now that the editor has been separated to another bundle I think we should check if the ShowUserStylesRequest request is available in the application and only then create the tool for layers.

ZakarFin:
We also need to document _somewhere_ that the functionality has been separated to another bundle. Applications will need to add it to their main.js for packaging it on geoportal views AND add it to their appsetups on the db (or by other means ~= hardcoding starting it after the app on geoportal views). I'm not sure if this documentation should be part of the API changelog, releasenotes or the migration guide on the oskari-server repository. Maybe the server migration guide + frontend release notes?

ZakarFin:
Otherwise looks good for merging. This will require changes in apps so we need to do them to the sample-application and pti-repos at least.

PR #1618, opened 2021-08-11
Change MyPlaces style editor to React component

kasparia:


kasparia:
SVGs still cause problem after saving form. On rerender apparently React component mistakenly adds fill styles again to DOM so pointing IDs for SVG get mixed up.

bundles/framework/myplaces3/CategoryHandler.js
```diff
@@ -180,42 +180,6 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.CategoryHandler',
             this._notifyUpdate();
         },
         editCategory: function (categoryId) {
-            var me = this;
```
ZakarFin, line 183: editCategory is now an empty function?

bundles/framework/myplaces3/MyPlacesStyleForm.jsx
```diff
@@ -0,0 +1,73 @@
...
+    };
+
+    const setName = (name) => {
+        setEditorState({ ...editorState, styleName: name });
+    };
+
+    return (
+        <LocaleProvider value={{ bundleKey: 'MyPlaces3' }}>
+            <CategoryButton onClick={ () => setEditorState({ ...editorState, modalVisibility: true }) }>
+                { getMessage('tab.editCategory') }
```
ZakarFin, line 44: Do you need `getMessage()` if you just use the `<Message>`-tag here?

bundles/framework/myplaces3/MyPlacesTab.js
```diff
@@ -94,36 +97,22 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.MyPlacesTab',
...
-                    exportLink.attr('title', this.loc('tab.export.tooltip'));
-                    exportLink.on('click', () => {
-                        window.location.href = this.instance.getService().getExportCategoryUrl(categoryId);
-                        return false;
-                    });
-                    panel.getContainer().append(exportLink);
+                    ReactDOM.render(
+                        <MyPlacesStyleForm
+                            layer={{ ...values, categoryId: categoryId }}
+                            visibility={ true }
```
ZakarFin, line 109: Do you need visibility prop if it's always true?

src/react/components/StyleEditor/constants.js
```diff
@@ -6,6 +6,16 @@ export const ANTD_FORMLAYOUT = {
...
 };
 
+// index passed to array corresponds straight the old value used for each fill style
+export const FILL_STYLE_FALLBACK = [
+    'thin_diagonal',
+    'thick_diagonal',
+    'thin_horizontal',
+    'thick_horizontal',
+    'transparent',
+    'solid'
```
ZakarFin, line 16: Import bundles/mapping/mapmodule/oskariStyle/constants.js and use the constants instead of repeating the string values here so it's more easy to find connections between these values.

src/react/components/VectorStyle/VectorStyleModal.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { VectorNameInput } from 'oskari-ui/components/VectorStyle';
+import { LocaleConsumer, Controller } from 'oskari-ui/util';
+import { Message, Modal } from 'oskari-ui';
+import { StyleEditor } from 'oskari-ui/components/StyleEditor';
+
+export const VectorStyleModal = LocaleConsumer(({editorState, onCancel, onModalOk, setEditorState, nameValidation, setName}) => {
```
ZakarFin, line 8: Shouldn't need to have LocaleConsumer here since Message tag is a consumer.

src/react/components/VectorStyle/VectorStyleModal.jsx
```diff
@@ -0,0 +1,39 @@
...
+            visible={ editorState.modalVisibility }
+            okButtonPros={ 'disabled' }
+            onOk={ () => onModalOk() }
+            onCancel={ onCancel }
+            cancelText={ <Message messageKey="cancel" /> }
+            okText={ <Message messageKey="save" /> }
+        >
+            <VectorNameInput
+                styleName={ editorState.styleName }
+                isValid={ nameValidation(editorState.styleName) }
```
ZakarFin, line 20: I think name validation (non-empty) can be baked in to this component and setName + nameValidation props removed. The onChange for name could be the setEditorState() with the updated name. Do you see some problem with this kind of simplification?

src/react/components/VectorStyle/index.js
```diff
@@ -0,0 +1,2 @@
+export { VectorNameInput } from './VectorNameInput';
```
ZakarFin, line 1: Only VectorStyleModal uses VectorNameInput so we could remove export for the input from here?

bundles/mapping/mapmodule/oskariStyle/constants.js
```diff
@@ -45,12 +45,12 @@ export const EFFECT = {
...
 
 export const FILL_STYLE = {
-    THIN_DIAGONAL: 0,
-    THICK_DIAGONAL: 1,
-    THIN_HORIZONTAL: 2,
-    THICK_HORIZONTAL: 3,
-    TRANSPAREN: 4,
-    SOLID: 5
+    THIN_DIAGONAL: 'thin_diagonal',
+    THICK_DIAGONAL: 'thick_diagonal',
```
ZakarFin, line 49: The change of these values creates a breaking change for current styles used over RPC API, the printing backend implementation and the current styles in the database. Changing these should be in another PR AND would require fallback for the current values that are supported.

Also I would rather have the value as diagonal/horizontal first (as the "main" style direction) and thin/thick after it (as a modifier for the style). But I'm open for discussing this in more details. The actual problem is that "everything breaks" for most styles using these if we make just this change without adding support for the current values also (now only the style editor supports these values, not the actual components that end-users see with these styles).

bundles/framework/myplaces3/MyPlacesStyleForm.jsx
kasparia: Works with Tooltip as well. Deleted this.

src/react/components/VectorStyle/VectorStyleModal.jsx
kasparia: Fixed.

bundles/framework/myplaces3/MyPlacesTab.js
kasparia: Removed this. Left it there by mistake.

src/react/components/VectorStyle/index.js
kasparia: Fixed this.

bundles/mapping/mapmodule/oskariStyle/constants.js
kasparia: Fixed these back to old one.

src/react/components/StyleEditor/constants.js
kasparia: Removed these for now.

PR #1620, opened 2021-08-19
Wfs single hover layer

okauppinen:
Add only one common hover layer for WFS layers. WFS layers registers hover style to hover service. If hovered WFS layer is changed then style and opacity is updated for common hover layer.

Also fixed hover for mvt rendered wfs layer. Had to make same solution than pure mvt layers have (copy of layer with same source + hover style and style function that renders RenderFeature only if hovered id property matches).


bundles/mapping/mapmodule/service/HoverHandler.js
```diff
@@ -21,7 +23,12 @@ export class HoverHandler {
         this._mapmodule.getMap().getViewport().addEventListener('mouseout', evt => {
             this.clearHover();
         }, false);
-        this.styleFactory = this._mapmodule.getGeomTypedStyles.bind(this._mapmodule);
+        this._styleFactory = this._mapmodule.getGeomTypedStyles.bind(this._mapmodule);
+        const olLayer = new olLayerVector({ source: new olSourceVector() });
+        olLayer.set(LAYER_HOVER, true, true);
+        olLayer.setZIndex(1); // TODO: how to handle layer ordering
```
ZakarFin, line 29: Is this required since the layer is included in the getOlLayers() array from plugins? Or is there a problem with layer ordering even with it?

PR #1622, opened 2021-08-20
Change map description flyout to React-popup

kasparia:


bundles/framework/maplegend/Flyout.js
```diff
@@ -52,9 +59,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.maplegend.Flyout',
          */
         startPlugin: function () {
             var me = this;
-            me.templateLayer = jQuery('<div class="maplegend-layer"></div>');
             me.templateTools = jQuery('<div class="maplegend-tools"><div class="layer-description"><div class="icon-info"></div></div></div>');
```
ZakarFin, line 62: Can this be removed as well?

bundles/framework/maplegend/Flyout.js
```diff
@@ -17,8 +25,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.maplegend.Flyout',
     function (instance) {
         this.instance = instance;
         this.container = null;
-        this.templateLayer = null;
-        this.templateLayerLegend = null;
+        this.sandbox = this.instance.getSandbox();
         this.state = null;
         this.templateNoLegend = jQuery('<div class="no-maplegend"></div>');
```
ZakarFin, line 30: Can `templateNoLegend` be removed as well?

bundles/framework/maplegend/Flyout.js
```diff
@@ -100,175 +105,47 @@ Oskari.clazz.define('Oskari.mapframework.bundle.maplegend.Flyout',
...
-                    layerTitle.find('div.icon-info').on('click', function (event) {
-                        event.stopPropagation();
-                        sandbox.postRequestByName('catalogue.ShowMetadataRequest', [{
-                            uuid: uuid
-                        }]);
+                const layerLegendImage = layer.getLegendImage ? layer.getLegendImage() : null;
+
+                if (layerLegendImage) {
+                    titles.push({
+                        title: Oskari.util.sanitize(layer.getName()),
```
ZakarFin, line 131: The title is now rendered as part of JSX without `dangerouslySetInnerHTML` so we don't need to sanitize it here.

bundles/framework/maplegend/Flyout.js
```diff
@@ -100,175 +105,47 @@ Oskari.clazz.define('Oskari.mapframework.bundle.maplegend.Flyout',
...
-                accordionPanel;
-
-            for (n = layers.length - 1; n >= 0; n -= 1) {
-                layer = layers[n];
-                layerContainer = this._createLayerContainer(layer);
+            const layers = this.sandbox.findAllSelectedMapLayers().slice(0);
+            const titles = [];
 
-                const layerTitle = jQuery('<div class="maplegend-layer-title">' + Oskari.util.sanitize(layer.getName()) + '</div>');
+            for (const layer of layers) {
```
ZakarFin, line 125: You could use map to make it a bit easier to read and you could take of that slice() for layers as we don't pass it past this method:
```suggestion
const showMetadata = (event) => this.showMetadataFlyout(event, uuid);
const titles = layers
    .filter(layer => 
        typeof layer.getLegendImage == 'function' && !!layer.getLegendImage())
    .map(layer => {
        const uuid = layer.getMetadataIdentifier();
        return {
            title: layer.getName(),
            uuid: uuid,
            legendImageURL: layer.getLegendImage(),
            loadError: false,
            showMetadataCallback: uuid ? showMetadata : null;
        };
    });
```
One change above is that `showMetadataCallback` is given if there's an `uuid`. Your code uses `layerLegendImage` and I'm not sure why would that be used as uuid is used to show metadata, not the legend.

ZakarFin: Maybe we should also rename titles -> legends as the variable?

bundles/framework/maplegend/MapLegendList.jsx
```diff
@@ -0,0 +1,59 @@
+import React, { Fragment, useState } from 'react';
+import PropTypes from 'prop-types';
+import { Collapse } from 'antd';
```
ZakarFin, line 3: Is there a reason to use antd directly instead of the one from oskariui? You wouldn't have to import the style here if you import from oskariui.

bundles/framework/maplegend/Flyout.js
```diff
@@ -100,175 +105,47 @@ Oskari.clazz.define('Oskari.mapframework.bundle.maplegend.Flyout',
...
-                        subLayerlegendDiv = me._createLegendDiv(sublayer, imagesAdded);
-                        if (!subLayerlegendDiv && subLayerlegendDiv !== null) {
-                            continue;
-                        }
-                        layerDiv.append(subLayerlegendDiv);
+            ReactDOM.render(
+                <LocaleProvider value={{ bundleKey: 'maplegend' }}>
+                    { titles.length === 0
+                        ? <Message messageKey='noLegendsText' />
+                        : <MapLegendList legendList={ titles } noImageText={ <Message messageKey='invalidLegendUrl' /> } />
```
ZakarFin, line 144: I think we can bake the `noImageText` in `MapLegendList` instead of giving it as prop.

bundles/framework/maplegend/MapLegendList.jsx
```diff
@@ -0,0 +1,59 @@
...
+                { title }
+                { uuid && <MetadataIcon onClick={ (event) => callback(event, uuid) } /> }
+            </Fragment>
+        );
+    };
+
+    const composeLegendImage = (item, index) => {
+        return (
+            <img
+                onError={ () => setListState(listState => {
```
ZakarFin, line 34: Using state feels a bit of an overkill for this. Could we just add a class for the img on error that hides it and shows the message instead? Maybe it could be it's own component even under oskariui? Or atleast have the state for the img instead of the whole list?

ZakarFin: Could try something like this:
```
css:
.maplegend img.errored {
    display:none;
}
.maplegend img + span {
    display: none;
}
.maplegend img.errored + span {
    display: block;
}
jsx: 
<img src={url} onError={() => this.classList.add("errored")} />
<Message messageKey='invalidLegendUrl' />
```
1) Message is wrapped as span by default
2) By default a span following the img is hidden
3) If the img triggers onError it gets the class "errored"
4) A span followed by img.errored will be shown

It's not ideal as the message will be on the DOM either way but using this just feels a bit iffy here. Maybe it would be a better way to create the common component for images that use state internally to show another component (Message here) instead on error?

bundles/framework/maplegend/Flyout.js
kasparia: Thought it would break something. Apparently no. Deleted it.

ZakarFin: It's still in the code. GitHub will resolve the conversation automatically when the line is removed.

bundles/framework/maplegend/Flyout.js
kasparia: Renamed.

bundles/framework/maplegend/MapLegendList.jsx
kasparia: Fixed.

bundles/framework/maplegend/MapLegendList.jsx
kasparia: Made a component for image and added own state for it. Now we don't add the whole list into state anymore.

bundles/framework/maplegend/MapLegendList.jsx
```diff
@@ -0,0 +1,39 @@
...
+            </Fragment>
+        );
+    };
+
+    return (
+        <Collapse>
+            { legendList.length > 0 && legendList.map((item) => {
+                return (
+                    <CollapsePanel key={ item.title } header={ composeHeader(item.title, item.uuid, item.showMetadataCallback) }>
+                        <LegendImage item={ item } />
```
ZakarFin, line 29:
```suggestion
                        <LegendImage url={ item.legendImageURL } />
```

bundles/framework/maplegend/LegendImage.jsx
```diff
@@ -0,0 +1,25 @@
...
+                    onError={ () => setImageState({ hasError: true }) }
+                    src={ item.legendImageURL }
+                />
+            }
+        </Fragment>
+    );
+};
+
+LegendImage.propTypes = {
+    item: PropTypes.object.isRequired
```
ZakarFin, line 24:
```suggestion
    url: PropTypes.string.isRequired
```

bundles/framework/maplegend/LegendImage.jsx
```diff
@@ -0,0 +1,25 @@
+import React, { Fragment, useState } from 'react';
+import { Message } from 'oskari-ui';
+import PropTypes from 'prop-types';
+
+export const LegendImage = ({ item }) => {
```
ZakarFin, line 5:
```suggestion
export const LegendImage = ({ url}) => {
```

bundles/framework/maplegend/LegendImage.jsx
```diff
@@ -0,0 +1,25 @@
...
+        <Fragment>
+            { imageState.hasError
+                ? <Message messageKey='invalidLegendUrl' />
+                : <img
+                    onError={ () => setImageState({ hasError: true }) }
+                    src={ item.legendImageURL }
+                />
+            }
+        </Fragment>
+    );
```
ZakarFin, line 20:
```suggestion
    const [hasError, setError] = useState(false);
    if (hasError) {
        return (Message messageKey='invalidLegendUrl' />);
    }
    return (<img src={ url } onError={ () => setImageState(true) } />);
```

PR #1631, opened 2021-08-31
Webpack 4 -> 5 changes

ZakarFin:
The custom localization plugin that we have needs some updating but the version included works (warns about deprecation for how it handles adding assets during build). The biggest problem is that Cesium doesn't support Webpack 5 properly yet so it's not worth it spending more time on this for now:

https://github.com/CesiumGS/cesium-webpack-example/pull/23
https://github.com/CesiumGS/cesium-webpack-example/pull/25
https://github.com/CesiumGS/cesium/issues/8471


ZakarFin:
Cesium with Webpack 5: https://github.com/CesiumGS/cesium-webpack-example/pull/28
We might want to take another crack at this.

PR #1634, opened 2021-09-01
Fix overflow issue in Firefox with layerselector

kasparia:
This should fix issue where content in layerselector goes under scroll bar. Other content should stay unaffected as every flyout is rather large in size.

### DIFF
--- a/bundles/framework/divmanazer/resources/scss/divman.scss
+++ b/bundles/framework/divmanazer/resources/scss/divman.scss
@@ -317,6 +317,7 @@ body {
         border: 0;
         overflow: auto;
         max-height: 500px;
+        min-width: 580px;
     }
 }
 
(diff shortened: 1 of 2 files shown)


bundles/framework/divmanazer/resources/scss/divman.scss
```diff
@@ -317,6 +317,7 @@ body {
         border: 0;
         overflow: auto;
         max-height: 500px;
+        min-width: 580px;
```
ZakarFin, line 320: I can see this causing confusion in the future because .oskari-attached has lower min-width BUT is the parent element of .oskari-flyoutcontentcontainer. So the child element would have larger min-width than the parent. I'd say we need to find another way of solving this.

kasparia: Fixed this with other suggested approach.

ZakarFin:
Fixes oskariorg/oskari-docs#251

PR #1636, opened 2021-09-08
Fix area fill to show in preview

kasparia:


ZakarFin:
Previously thin and thick versions of fills with line patterns was the same on preview.

PR #1638, opened 2021-09-13
Fix dropdowns in statistical maps

ZakarFin:
Broken by recent AntD upgrade and addition in global reset CSS because of that.


ZakarFin:
Related #1629

PR #1639, opened 2021-09-13
Visual fixes for infobox content

ZakarFin:
Reserve space for title AFTER the popup size has been determined. Previously the popup might be wider when space for title was calculated, adaptSize() changed the popup size changed making a long title wrap to more lines than was used for calculating height.

Also fixes another small issue caused by AntD global reset for box-sizing affecting layer name boxes on GFI responses.


ZakarFin:
Related #1629

PR #1647, opened 2021-09-28
Add documentation for transient parameter in marker data object

WTurunen:
This parameter seems to be missing from the official documentation


ZakarFin:
Awesome, thanks!

PR #1649, opened 2021-09-29
Refactor vector layer styling

okauppinen:
WFS layer style generator creates styles for: default, featureStyle and optionalStyles. Each of these contains default, selected, hover, selectedHover styles. mapmodule creates dot, line and area ol styles for each style:
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/util/style.js#L198-L250

If layer doesn't have optionalStyles, mixed geometries or cluster, it can be styled by one ol style without style function and creating lot of extra ol styles. Also creating stylefunction is quite complicated because plugin, mapmodule/styleFacotry, layer handlers and styleGenerator does their own tricks.

Also layer plugin implemented styleGenerators has duplicate code and they may act differently.

This pr's main goal is to simplify and unify layer style creation :
- separate selected features styling to own handler (like hovered features)
- allow layers to use ol style without style function when it's possible (useStyleFunction(layer))
- move most style related functionality to mapmodule (generator.ol)
- use generator.ol instead of layer plugins/handlers own implementations
- cache wfs default typed styles and style function

Other changes:
- SelectedFeatureHandler sends WFSFeaturesSelectedEvent
- SelectionPlugin stores selectFromAllLayers and its added to WFSSetFilter

Maybe WFSLayerService isn't needed anymore because in many places selected layers are filtered by hasFeaturedata instead of asking selected wfs layers from layer service. Also transport doesn't need it anymore.


bundles/framework/featuredata2/event/WFSSetFilter.js
```diff
@@ -11,9 +11,10 @@ Oskari.clazz.define('Oskari.mapframework.bundle.featuredata2.event.WFSSetFilter'
      *
      */
 
-    function (geojson, filters) {
+    function (geojson, filters, allLayers) {
```
ZakarFin, line 14: This event is used like it was a request. I noticed it some time ago but didn't do anything about it then. But maybe we should do something about it/change it to request. I'll need to take a closer look to figure out a better approach.

bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileLayer.js
```diff
@@ -15,6 +16,7 @@ Oskari.clazz.define('Oskari.mapframework.mapmodule.VectorTileLayer',
         _createEmptyStyle: function () {
             return createDefaultStyle();
         },
+        getClusteringDistance () {},
```
ZakarFin, line 19: There should be some comment about why a getter with no return value is required. Or this should inherit some generic impl from a parent class or something like that. Why is an empty function required?

bundles/mapping/mapmodule/oskariStyle/generator.ol.js
```diff
@@ -1,15 +1,223 @@
...
+    return defaultStyles[layerType] || {};
+};
+
+const getFeatureStyle = (layer, extendedDef = {}) => {
+    const style = layer.getCurrentStyle();
+    const defaultDef = getDefaultStyle(layer.getLayerType());
+    return merge(defaultDef, style.getFeatureStyle(), extendedDef);
+};
+
+export const useStyleFunction = layer => {
```
ZakarFin, line 32: I'm not sure what would be a good name for this but useX() signals that some operation is performed for the current layer instead of the parameter layer is used to detect if the layer requires a function to be used for styling. 
```suggestion
export const doesLayerRequireStyleFunction = layer => {
```
I haven't looked where this is used but could the "function vs some other styling value" be detected without this?

bundles/mapping/mapmodule/oskariStyle/generator.ol.js
```diff
@@ -1,15 +1,223 @@
...
+    case 'Point':
+    case 'MultiPoint':
+        styleType = STYLE_TYPE.POINT; break;
+    case 'GeometryCollection':
+        styleType = STYLE_TYPE.COLLECTION; break;
+    }
+    return styleType;
+};
+
+export const getOlStyleForLayer = (mapmodule, layer, extendedDef) => {
```
ZakarFin, line 58: Add some documentation here. What is the extendedDef? Is layer an instance of Oskari WFSLayer or an OpenLayers layer object?

bundles/mapping/mapmodule/oskariStyle/generator.ol.js
```diff
@@ -1,15 +1,223 @@
...
+    const optional = layer.getCurrentStyle().getOptionalStyles().map(optionalDef => {
+        return {
+            filter: getOptionalStyleFilter(optionalDef),
+            typed: mapmodule.getGeomTypedStyles(merge(featureStyle, optionalDef))
+        };
+    });
+    return getStyleFunction({ typed, optional });
+};
+
+export const getStyleForGeometry = (geometry, styleTypes) => {
```
ZakarFin, line 74: Do all of these need to be exported? If not remove exports so it's easier to read what functions are helpers for this file vs used from outside.

bundles/mapping/mapmodule/oskariStyle/generator.ol.js
```diff
@@ -1,15 +1,223 @@
...
+        }
+    };
+    return styleTypes[type];
+};
+
+// Style for cluster circles
+const clusterStyleCache = {};
+const clusterStyleFunc = feature => {
+    const size = feature.get('features').length;
+    const cacheKey = [`${size}`];
```
ZakarFin, line 99: Cache key is an array? Should be able to use just as string. And size is really only 14 or 12 for radius so we could just do:
```suggestion
    const radius = size > 9 ? 14 : 12;
    const cacheKey = `${radius }`;
```

bundles/mapping/mapmodule/oskariStyle/generator.ol.js
```diff
@@ -1,15 +1,223 @@
...
+    const defaultDef = getDefaultStyle(layer.getLayerType());
+    return merge(defaultDef, style.getFeatureStyle(), extendedDef);
+};
+
+export const useStyleFunction = layer => {
+    const styleType = geometryTypeToStyleType(layer.getGeometryType());
+    const hasOptionalStyles = layer.getCurrentStyle().getOptionalStyles().length > 0;
+    const hasCluster = typeof layer.getClusteringDistance() !== 'undefined';
+    return hasOptionalStyles || hasCluster ||
+        !styleType || styleType === STYLE_TYPE.COLLECTION;
```
ZakarFin, line 37: Can we use non-function style if labels are shown for features?

bundles/mapping/mapmodule/service/SelectedFeatureHandler.js
```diff
@@ -0,0 +1,178 @@
...
+        area: {
+            effect: 'none',
+            color: '#000000',
+            width: 4
+        },
+        width: 3
+    }
+};
+
+export class SelectedFeatureHandler {
```
ZakarFin, line 18: I'd like to think it's a separate concern to track which features form a "selection" and how that selection is reflected on the map. Also I'm not sure if it's good to have the same service trying to dig up the features from OpenLayers sources/layers since it creates a tight coupling with the mapping library. I don't have a good answer how this should be handled but the PR changes both how to select features and track that selection AND also the styling for vector features so it's a lot to take in.

ZakarFin:
Did a PR for this to separate concerns of tracking selected features and how they should be shown on the map https://github.com/okauppinen/oskari/pull/1

PR #1650, opened 2021-10-01
VectorStyle update for userlayer and myplaces

okauppinen:
Create VectorStyles in setOptions to get updated style for userlayer and myplaces. 
Other changes:
- Add AbstractVectorLayer (extends AbstractLayer) and move common code from VectorTileLayer to it
- change WFSLayer and VectorTileLayer to extend AbstractVectorLayer
- Add setStyles for AbstractLayer
- Use setOptions instead of using constructor to run extended setOptions method

Alternative solution is to update using model builder:
https://github.com/oskariorg/oskari-frontend/pull/1651

### DIFF
--- a/bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileModelBuilder.js
+++ b/bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileModelBuilder.js
@@ -1,24 +1,12 @@
-import { VectorStyle } from '../../domain/VectorStyle';
 export class VectorTileModelBuilder {
     parseLayerData (layer, mapLayerJson, maplayerService) {
         const { options = {}, style } = mapLayerJson;
-        const { externalStyles = {}, hover } = options;
-
-        // normal styles are set in AbstractVectorLayer setOptions
-        // set external styles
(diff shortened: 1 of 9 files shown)


bundles/mapping/mapmodule/plugin/vectortilelayer/VectorTileModelBuilder.js
```diff
@@ -2,15 +2,10 @@ import { VectorStyle } from '../../domain/VectorStyle';
...
-            const style = new VectorStyle(name, null, 'normal', styles[name]);
-            layer.addStyle(style);
-        });
-        // Remove styles from options to be sure that VectorStyle is used
-        delete options.styles;
+        const { externalStyles = {}, hover } = options;
 
+        // normal styles are set in AbstractVectorLayer setOptions
+        // set external styles
         Object.keys(externalStyles).forEach(name => {
```
ZakarFin, line 9: We could do the same thing for these external styles: override setOptions() for VectorTileLayer and call super.setOptions() to get the non-externalstyles parsed in the layer (instead of modelbuilder) so we don't run in the same problem later when vectortilelayer styles are updated by the admin/user?

PR #1651, opened 2021-10-01
Layerservice update

okauppinen:
Move refreshLayerOnMap from admin layereditor to layer service. Use it for myplaces and userlayer update. Userlayer bundle adds organization and group to layer. Normally layer plugin's model builder does that. Also bundle mark that layer has been added by userlayer bundle (users own layer). These had to be added before layer is added (or send another events). So decided to add modifier to refreshLayerOnMap.

This is alternative solution to get VectorStyles updated on userlayer and myplaces update:
https://github.com/oskariorg/oskari-frontend/pull/1650


ZakarFin:
Let's use the version in #1650 instead.

PR #1671, opened 2021-10-28
New DataForMapLocationEvent

kuosman:
# New DataForMapLocationEvent 

Make new `DataForMapLocationEvent` which is sent when GFI results come from ajax or wfs.

Event includes: 
- clicked coordinates (`x`, `y`)
- GFI content (`content`), can be: string (for example html, xml or string without syntax), JSON object, GeoJSON object
- result type (`type`), can be: json, geojson, text
- layer id (`layerId`) from which layer results came from

## RPC 

This event is also enabled in RPC. 
RPC gets the following Object from the event:
```javascript
{
  content: "<table><tr><td>test</td></tr></table>", // can be also json object, xml string or string without html syntax
  x: 423424,
  y: 6652055,
  layerId: 1,
  type: "text" // or json or geojson
}
```

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -26,6 +26,23 @@ Event returns following Object:
 }
 </code>
 
+### [mod] [rpc] getAllLayers
+
+The RPC `getAllLayers()` function now includes layer `attributes` `data` block. The object includes optional configuration like localized names for vector feature properties etc.
+
+```javascript
+{
+    id: layerId,
+    opacity: layerOpacity,
+    visible: layerVisibility,
+    name : layerName,
+    minZoom: minZoomLevel,
+    maxZoom: maxZoomLevel,
+    config: layerAttributesDataBlock
+}
+```
+
+
 ## 2.5.0
 
 ### [add] userstyle
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -13,7 +13,7 @@ Some extra tags:
 
 ### [add] [rpc] DataForMapLocationEvent
 
-Added new DataForMapLocationEvent. Event sends GFI results.
+Added new DataForMapLocationEvent. This allows applications to get programmatic access to the content that is normally shown on the map in an `infobox` popup when the user clicks the map (GFI/wfs feature clicks).
 
 Event returns following Object:
 <code>
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -9,6 +9,23 @@ Some extra tags:
 - [rpc] tag indicates that the change affects RPC API
 - [breaking] tag indicates that the change is not backwards compatible
 
+## 2.6.0
+
+### [add] [rpc] GFIResultEvent
+
+Added new GFIREsultEvent. Event sends GFI results.
+
+Event returns following Object:
+<code>
+{
+  content: "<table><tr><td>test</td></tr></table>", // or json object
+  x: 423424,
+  y: 6652055,
+  layerId: 1,
+  type: "text" // or json or geojson
+}
+</code>
+
 ## 2.5.0
 
 ### [add] userstyle
@@ -17,7 +34,7 @@ Added initial bundle documentation. Userstyle functionality has been moved from 
 
 ### [mod] MapModulePlugin.MapLayerUpdateRequest
 
-Added documentation for the request. The request itself has been available from 1.x already but the it wasn't documented. 
+Added documentation for the request. The request itself has been available from 1.x already but the it wasn't documented.
 Since 2.5 it can also be used to force reload of features from a service on a vector/WFS-layer.
 
 ## 2.4.0
@@ -112,7 +129,7 @@ Object with optional parameters for options as default for all locations.
 Added fourth parameter "options".
 ```javascript
 { srsName, animation }
-``` 
+```
 `srsName`: The projection in which the given coordinates are
 `animation`: Animation to use on map move. Possible values: `fly`, `pan`.
 
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -11,9 +11,9 @@ Some extra tags:
 
 ## 2.6.0
 
-### [add] [rpc] GFIResultEvent
+### [add] [rpc] DataForMapLocationEvent
 
-Added new GFIREsultEvent. Event sends GFI results.
+Added new DataForMapLocationEvent. Event sends GFI results.
 
 Event returns following Object:
 <code>
--- a/api/framework/rpc/bundle.md
+++ b/api/framework/rpc/bundle.md
@@ -154,7 +154,8 @@ Returns events that are supported by rpc functionality. For example:
         "SearchResultEvent": true,
         "UserLocationEvent": true,
         "DrawingEvent": true,
-        "FeatureEvent": true
+        "FeatureEvent": true,
+        "GFIResultEvent": true
     }
 
 **getSupportedFunctions()**
--- a/api/framework/rpc/bundle.md
+++ b/api/framework/rpc/bundle.md
@@ -126,7 +126,8 @@ Returns all the layers available on map. If layer has minimum zoom level and max
         visible: layerVisibility,
         name : layerName,
         minZoom: minZoomLevel,
-        maxZoom: maxZoomLevel
+        maxZoom: maxZoomLevel,
+        config: layerAttributesDataBlock
     }
 
 **getMapPosition()**
--- a/api/framework/rpc/bundle.md
+++ b/api/framework/rpc/bundle.md
@@ -156,7 +156,7 @@ Returns events that are supported by rpc functionality. For example:
         "UserLocationEvent": true,
         "DrawingEvent": true,
         "FeatureEvent": true,
-        "GFIResultEvent": true
+        "DataForMapLocationEvent": true
     }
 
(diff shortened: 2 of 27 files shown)


kuosman:
Hi. 
I fixed this event so at also vector feature clicking sends this event when clicking on the map.
Now everything that would be shown on the inbox/map sent out this event.

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -385,6 +387,111 @@ Oskari.clazz.define(
...
+        _sendGFIResultEvent: function (data) {
+            /**
+             * Checks at if data has JSON
+             * @param {Object} data
+             * @returns {Boolean} true if data has JSON Object, other false
+             */
+            const hasJSON = (data) => {
+                try {
+                    if (typeof data === 'string') {
+                        JSON.parse(data);
```
ZakarFin, line 406: JSON.parse() doesn't throw an error with values like `2`, `'2'`, `true`, `false`, `null`.

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -385,6 +387,111 @@ Oskari.clazz.define(
...
+                if (hasJSON(feature.content) && !hasGeoJSON(feature.content)) {
+                    return 'json';
+                } else if (hasJSON(feature.content) && hasGeoJSON(feature.content)) {
+                    return 'geojson';
+                } else if (typeof feature === 'object' && !hasGeoJSON(feature)) {
+                    return 'json';
+                } else if (typeof feature === 'object' && hasGeoJSON(feature)) {
+                    return 'geojson';
+                }
+                return 'text';
```
ZakarFin, line 449: Reduce calls to parsing functions:
```suggestion
            const getGfiResponseType = (feature) => {
                const isJSON = hasJSON(feature.content);
                if (isJSON) {
                    if (hasGeoJSON(feature.content)) {
                        return 'geojson';
                    }
                    return 'json';
                }
                return 'text';
```

kuosman: How about WFS feature clicks ? Now suggested change not sends this event out when clicking WFS feature, there clicked feature attributes data come directly `feature` object. 

For example: 
```javascript
feature = { 
    'Attribute1': 'aaa',
    'Attribute2' 'bbb'
};
```

Or keep this event only the GFI-stuff that is fetched through the server ?

ZakarFin: True, are there any other exceptions of could we do something like:
```suggestion
            const isWMSgfi = (feature) => feature.type = 'wmslayer' &&
                ['presentationType', 'content', 'type'].every(key => !!feature[key]);
            const getGfiResponseType = (feature) => {                
                if (isWMSgfi(feature)) {
                    const isJSON = hasJSON(feature.content);
                    if (isJSON) {
                        if (hasGeoJSON(feature.content)) {
                            return 'geojson';
                        }
                        return 'json';
                    }
                    return 'text';
                } else {
                    // Content is WFS feature properties
                    return 'json';
                }
            }
```

ZakarFin: BTW these helper-functions should be moved out from this function so we don't need to declare them for each call. Since they can operate on just the params the receive as input they can be moved.

ZakarFin: Try reducing the JSON.parse() calls in general. Since now it's done to detect if the content is JSON. Then again to parse the content to an object etc.

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -385,6 +387,116 @@ Oskari.clazz.define(
...
+            };
+
+            /**
+             * Gets GFI content
+             * @param {Object} feature
+             * @returns GFI content
+             */
+            const getGfiContent = (feature) => {
+                if (hasJSON(feature.content) && typeof feature.content !== 'object') {
+                    return JSON.parse(feature.content);
```
ZakarFin, line 459: Here hasJSON() calls JSON.parse() only to return boolean to call JSON.parse() again on the same content.

api/mapping/mapmodule/event/GFIResultEvent.md
```diff
@@ -0,0 +1,94 @@
+# GFIResultEvent [rpc]
+
+Notifies user that map gets GFI results.
+
+## Description
+
+Event is used to nofify user about the GFI response. Event includes information about the click coordinates and content of GFI.
```
ZakarFin, line 7: Maybe add something like: "Note! This event is triggered for each layer and vector feature that is hit. You can detect if the events are from the same click by comparing coordinates"?

api/CHANGELOG.md
```diff
@@ -11,6 +11,21 @@ Some extra tags:
 
 ## 2.6.0
 
+### [add] [rpc] DataForMapLocationEvent
+
+Added new DataForMapLocationEvent. Event sends GFI results.
```
ZakarFin, line 16:
```suggestion
Added new DataForMapLocationEvent. This allows applications to get programmatic access to the content that is normally shown on the map in an `infobox` popup when the user clicks the map (GFI/wfs feature clicks).
```

api/mapping/mapmodule/event/DataForMapLocationEvent.md
```diff
@@ -0,0 +1,96 @@
+# DataForMapLocationEvent [rpc]
+
+Notifies user that map gets GFI results.
+
+## Description
+
+Event is used to nofify user about the GFI response. Event includes information about the click coordinates and content of GFI.
```
ZakarFin, line 7:
```suggestion
The event allows applications to get programmatic access to the content that is normally shown on the map in an `infobox` popup when the user clicks the map (GFI/wfs feature clicks).
```

api/mapping/mapmodule/event/DataForMapLocationEvent.md
```diff
@@ -0,0 +1,96 @@
+# DataForMapLocationEvent [rpc]
+
+Notifies user that map gets GFI results.
```
ZakarFin, line 3:
```suggestion
Notifies application that a click on the map resulted in results for that location (GFI style content).
```

api/framework/rpc/bundle.md
```diff
@@ -155,7 +155,8 @@ Returns events that are supported by rpc functionality. For example:
         "SearchResultEvent": true,
         "UserLocationEvent": true,
         "DrawingEvent": true,
-        "FeatureEvent": true
+        "FeatureEvent": true,
+        "GFIResultEvent": true
```
ZakarFin, line 159:
```suggestion
        "DataForMapLocationEvent": true
```

bundles/framework/rpc/instance.js
```diff
@@ -149,7 +149,7 @@ Oskari.clazz.define(
             if (allowedEvents === null || allowedEvents === undefined) {
                 allowedEvents = ['AfterMapMoveEvent', 'MapClickedEvent', 'AfterAddMarkerEvent', 'MarkerClickEvent',
                     'RouteResultEvent', 'FeedbackResultEvent', 'SearchResultEvent', 'UserLocationEvent', 'DrawingEvent', 'FeatureEvent', 'InfoboxActionEvent', 'InfoBox.InfoBoxEvent',
-                    'RPCUIEvent', 'map.rotated', 'MapTourEvent', 'TimeChangedEvent'];
+                    'RPCUIEvent', 'map.rotated', 'MapTourEvent', 'TimeChangedEvent', 'GFIResultEvent'];
```
ZakarFin, line 152:
```suggestion
                    'RPCUIEvent', 'map.rotated', 'MapTourEvent', 'TimeChangedEvent', 'DataForMapLocationEvent'];
```

bundles/mapping/mapmodule/plugin/getinfo/GfiHelper.js
```diff
@@ -0,0 +1,108 @@
...
+export const hasGfiData = (content, type) => {
+    const hasStringGfiData = typeof content === 'string' && content !== 'unknown' && content !== '';
+    const hasGeoJsonGfiData = type === 'geojson' && content.features && content.features.length > 0;
+    const hasJsonGfiData = type === 'json' && content !== '';
+    if (hasStringGfiData ||
+        hasGeoJsonGfiData ||
+        hasJsonGfiData) {
+        return true;
+    }
+    return false;
```
ZakarFin, line 107:
```suggestion
    return hasStringGfiData || hasGeoJsonGfiData || hasJsonGfiData;
```

tests/rpc/src/common_parameters.js
```diff
@@ -356,7 +356,8 @@ var supportedEvents = [
   "InfoboxActionEvent",
   "InfoBox.InfoBoxEvent",
   "RPCUIEvent",
-  "map.rotated"
+  "map.rotated",
+  "GFIResultEvent"
```
ZakarFin, line 360:
```suggestion
  "DataForMapLocationEvent"
```

PR #1677, opened 2021-11-05
Fix getHoverOptions is not a function error

kuosman:
Fix warning at getHoverOptions is not a function.
image


ZakarFin:
Excellent, thanks!

PR #1680, opened 2021-11-05
React-based flyout implementation

ZakarFin:
This is a bit like React.js implementation of https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/divmanazer/extension/ExtraFlyout.js in that it looks like a Flyout window but isn't tied to the events/requests that link tiles and flyouts functionally.

We probably need more than open and close functions to make this work well with the current functionalities.

Recognized TODOs:

- [x] Should "bring to front" when clicked (atleast on the header)
- [x] The flyout can't be dragged out of screen
- [x] Changing window size to smaller will reposition popups on center of the screen
- [ ] The flyout should never be bigger than available viewport/browser window (height is limited currently but it's only for starting window size/not updated when window size changes)
- [ ] If viewport is "small enough" the flyout should cover the whole viewport and dragging should be disabled
- [ ] API tuning to make this work as a replacement for current windowing functionality
- [ ] Touch events are not handled yet
 
Current usage example:
```javascript
import { showFlyout, showPopup } from 'oskari-ui/components/window';

const btn = document.createElement('button');
btn.value='Test';
document.body.appendChild(btn);

let popupController= null;
const windowClosedCallback = () => {
    // closed -> cleanup
    popupController = null;
};
let implToggle = false;
btn.addEventListener('click', () => {
    if (popupController) {
        popupController.close();
        return;
    }
    
    if (implToggle) {
        popupController = showPopup('Title', 'Content', windowClosedCallback);
    } else {
        popupController = showFlyout('Title', 'Content', windowClosedCallback);
    }
    implToggle = !implToggle;

    // content can be updated with controller
    popupController.update('New title', 'New content');
    // programmatic way of bringing the popup on top of other popups
    popupController.bringToTop();
});
```
The popupController is an object that has functions to close, update and bring the popup to top (on top of other popups created with the same functionality).

### DIFF
--- a/src/react/components/window/util.js
+++ b/src/react/components/window/util.js
@@ -40,19 +40,52 @@ export const createDraggable = (position, setPosition, elementRef) => {
     }
     const availableWidth = getAvailableWidth();
     const availableHeight = getAvailableHeight();
+    const halfWidth = width/2;
+    const halfHeight = height/2;
+    const screenTopLimit = -5;
     const onMouseMove = (event) => {
         // prevents text selection from other elements while dragging
         event.preventDefault();
+        const element = elementRef.current;
+        if (!element) {
+            return;
+        }
         position.x += event.movementX;
         position.y += event.movementY;
-        const outOfScreen = (position.x < 0 || position.x + width > availableWidth) || (position.y < 0 || position.y + height > availableHeight);
-        const element = elementRef.current;
-        if (!outOfScreen && element) {
+        const outFromLeft = position.x < -halfWidth;
+        const outFromRight = position.x + halfWidth > availableWidth;
+        const outFromUp = position.y < screenTopLimit; // the header should remain visible to make it possible to drag back on screen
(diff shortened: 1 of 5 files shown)


src/react/components/window/util.js
```diff
@@ -0,0 +1,83 @@
...
+        // prevents text selection from other elements while dragging
+        event.preventDefault();
+        position.x += event.movementX;
+        position.y += event.movementY;
+        const outOfScreen = (position.x < 0 || position.x + width > availableWidth) || (position.y < 0 || position.y + height > availableHeight);
+        const element = elementRef.current;
+        if (!outOfScreen && element) {
+            // don't make the actual move if we would move off-screen or we don't get an element to move
+            element.style.transform = `translate(${position.x}px, ${position.y}px)`;
+        }
```
ZakarFin, line 53:
```suggestion
        } else {
        // notify user with something that the draggable "has hit a wall"
        }
```

PR #1681, opened 2021-11-07
Announcements tool and plugin

osqq:
Announcements tool and plugin functionalities. 

Tool:
You can select any valid announcement made in admin-announcements (non valid ones are disabled and shows warning on the popup). Selection is done from popup and selected announcements' names are displayed in sidebar. If previously selected announcement expires, it will get unselected and wont be displayed in the plugin.

Plugin:
Displays selected announcements, expired ones are not displayed. If there's nothing to display, plugin UI is hidden.

Added RPC function to fetch Announcements.


bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,357 @@
...
+            announcements: jQuery(
+                '<div id="publisher-layout-announcements" class="tool-options">' +
+                    '<div>' +
+                        '<input type="text" name="publisher-announcements" disabled />' +
+                        '<button/>' +
+                    '</div>' +
+                '</div>'),
+            announcementsPopup: jQuery(
+                '<div>' +
+                    '<div id="publisher-announcements-inputs">' +
```
ZakarFin, line 28: These id's are only used for styling? Change them to classes instead.

bundles/admin/admin-announcements/resources/locale/sv.js
```diff
@@ -33,6 +33,15 @@ Oskari.registerLocalization(
...
                 "cancelAnnouncementConfirm": "Är du säker?"
             },
+            "tool": {
+                "buttonLabel": "Välj Aviseringar",
+                "popup": {
+                    "title": "Aviseringar",
+                    "close": "Stäng"
+                },
+                "announcementsName": "Namn",
+                "announcementsTime": "VaDatumintervalllid"
```
ZakarFin, line 43: VaDatumintervalllid looks weird with CamelCasing. Can you confirm this is correct?

bundles/admin/admin-layereditor/view/AdminLayerForm/AdditionalTabPane/SelectedTime.jsx
```diff
@@ -7,7 +7,7 @@ import { StyledFormField } from '../styled';
 export const SelectedTime = ({ layer, controller }) => {
     const value = layer.params ? layer.params.selectedTime : '';
     const { capabilities } = layer;
-    if (!Array.isArray(capabilities.times)) {
+    if (!capabilities.times) {
```
ZakarFin, line 10: This has nothing to do with announcements and the array check is needed to prevent js error. Please rollback this change.

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,357 @@
...
+            service.fetchAdminAnnouncements((announcements) => {
+                me.announcements = announcements;
+                const toolPluginAnnouncementsConf = me._getToolPluginAnnouncementsConf();
+                if (toolPluginAnnouncementsConf !== null) {
+                    me.getPlugin().updateAnnouncements(toolPluginAnnouncementsConf.config.announcements);
+                    toolPluginAnnouncementsConf.config.announcements.forEach(announcement => {
+                        const filteredAnnouncement = me.announcements.filter(ann => ann.id === announcement);
+                        if (me.isAnnouncementValid(filteredAnnouncement[0])) {
+                            me.annTitles.push(filteredAnnouncement[0].title);
+                            me.selectedAnnouncements.push(filteredAnnouncement[0]);
```
ZakarFin, line 77: Is there a reason why the same data is in two different variables? I can see why there would be an "all announcements" and "selected announcements" that is a subset of all. But you have me.selectedAnnouncements + me.annTitles that is titles for the selecteded announcements and  me.announcements for the full data set.

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,357 @@
...
+                const annTime = announcement.begin_date.replace(/-/g, '/') + ' - ' + announcement.end_date.replace(/-/g, '/');
+
+                announcementInput.find('input[type=checkbox]').attr({
+                    'id': announcement.id,
+                    'name': 'announcement',
+                    'value': announcement.title
+                });
+                announcementInput.find('label').html(annName).attr({
+                    'for': announcement.id
+                });
```
ZakarFin, line 181: If you must use id's, prefix them with something to make it less likely they will conflict with id's on the rest of the page: 

```suggestion
const idPrefix = 'oskari_announcement_select_';
                announcementInput.find('input[type=checkbox]').attr({
                    'id': idPrefix  + announcement.id,
                    'name': 'announcement',
                    'value': announcement.title
                });
                announcementInput.find('label').html(annName).attr({
                    'for': idPrefix  + announcement.id
                });
```

Also since "name" is always "announcement" you should move it to the template.

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,357 @@
...
+                } else {
+                    content.find('div#ann-time').append('<div>' + annTime + '<div class="icon-warning-light"></div></div>');
+                    announcementInput.find('input[type=checkbox]').prop('disabled', true);
+                    content.find('div#ann-title').append(announcementInput);
+                }
+            });
+
+            // WHAT TO DO WHEN ANNOUNCEMENTS ARE SELECTED
+            content.find('input[name=announcement]').on('change', function () {
+                var announcement = me.announcements.find(item => item.title === jQuery(this).val());
```
ZakarFin, line 198: This means that if you have two announcements with the same title like "Maintenance Break" both will be selected or removed. You have an id for pretty much every element in the DOM added by this functionality, why not refer to the id here?

Using id's is problematic and I usually request them to be removed in general. With a conflicting id somewhere else on the page (might be a totally different functionality) will break this this code/make it work unexpectedly. These problems are hard to debug once you hit them (usually years later and in a functionality that has nothing to do with announcements). Please consider refactoring the code removing id's from the DOM elements and finding a different way for referencing the elements (by searching them under a specific parent elements and using classes etc).

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,357 @@
...
+                    });
+                    me.annTitles = me.annTitles.filter(function (e) { return e !== announcement.title; });
+
+                    me.getPlugin().updateAnnouncements(me.selectedAnnouncements);
+                } else {
+                    me.selectedAnnouncements.push(announcement);
+                    me.annTitles.push(announcement.title);
+
+                    me.getPlugin().updateAnnouncements(me.selectedAnnouncements);
+                }
```
ZakarFin, line 212:
```suggestion
                } else {
                    me.selectedAnnouncements.push(announcement);
                    me.annTitles.push(announcement.title);

                }
               me.getPlugin().updateAnnouncements(me.selectedAnnouncements);
```

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,357 @@
...
+                jQuery('div.basic_publisher').find('input[name=publisher-announcements]').val(me.annTitles.toString());
```
ZakarFin, line 213: What does this line do? It updates a disabled input that is shown to the user on the published panel to show titles of the selected announcements?

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,357 @@
...
+            });
+
+            popup.show(title, content, [closeButton]);
+            me.announcementsPopup = popup;
+            me.isAnnouncementsDialogOpen = true;
```
ZakarFin, line 218: Since you already have reference to the popup here you just set it to null/undefined when the popup is closed. That way you could get rid of `isAnnouncementsDialogOpen`.

bundles/framework/announcements/event/AnnouncementsChangedEvent.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @class Oskari.framework.announcements.event.AnnouncementsChangedEvent
+ *
+ * Used to notify announcements plugin that the announcements have changed.
+ */
+Oskari.clazz.define('Oskari.framework.announcements.event.AnnouncementsChangedEvent',
```
ZakarFin, line 6: This is not used anywhere and can be removed?

bundles/framework/announcements/instance.js
```diff
@@ -18,8 +18,47 @@ Oskari.clazz.define('Oskari.framework.bundle.announcements.AnnouncementsBundleIn
...
+            });
+        },
+
+        /**
+         * @method update
+         * implements BundleInstance protocol update method - does nothing atm
+         */
+        stop: function () {
+            this.sandbox = null;
+            jQuery('div#announcements-popup').remove();
```
ZakarFin, line 59: Is the popup.remove() needed here? What popup is it? Could you save a reference to it and use the ref to close it instead of removing an element with that id from the DOM?

osqq: No, seem to have missed few things while doing the last cleanups!

bundles/framework/announcements/plugin/AnnouncementsPlugin.js
```diff
@@ -0,0 +1,242 @@
...
+
+            var announcements = [];
+
+            for (const i of announcementsIds) {
+                for (const j of me.allAnnouncements) {
+                    if (j.id === i) {
+                        announcements.push(j);
+                    }
+                }
+            }
```
ZakarFin, line 162:
```suggestion
            const announcements = this.allAnnouncements.filter(ann => announcementsIds.includes(ann.id));
```

bundles/framework/announcements/plugin/AnnouncementsPlugin.js
```diff
@@ -0,0 +1,242 @@
...
+
+            if (!me.open) {
+                delete me.announcementsContent;
+            } else {
+                jQuery('div.announcements-content').children().remove();
```
ZakarFin, line 167: This seems dangerous. Find a reference to the plugin UI instead of searching through the whole DOM.

bundles/framework/announcements/plugin/AnnouncementsPlugin.js
```diff
@@ -0,0 +1,242 @@
...
+                }
+
+                var announcementsDiv = me.announcementsContent.find('div.announcements-content'),
+                    div = me.templates.announcement.clone();
+
+                const annTime = announcement.begin_date.replace(/-/g, '/') + ' - ' + announcement.end_date.replace(/-/g, '/');
+
+                div.find('button').append(announcement.title);
+                div.find('div#announcement-description').append(announcement.content);
+                div.find('div#announcement-time').append(annTime);
```
ZakarFin, line 192: This will generate errors for conflicting ids on the dev console when there are more than one announcement since you repeat the same id for all. Use classes instead.

bundles/framework/announcements/plugin/AnnouncementsPlugin.js
```diff
@@ -0,0 +1,242 @@
...
+        /**
+         * @method updateAnnouncements
+         * Updates announcements to plugin config
+         */
+        updateAnnouncements: function (announcements) {
+            var me = this;
+            var annIds = [];
+            for (const i of announcements) {
+                i.id === undefined ? annIds.push(i) : annIds.push(i.id);
+            }
```
ZakarFin, line 210:
```suggestion
            var annIds = announcements.map(i => i.id || i);
```

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
osqq: Yes, just displays the selected announcements on the tool sidebar

bundles/framework/publisher2/resources/scss/style.scss
```diff
@@ -37,11 +37,34 @@ div.notLoggedIn {
     padding: 10px;
 }
 
+div#publisher-announcements-inputs {
```
ZakarFin, line 40: You should have these in the admin-announcements scss, not under `publisher2`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oskari-frontend",
-  "version": "2.6.0-dev",
+  "version": "2.5.0",
```
ZakarFin, line 3: Don't change the version.

packages/mapping/olcs/mapmodule/bundle.js
```diff
@@ -312,6 +311,12 @@ Oskari.clazz.define(
                     'src': '../../../../bundles/mapping/mapmodule/resources/scss/mylocation.scss'
                 },
 
+                // Announcements plugin
+                {
+                    'type': 'text/javascript',
+                    'src': '../../../../bundles/mapping/mapmodule/plugin/announcements/AnnouncementsPlugin.js'
+                },
+
```
ZakarFin, line 319: This is part of the announcements bundle. Don't link it here.

bundles/mapping/mapmodule/resources/locale/fi.js
```diff
@@ -27,6 +27,9 @@ Oskari.registerLocalization(
             "defaultTitle" : "Oletustyyli"
         },
         "plugin": {
+            "AnnouncementsPlugin": {
+                "title": "Ilmoitukset"
+            },
```
ZakarFin, line 32: Is this really needed here in mapmodule when it's not part of mapmodule?

bundles/framework/announcements/service/AnnouncementsService.js
```diff
@@ -0,0 +1,73 @@
...
+                }
+            });
+        },
+
+        /**
+        * @method fetchAnnouncements
+        *
+        * Makes an ajax call to get announcements (all except expired ones)
+        */
+        fetchAnnouncements: function (callback) {
```
ZakarFin, line 55: You have the same code here: https://github.com/oskariorg/oskari-frontend/blob/2.5.1/bundles/framework/announcements/view/AnnouncementsHandler.js#L21

Refactor so the fetch is only in one place (probably this service?).

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
ZakarFin: Maybe add a comment above the line like "Shows user the currently selected announcement titles on the next to the tool". Could also show these with ul/li tags so it would be obvious that this isn't a functional input but an informative one.

osqq:
All comments have been addressed. Only problem seems to be that using service with AnnouncementsHandler is problematic as the service seems to take longer to register than the Announcement bundle takes to initialize. Service is initially undefined so using it causes Announcement bundle to not work. Any recommended workaround?

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -0,0 +1,349 @@
...
+                        if (toolPluginAnnouncementsConf.config.announcements[i] === announcement.id) {
+                            return true;
+                        }
+                    }
+                } else {
+                    return false;
+                }
+            } else {
+                return false;
+            }
```
ZakarFin, line 248: Something like this would make it more readable:
```suggestion
            if (!toolPluginAnnouncementsConf || !toolPluginAnnouncementsConf.config) {
                return false;
            }
            const announcementSelection = toolPluginAnnouncementsConf.config.announcements;
            if (!Array.isArray(announcementSelection)) {
                return false;
            }
            return announcementSelection.includes(announcement.id)
```

bundles/framework/announcements/instance.js
```diff
@@ -18,8 +18,46 @@ Oskari.clazz.define('Oskari.framework.bundle.announcements.AnnouncementsBundleIn
...
 
         afterStart: function () {
-        }
+            var me = this;
+            if (me.started) {
+                return;
+            }
+            me.started = true;
+            me.sandbox = Oskari.getSandbox();
+            me.sandbox.register(me);
```
ZakarFin, line 27: I'm pretty sure this is already done on DefaultExtension.start() and could be removed here. You should be able to reference sandbox with just this.getSandbox() at this point.

ZakarFin:
> All comments have been addressed. Only problem seems to be that using service with AnnouncementsHandler is problematic as the service seems to take longer to register than the Announcement bundle takes to initialize. Service is initially undefined so using it causes Announcement bundle to not work. Any recommended workaround?

If announcements is started before admin-announcements like it probably should this shouldn't be an issue. So check the bundle order in which they are started in. If you need to start admin before the regular announcements then you could listen to that Oskari app.start event where you register rpc-functions and try getting reference to the service from sandbox at that point. If the order is not the issue I can take another look at what might be the problem.

osqq:
The service is in Announcements, registered in the instance. The bundle import order is correct and in the bundle.js the service is listed first, then instance, then flyout. Could it be because 'afterStart' is used instead of 'start'? Then again using 'start' caused other issues previously if I'm not mistaken. The service does work elsewhere in Announcements, just not in the initial fetch in AnnouncementsHandler, so it's not a major issue but having a separate fetch ajax function in the handler instead of using the service like elsewhere seems a bit silly.

ZakarFin:
It's a separate case when they are imported and in which order they are started in. So if admin bundle is started before announcements you could run in this kind of issue.

bundles/framework/publisher2/resources/scss/style.scss
```diff
@@ -37,11 +37,34 @@ div.notLoggedIn {
...
+    display: grid;
+    grid-column-gap: 50px;
+    grid-template-columns: auto auto;
+
+    .icon-warning-light {
+        float: right;
+        margin-left: 12px;
+    }
+}
+
```
ZakarFin, line 50: Move these scss-definitions to admin-announcements where the announcement tool is defined.

bundles/framework/publisher2/resources/scss/style.scss
```diff
@@ -37,11 +37,34 @@ div.notLoggedIn {
...
+        padding-left: 28px;
+      
+        input[name=publisher-announcements] {
+            width: 125px;
+    
+            button {
+                float: right;
+            }
+        }
+    }
```
ZakarFin, line 67: Move these scss-definitions to admin-announcements where the announcement tool is defined.

ZakarFin:
Thanks!

PR #1689, opened 2021-11-15
Handle global legend when layer has no styles

ZakarFin:


ZakarFin:
Looks like server does this already so this shouldn't be needed. Server also produces a proper proxy url for it.

ZakarFin:
Instead we need to accept a style without a name but having a legend (this is the "global override") -> see second commit.

PR #1697, opened 2021-11-26
Fix bug #252

WTurunen:
This PR relates to https://github.com/oskariorg/oskari-docs/issues/252. 

It fixes the setEnabled method in MapLegend.js which seemed to cause the issue.


ZakarFin:
Is this needed? Was there a problem on the geoportal publisher side also? Or just that the sample-application didn't package the bundle for embedded maps (oskariorg/sample-application#19)?

WTurunen:
Yes, this fixes the problem of the flickering maplegend tool.

The sample application didn't have the maplegend bundle, so I added it to make it work there also.

ZakarFin:
Perfect, thanks!

PR #1709, opened 2021-12-10
GFI now use currently selected layer style

kuosman:
This PR fixes front-end from https://github.com/oskariorg/oskari-docs/issues/261

Now styles come in additional params.

### DIFF
--- a/bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
+++ b/bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
@@ -316,9 +316,9 @@ Oskari.clazz.define(
                     }
 
                     result[layer.getId()] = {
-                         ...params,
-                         STYLES: layer.getCurrentStyle().getName()
-                     };
+                        ...params,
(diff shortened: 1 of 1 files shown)


bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -311,10 +311,13 @@ Oskari.clazz.define(
...
                 .reduce((result, layer) => {
                     const params = layer.getParams();
-                    if (typeof params !== 'object' || !Object.keys(params).length) {
+                    if (typeof params !== 'object') {
                         return result;
                     }
+
+                    params.STYLES = layer.getCurrentStyle().getName();
                     result[layer.getId()] = params;
+
```
ZakarFin, line 320: Don't modify the params directly as it creates a side-effect that might very well change how the layer behaves after calling this:
```suggestion
                    result[layer.getId()] = {
                         ...params,
                         STYLES: layer.getCurrentStyle().getName()
                     };
```

ZakarFin:
Thanks!

PR #1710, opened 2021-12-14
Style editor preview button

okauppinen:
Move preview SVG to tab/format radio button selector.


src/react/components/StyleEditor/constants.js
```diff
@@ -6,6 +6,38 @@ export const ANTD_FORMLAYOUT = {
...
+    }, {
+        data: '<svg width="32" height="32"><path fill="#000000" stroke="#000000" d="m 8.969,15.99975 c 0,1.99219 0.70313,3.51563 2.10938,4.92188 1.40625,1.40625 2.92968,2.10937 4.92187,2.10937 1.99219,0 3.51563,-0.70312 4.92188,-2.10937 1.40625,-1.40625 2.10937,-2.92969 2.10937,-4.92188 0,-1.99219 -0.70312,-3.51562 -2.10937,-4.92187 C 19.51588,9.67163 17.99244,8.9685 16.00025,8.9685 c -1.99219,0 -3.51562,0.70313 -4.92187,2.10938 -1.40625,1.40625 -2.10938,2.92968 -2.10938,4.92187 0,0 0,0 0,0"/></svg>'
+    }, {
+        offsetX: 16,
+        offsetY: 5.41,
+        data: '<svg width="32" height="32"><path fill="#000000" stroke="#000000" d="m 19.280933,16.92943 c 0,0 0,-10.8984403 0,-10.8984403 0,0 -6.5625,0 -6.5625,0 0,0 0,10.8984403 0,10.8984403 0,0 -4.5703104,0 -4.5703104,0 0,0 7.8515604,8.78906 7.8515604,8.78906 0,0 7.85156,-8.78906 7.85156,-8.78906 0,0 -4.57031,0 -4.57031,0 0,0 0,0 0,0"/></svg>'
+    }
+];
```
ZakarFin, line 39: The original idea was that these could be configured for an instance. However, they were hard-coded as a copy to one of the files. If they are hard-coded we should do it in `Oskari.getMarkers()` instead. Maybe change this to a function that returns Oskari.getMarkers() or just call Oskari.getMarkers() instead of using the constant?
```suggestion
export const MARKERS = Oskari.getMarkers();
```

src/react/components/StyleEditor/Preview/area.js
```diff
@@ -63,7 +64,7 @@ export const getAreaSVG = (areaParams, areaFills) => {
    path.setAttribute('stroke-linejoin', linejoin);
 
    const patternId = 'patternPreview' + patternIdCounter++;
-   const fillPatternSVG = getPatternSVG(areaFills, pattern);
+   const fillPatternSVG = getPatternSVG(pattern);
    fillPatternSVG.setAttribute('stroke', color);
    path.setAttribute('fill', 'url(#' + patternId + ')');
```
ZakarFin, line 69: Does this work properly since the patternId isn't used for the pattern anymore? I don't remember if this was a fix for some non-chrome browser to have the id change or was it so that opening, closing and reopening the form left the pattern unfilled. But if it now works without it I think we can remove it. If it works only because we get lucky with generated patterns matching here and in getPatternSVG() we should do something about it.

okauppinen: getPatternSVG returns parsePathFromSVG(patternSVG.data) and path is passed to _composeSvgPattern. So it doesn't matter what patternId getFillOption returns. So this should work but the solution is quite messy. Maybe I will create another PR to simplify svg and pattern creation.

PR #1729, opened 2022-01-20
Oskari-ui changes for myplacesimport react form

okauppinen:
Changes for LocalizationComponent which are needed to localize name (,source and desc) for myplace and userlayer:
- placeholders can be given as input elements (child) props. Placeholders are lang suffixed (except default language). Maybe some hint could be added to ?-mark header tooltip if user doesn't use default language or yellow !-mark.
`<Input namplaceholder={Name}> => languages = [en,fi,sv] => Name, Name in Finnish, Name in Swedish`
- showDivider adds dividers between languages in collapse (like 'in Finnish').  Placeholders + multiple elements looks messy without divider (or at least some padding must be added)
- ?-icon with tooltip is added to more languages collapse header
- required can be passed with element props if input doesn't have required value. This is little ugly but in most cases value has  to be validated in parent component to disable button etc. So haven't to check existing value twice. Required contains languages which values are still required from user.
`<Input required=['en']>`

See: https://github.com/oskariorg/oskari-frontend/pull/1730 how myplaces is intended to use this

Add FileInput with clumsy styling (for myplacesimport). 

Possibility to show 'Loading...' text in Spin.

If Message is used by other bundle with LocaleProvider. Message gets localization from that bundle even using bundleKey='oskariui'. Changed to use bundleKey if provided. Not sure if it's better to wrap ui components inside LocaleProvider instead of changing Message's bundleKey handling.

### DIFF
--- a/src/react/components/FileInput.jsx
+++ b/src/react/components/FileInput.jsx
@@ -1,118 +1,150 @@
-import React from 'react';
+import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { Message, Tooltip } from 'oskari-ui';
 import { Messaging } from 'oskari-ui/util';
-import Upload from 'antd/es/upload/Upload';
 import { CloudUploadOutlined, CloseCircleOutlined  } from '@ant-design/icons';
 
 const MAX_SIZE = 10; //MB
 const MAX_COUNT = 5;
 
-const Padding = styled('div')`
-    padding: 12px 24px;
-`;
-
 const Border = styled('div')`
+    display: block;
+    width: 100%;
+    min-height: 70px;
+    padding: 12px;
     background: #fafafa;
     border: 1px dashed #d9d9d9;
-    width: 100%;
-    cursor: pointer;
     transition: border-color .3s;
-    margin-bottom: 10px;
-    display: inline-block;
     &:hover{
         border-color: #40a9ff;
     }
 `;
-const Icon = styled(CloudUploadOutlined)`
+const StyledLabel = styled('label')`
+    cursor: pointer;
+    padding: 12px;
+    &:hover{
+        color: #40a9ff;
+    }
+`;
+const HiddenFileInput = styled('input')`
+    display: none;
+`;
+const StyledFileList = styled('div')`
+    display: inline-block;
+    padding: 0px 12px;
+    margin-top: 8px;
+`;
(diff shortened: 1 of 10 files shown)


src/react/components/FileInput.jsx
```diff
@@ -0,0 +1,157 @@
...
+            return;
+        }
+        if (inputFiles.length + currentFiles.length > maxCount) {
+            const errorKey = maxCount === 1 ? 'multipleNotAllowed' : 'maxCountExceeded';
+            showError(errorKey);
+            return;
+        }
+        const validFiles = [];
+        //FileList doesn't have filter
+        inputFiles.forEach(file => validate(file) && validFiles.push(file));
```
ZakarFin, line 91:
```suggestion
        //FileList doesn't have filter
        const validFiles =  inputFiles.filter(file => validate(file));
```

src/react/components/FileInput.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+    const onDrop = event => {
+        event.preventDefault();
+        handleInputFiles(event.dataTransfer.files);
+    };
+    const handleInputFiles = inputFiles => {
+        // Special case for single option to override existing file
+        if (!multiple && inputFiles.length === 1) {
+            const file = inputFiles[0];
+            if(validate(file)){
```
ZakarFin, line 78:
```suggestion
            if (validate(file)) {
```

src/react/components/FileInput.jsx
```diff
@@ -0,0 +1,157 @@
...
+        const validFiles = [];
+        //FileList doesn't have filter
+        inputFiles.forEach(file => validate(file) && validFiles.push(file));
+        const files = [...currentFiles, ...validFiles];
+        onFiles(files);
+        setFiles(files);
+    };
+
+    const validate = file => {
+        const validType = allowedTypes.length ? allowedTypes.includes(file.type) : true;
```
ZakarFin, line 98:
```suggestion
        const validType = !allowedTypes.length || allowedTypes.includes(file.type);
```

src/react/components/FileInput.jsx
```diff
@@ -0,0 +1,157 @@
...
+        if (!validType) {
+            showError('invalidType');
+        }
+        const validSize = file.size < (maxSize * 1024 * 1024);
+        if(!validSize) {
```
ZakarFin, line 103:
```suggestion
        if (!validSize) {
```

PR #1730, opened 2022-01-20
Myplaces locale name

okauppinen:
Change myplaces to use locale object and LocalizationComponent instead of name. For new layer default name isn't used.

TODO: if mandatory is missing show note in button's tooltip => With window.showPopup it is possible to use tooltip + button but changing modal (has only buttonProps) to showpopup should be done in own PR.

AbstractLayer _name can be object but it doesn't have getter for whole object. getName returns `_name[lang || default || some]`. Maybe for userdata layers backend may return only locale (remove name) and getName is overrided to get lang || default name from locale. Will fix/unify this after locale userlayer.

Needs: https://github.com/oskariorg/oskari-frontend/pull/1729 and https://github.com/oskariorg/oskari-server/pull/793

### DIFF
--- a/bundles/framework/myplaces3/CategoryHandler.js
+++ b/bundles/framework/myplaces3/CategoryHandler.js
@@ -232,12 +232,12 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.CategoryHandler',
             dialog.show(title, message, [okBtn]);
         },
         saveCategory: function (category, callback) {
-            const id = category.categoryId;
+            const { id, locale, isDefault, style } = category;
             const data = {
                 id,
-                name: category.name,
-                isDefault: category.isDefault,
-                style: JSON.stringify(category.style)
+                isDefault,
+                locale: JSON.stringify(locale),
+                style: JSON.stringify(style)
             };
             this.instance.getService().commitCategory(data, layerJson => {
                 const isNew = !id;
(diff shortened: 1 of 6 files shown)


bundles/framework/myplaces3/MyPlacesLayerForm.jsx
```diff
@@ -1,46 +1,74 @@
...
-            />
-            { !hasName && <Message messageKey='validation.categoryName' /> }
+            <LocalizationComponent
+                placeholders={ getPlaceholders() }
+                value={ locale }
+                languages={ Oskari.getSupportedLanguages() }
+                onChange={ updateLocale }
+                mandatoryFields = { getMandatory(defaultLang) }
+            >
+                <PaddedInput type='text' name='name' validate={validate}/>
```
ZakarFin, line 59: Let's try passing placeholder with the input `<PaddedInput placeholder={'placeholder text'} useLangSuffix={true} />` instead of a prop object for LocalizationComponent. In the above example useLangSuffix=true would make LocalizationComponent attach the "in english" etc to the placeholder text value and false would leave it as is. Maybe default it to true even?

ZakarFin: Maybe we should just leave out the suffix prop for placeholder and assume it is always wanted. If we need to filter it out for some future case we can add it then.

bundles/framework/myplaces3/CategoryHandler.js
```diff
@@ -181,15 +181,21 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.CategoryHandler',
...
                 this.saveCategory({
-                    ...layer,
-                    name,
+                    id: categoryId,
+                    locale,
                     style
                 });
             };
-            showModal(layer.name, layer.style, saveLayer);
+            const names = { ...layer._name }; // TODO: getter?
```
ZakarFin, line 192: Lets add a getter for "raw" name or locale or something other than using `layer._name` directly.

PR #1733, opened 2022-01-27
Userlayer react form

okauppinen:
Add react form for import/edit userlayer. Remove Flyout, StyleForm and scss. Move ajax/fetch from userlayertab to service. Refactor instance to listen service's 'update' to refresh tab instead of using callback functions. Refactor myplacesimport to be normal bundle instance instead of default extension. This causes side effect that popup isn't closed if some tile is clicked (extension'ss flyout is opened/attached).  If edit or import fails then form isn't closed. Otherwise the new form should work as before.

Notifications is handled via Messaging and handled mostly by service. Tried to avoid using callback functions.

Also removed unused localization.

Requires: https://github.com/oskariorg/oskari-server/pull/794 and https://github.com/oskariorg/oskari-frontend/pull/1734 (import type button)


bundles/framework/myplacesimport/service/MyPlacesImportService.js
```diff
@@ -33,74 +31,223 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplacesimport.MyPlacesImportSer
...
+        }).catch(error => {
+            this.log.error(error);
+            this._showError('flyout.error.generic');
+            errorCb();
+        });
+    },
+
+    updateUserLayer: function (layerId, values, successCb, errorCb) {
+        const id = this.getActualId(layerId);
+        fetch(this.getEditLayerUrl(id, values), {
```
ZakarFin, line 96: I would just leave `id` in url and post the other values as form data instead so the URL doesn't contain the whole payload. Thoughts?

bundles/framework/myplacesimport/service/MyPlacesImportService.js
```diff
@@ -33,74 +31,223 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplacesimport.MyPlacesImportSer
...
-    getUserLayers: function (successCb, errorCb, id) {
-        const me = this;
-        let url = this.urls.get;
-
-        if (id) {
-            url = url + '&id=' + id;
-        }
-
+    getUserLayers: function (id) {
         jQuery.ajax({
```
ZakarFin, line 160: We should update this and delete to fetch as well so we don't use both jQuery.ajax() and fetch()

okauppinen: updated to fetch

bundles/framework/myplacesimport/service/MyPlacesImportService.js
okauppinen: changed to post style and locale as form data

PR #1739, opened 2022-02-04
Markers react

okauppinen:
Add react form for creating map markers. Updated style editor to handle one format case by adding preview component inside tab (implemented only for point) and removing tab selector. 

Tried to remove old stuff from plugin and clarify marker handling (add, remove, visibility) but now when I look changes it seems more like 'tolppakuti' than success (too much changes to be sure that didn't brake anything). New form's output is oskari style which can be used via mapmodule getstyle straight to feature so it didn't make sense that plugin request itself to add marker (oskaristyle -> markerdata -> request -> markerdata -> oskaristyle -> olstyle). Also storing markers(data) to four different places (_markers, _unVisibleMarkers, _markerFeatures, layer source) adds complexity. Tried to store markers only to layer source but it seemed that getting color, size and shape from feature style is complicated (needed for event and state). So stored them to _styleData (maybe them can also be added to feature properties because properties aren't used to GFI etc.). Invisible markers are handled by storing them to _hiddenMarkers (could also set null style and store ol style to visibility changes), so markers doesn't change on layer.

Added common message getters to AbstractMapModule and AbtstractMapModulePlugin to use new localization methods for mapplugins. Same way than plugin.getLocalization to avoid binding getmsg in every plugin.

Created draft pr to review that does plugin changes (gains vs. risks to break things) make any sense or should leave plugin as it was and add only oskaristyle <-> markerdata converters and use new form. At least I have to test more (state, link, add/hide/remove requests) and add junit tests. Also implement line and polygon preview in tab.

### DIFF
--- a/bundles/mapping/mapmodule/plugin/markers/MarkersPlugin.ol.js
+++ b/bundles/mapping/mapmodule/plugin/markers/MarkersPlugin.ol.js
@@ -132,15 +132,6 @@ Oskari.clazz.define('Oskari.mapframework.mapmodule.MarkersPlugin',
             this._registerTools();
         },
         /**
-         * Creates a marker layer
-         * @private
-         */
-        _createMapMarkerLayer: function () {
-            const markerLayer = new olLayerVector({ title: 'Markers', source: new olSourceVector() });
-            this.getMap().addLayer(markerLayer);
-            return markerLayer;
-        },
-        /**
          * Handles generic map click
          * @private
          * @param  {Oskari.mapframework.bundle.mapmodule.event.MapClickedEvent} event map click
@@ -180,7 +171,8 @@ Oskari.clazz.define('Oskari.mapframework.mapmodule.MarkersPlugin',
         },
         getMarkersLayer: function () {
             if (!this._layer) {
-                this._layer = this._createMapMarkerLayer();
+                this._layer = new olLayerVector({ title: 'Markers', source: new olSourceVector() });
+                this.getMap().addLayer(this._layer);
             }
             return this._layer;
         },
--- a/bundles/mapping/mapmodule/plugin/markers/MarkersPlugin.ol.js
+++ b/bundles/mapping/mapmodule/plugin/markers/MarkersPlugin.ol.js
@@ -353,32 +353,21 @@ Oskari.clazz.define('Oskari.mapframework.mapmodule.MarkersPlugin',
          * @param  {String} id  optional marker id for marker to change it's visibility, all markers visibility changed if not given.
          */
         changeMapMarkerVisibility: function (visible, id) {
-            const layerSource = this.getMarkersLayer().getSource();
-            if (id) {
-                if (visible) {
-                    const feature = this._hiddenMarkers[id];
-                    layerSource.addFeature(feature);
-                    delete this._hiddenMarkers[id];
-                } else {
-                    const feature = layerSource.getFeatureById(id);
-                    this._hiddenMarkers[id] = feature;
-                    layerSource.removeFeature(feature);
-                }
+            if (!id) {
+                this.getMarkersLayer().setVisible(visible);
                 return;
             }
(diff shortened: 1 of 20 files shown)


bundles/mapping/mapmodule/request/AddMarkerRequestHandler.js
```diff
@@ -16,7 +16,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.request.AddMarkerReque
         }
         // Else format old data to new form and inform user about this
         else {
-            this.sandbox.printWarn('AddMarkerRequest data is debricated format, formatted this to the new format before processing request. Please check your request!');
+            this._log.warn('AddMarkerRequest data is debricated format, formatted this to the new format before processing request. Please check your request!');
```
ZakarFin, line 19:
```suggestion
            this._log.warn('AddMarkerRequest data is in deprecated format. Modifying data format before processing request. Please check your request!');
```

src/react/components/StyleEditor.jsx
```diff
@@ -84,20 +84,30 @@ export const StyleEditor = ({ oskariStyle, onChange, format, tabs }) => {
...
-                </TabSelector>
+                {showSelector  && renderTab() }
                 <Card>
                     <StaticForm form={ form } onValuesChange={ onUpdate }>
-                        { selectedTab === 'point' && <PointTab oskariStyle={ style } /> }
-                        { selectedTab === 'line' && <LineTab oskariStyle={ style } /> }
-                        { selectedTab === 'area' && <AreaTab oskariStyle={  style } /> }
+                        { selectedTab === 'point' && <PointTab oskariStyle={ style } showPreview={showPreview}/> }
+                        { selectedTab === 'line' && <LineTab oskariStyle={ style } showPreview={showPreview} /> }
+                        { selectedTab === 'area' && <AreaTab oskariStyle={  style } showPreview={showPreview} /> }
```
ZakarFin, line 110: Since only Tab-component handling the prop is PointTab I would remove the prop from others.

bundles/mapping/mapmodule/plugin/markers/MarkersPlugin.ol.js
```diff
@@ -148,315 +117,182 @@ Oskari.clazz.define('Oskari.mapframework.mapmodule.MarkersPlugin',
...
-                me._waitingUserClickToAddMarker = true;
-            });
+            this.dialog.show(
+                this.getMsg('title'),
+                this.getMsg('dialog.message'),
+                [clearBtn, closeBtn]
+            );
+            this.getMapModule().getMapEl().addClass(
+                'cursor-crosshair'
+            );
```
ZakarFin, line 236: Should we use `this.getMapModule().setCursorStyle('crosshair')` instead since that's available? Not sure where `.cursor-crosshair` is defined in css but if it's on the marker plugin css we could remove that as well.

bundles/mapping/mapmodule/plugin/markers/MarkersPlugin.ol.js
```diff
@@ -148,315 +117,182 @@ Oskari.clazz.define('Oskari.mapframework.mapmodule.MarkersPlugin',
...
+                this.log.warn('Undefined coordinate in', markerData);
+                return {};
             }
             // Remove null values to get defaults
             Object.keys(markerData).forEach(key => markerData[key] === null && delete markerData[key]);
-            // Combine default values with given values
-            var data = jQuery.extend(true, {}, this._defaultData, markerData);
+
+            if (markerData.color && markerData.color.charAt(0) === '#') {
+                markerData.color = markerData.color.substring(1);
```
ZakarFin, line 273: This modifies the input parameter and might cause problems later on. Maybe use the spread operator at the start of the function to generate a copy of the input before modifying it rather than at the return statement?

PR #1741, opened 2022-02-09
UserData trimmed localization check

okauppinen:
Move getValueFromLocale to helper and trim value before check that localized value exists.


bundles/mapping/mapmodule/util/UserDataHelper.js
```diff
@@ -0,0 +1,10 @@
+export const getValueFromLocale = (locale, key, lang = Oskari.getLang()) => {
+    let values = locale[lang] || {};
+    let value = values[key];
+    if (value && value.trim()) {
+        return Oskari.util.sanitize(value);
```
ZakarFin, line 5: Sanitize doesn't play nice with React-components so we should remove the call from here and only include it on UI-parts that are not rendered with React.

PR #1751, opened 2022-02-23
Update latest versions for most dependencies

ZakarFin:
Some can't be updated without updating to Webpack 5 which at least previously didn't work with Cesium (didn't try now as it requires updates on our plugins as well). Still has some non-updated parts and requires more testing.

Note! Updating webpack-cli also requires changing ` --env.appdef=applications` to ` --env=appdef=applications` on (sample-)application `build/start` scripts.

Note2! The build process is even slower with the newer version for some reason.


ZakarFin:
Could try downgrading file-loader back to original version since looks like it's the main culprit for breaking `mapmodule.getImageUrl()` (some images for plugins are not shown with current config).

ZakarFin:
Some minor patches after major updates in:
- https://github.com/oskariorg/oskari-frontend/pull/2654
- https://github.com/oskariorg/oskari-frontend/pull/2656
- https://github.com/oskariorg/oskari-frontend/pull/2659
- https://github.com/oskariorg/oskari-frontend/pull/2406

PR #1755, opened 2022-03-02
Feature/views react dialog

Creaticide:


ZakarFin:
This is React.js implementation replacing the old jQuery version of the popup for saving/editing user views.

PR #1756, opened 2022-03-03
Feature/react account data listing

Creaticide:
Changes account data listing (AccountTab) from JQuery to React.


bundles/framework/personaldata/AccountTab.jsx
```diff
@@ -0,0 +1,43 @@
...
+            <div className="bottomlinks">
+                {changeInfoUrl && (
+                    <span><a href={changeInfoUrl}><Message messageKey='tabs.account.changeInfo' bundleKey={BUNDLE_NAME}></Message></a></span>
+                )}
+            </div>
+        </div>
+    )
+};
+
+export default AccountTab;
```
ZakarFin, line 43: We've used named exports instead of default exports in other parts of the code so for consistency this should use named export as well.

bundles/framework/personaldata/Flyout.js
```diff
@@ -141,7 +158,16 @@ Oskari.clazz.define('Oskari.mapframework.bundle.personaldata.Flyout',
                 var panel = Oskari.clazz.create('Oskari.userinterface.component.TabPanel');
                 panel.setTitle(tab.getTitle());
                 panel.setId(tabId);
-                tab.addTabContent(panel.getContainer());
+
+                if (tab.hasOwnProperty('getJsx')) {
+                    panel.setContent(ReactDOMServer.renderToString(tab.getJsx(), document.createElement('div')));
```
ZakarFin, line 163: I'm not sure what using ReactDOMServer means here, but could we do it like this:
```suggestion
                    const container= document.createElement('div');
                    panel.setContent(container);
                    ReactDOM.render(tab.getJsx(), container);
```

bundles/framework/personaldata/AccountTab.jsx
Creaticide: Export changed.

bundles/framework/personaldata/Flyout.js
Creaticide: Changed like suggested.

PR #1760, opened 2022-03-09
Statsgrid classification refactor and points svg

okauppinen:
Add react rendered points svg legend and remove jquey legend. Split legends to own files.
Refactor classification's class components to functional components.
Refactor to use antd slider for point size and colorpicker + antd select for colors.
Change classification 'name' property to 'color' and use # prefixed colors always. Requires backend migration:
Refactor classification state and props handling. Now props are more like in stateservice. For example classification values are used from activeIndicator.classification and indicators contains only selected indicators.

Refactor classification service to add needed properties/values to response instead of functions. Refactor group [regionIds] to {count, color, range, sizePx, regionIds:[]}.

Use openlayers circle style for rendering points (don't create own svg icons).

Change to use Messaging API (warning level) to show errors instead of popup.

Remove old EditClassification.js and Legend.js as them aren't used (don't know why these weren't removed when react classification was implemented).

All automatic changes to classification values should be done in one place. Now added validateClassification which updates color on mapStyle change. Maybe components should use only state service and state service should handle cache and async stuff and listen statistic services events.


bundles/statistics/statsgrid2016/service/StateService.js
```diff
@@ -91,28 +91,56 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.StateService',
...
+                this.service.getUILabels(ind, label => indicators.push({ hash, title: label.full }));
+            });
+            return {
+                activeIndicator,
+                regionset,
+                indicators,
+                seriesStats: this.service.getSeriesService().getSeriesStats(activeIndicator.hash)
+            };
+        },
+        getStateForClassifiation: function () {
```
ZakarFin, line 113:
```suggestion
        getStateForClassification: function () {
```

PR #1768, opened 2022-03-16
Styling changes for making UI-layout more manageable

ZakarFin:
The purpose is making upcoming layout changes easier and clean up some of the dead CSS/code regarding sizing/layout.

Goals:
- Use one root element for app that is assigned at startup (or currently tries to find an element with id `oskari` as root)
- If there is no root element, assign required style changes to body-tag assuming fullscreen app.
- Remove most of DOM-structure and CSS from JSP/base HTML at server and move them to client side code.
- Try keeping required changes to minimum/mostly functional as is
- Remove usage of hardcoded ids for assumed elements from code (`#mapdiv`, `#contentMap`, `#maptools` etc). Use getters from code to get references to elements.
- Remove dead code that was used for removed functionalities to calculate sizes of elements etc.

TODO/broken things:
- Mapmodule mobile <> desktop mode change is broken (starting on either mode works, changing from mode to another doesn't)
- UI during initial loading/app startup looks broken -> Maybe show a spinner instead of an empty navigation bar while starting up

Note: Current JSP/HTML changes: https://github.com/oskariorg/sample-server-extension/pull/35

The geoportal_white.jsp can be used to see what the geoportal looks like when contentMap/mapdiv is still on the base JSP. Also the menu on the white geoportal is on the right side of the page (instead of the usual left). It's mostly working like before. Differences are:
- size of the navigation bar is actually the one set on the HTML-style (170px to restrict logo from stretching it more instead of 170px that was enforced by frontend code without these changes). 
- Scrollbar shows for the whole page instead of the navigation bar if the window is too small to show the whole nav bar.


src/react/components/window/index.js
```diff
@@ -220,4 +220,19 @@ export const showBanner = (content, onClose, options = {}) => {
         close: removeWindow,
         bringToTop
     };
-}
+};
+
+export const getNavigationDimensions = () => {
+    let nav = [...Oskari.getRootEl().children].find(c => c.localName === 'nav');
+    if (!nav) {
+        return null;
```
ZakarFin, line 228: Return all zeroes instead?

bundles/framework/divmanazer/instance.js
```diff
@@ -445,6 +466,7 @@ Oskari.clazz.define('Oskari.userinterface.bundle.ui.UserInterfaceBundleInstance'
             var me = this;
             const { top, left } = this.getFlyoutDefaultPositions().attach;
             var flyoutOpts = {
+                container: jQuery('#oskari'),
```
ZakarFin, line 469:
```suggestion
                container: this.flyoutContainer,
```

bundles/framework/divmanazer/instance.js
```diff
@@ -181,11 +183,30 @@ Oskari.clazz.define('Oskari.userinterface.bundle.ui.UserInterfaceBundleInstance'
...
+                // no nav
+                return {
+                    top: 0,
+                    left: 0
+                };
+            }
+
+            let left = 0;
+            let top = 0;
+            if (dimensions.left === 0) {
```
ZakarFin, line 199: dimensions now has `placement`. Use that instead to determine left. top. right, bottom.

bundles/framework/publisher2/view/PublisherSidebar.js
```diff
@@ -72,7 +72,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PublisherSidebar
 
             me.progressSpinner.insertTo(content);
 
-            container.append(content);
+            container.prepend(content);
```
ZakarFin, line 75: Make this based on getNavigationDimensions(); placement?

bundles/framework/publisher2/instance.js
```diff
@@ -274,8 +274,11 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
          * @param {Layer[]} deniedLayers layers that the user can't publish (optional)
          */
         setPublishMode: function (blnEnabled, data, deniedLayers) {
-            var me = this;
-            var map = jQuery('#contentMap');
+            const me = this;
+            const root = jQuery(Oskari.getRootEl());
+            const navigation = root.find('nav');
+            navigation.css('display', blnEnabled ? 'none' : 'block');
+            const map = root.find('#contentMap');
```
ZakarFin, line 281: Could we get rid of #contentMap id?

ZakarFin:
Development moved to branch under oskari-frontend to make it easier to add smaller PRs to the larger work #2042

PR #1769, opened 2022-03-17
Myplaces list to react.

Creaticide:
My places list converted to React.

### DIFF
--- a/bundles/framework/myplaces3/MyPlacesList.jsx
+++ b/bundles/framework/myplaces3/MyPlacesList.jsx
@@ -1,0 +1,112 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Table } from 'oskari-ui/components/Table';
+import { Message, Confirm } from 'oskari-ui';
+import styled from 'styled-components';
+import { LOCALE_KEY } from './constants';
+import { DeleteOutlined, EditOutlined } from '@ant-design/icons'
+import { red } from '@ant-design/colors'
+
+const DELETE_ICON_STYLE = {
+    color: red.primary
+};
+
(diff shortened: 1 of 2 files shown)


bundles/framework/myplaces3/MyPlacesTab.js
```diff
@@ -96,16 +94,29 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.MyPlacesTab',
...
+                            data={places}
+                            handleDelete={(data) => this.deletePlace(data)}
+                            handleEdit={(data) => this.editPlace(data)}
+                            showPlace={(geometry, categoryId) => this.showPlace(geometry, categoryId)}
+                            getGeometryIcon={(geometry) => this.instance.getService().getDrawModeFromGeometry(geometry)}
+                        />,
+                        listContainer
+                    );
+
                     ReactDOM.render(
```
ZakarFin, line 120: Combine these instead of calling ReactDOM.render() twice.

bundles/framework/myplaces3/MyPlacesList.jsx
```diff
@@ -0,0 +1,116 @@
...
+        }
+    ];
+
+    return (
+        <StyledTable
+            columns={columnSettings}
+            dataSource={data.map(item => ({
+                key: item.id,
+                ...item
+            }))}
```
ZakarFin, line 104: Spread feature properties to table items and save original item under _place if it's needed. Then you can use the getSorterFor() for everything:
```suggestion
            dataSource={data.map(item => ({
                key: item.id,
                ...item.properties,
                _place: item
            }))}
```

PR #1775, opened 2022-03-21
Feature/personaldata table improvements

Creaticide:
Table sorting, styles, font fix, tooltips to personal data.

### DIFF
--- a/bundles/framework/personaldata/resources/scss/personaldata.scss
+++ b/bundles/framework/personaldata/resources/scss/personaldata.scss
@@ -89,8 +89,3 @@ textarea#view_description {
 .personaldata .tabsHeader {
     margin-right: 21px;
 }
-
-.table-actions {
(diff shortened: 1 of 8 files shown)


bundles/framework/personaldata/resources/scss/personaldata.scss
```diff
@@ -88,4 +88,9 @@ textarea#view_description {
...
 .personaldata .tabsHeader {
     margin-right: 21px;
-}
\ No newline at end of file
+}
+
+.table-actions {
+    display: flex;
+    justify-content: space-evenly;
+}
```
ZakarFin, line 96: Let's do this with styled-components instead.

Creaticide: Added export under oskari-ui/Table.

PR #1778, opened 2022-03-22
Statsgrid components refactor

okauppinen:
Change to use antd components (select, checkbox,..). Use styled components and remove scss. Refactor manual classification option to open dialog instead of passing component as props. Refactor legend to render both choropleth and points with react. Don't add text to point svg and use same row component (icon, range, count) for both. Add methods for creating ranges to classification service because geostats fails if some class/group doesn't get any hits.


bundles/statistics/statsgrid2016/service/ClassificationService.js
```diff
@@ -259,18 +216,48 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationService',
...
         },
-        getBoundsFallback: function (bounds, count, dataMin, dataMax) {
-            return this._tryKnownBounds(bounds, count, dataMin, dataMax) || equalSizeBands(count, dataMin, dataMax);
+
+        getValueRanges: function (data, bounds, format) {
+            const sorted = [...data].sort((a, b) => a - b);
+            const ranges = [];
+            for (let i = 0; i < bounds.length - 1; i++) {
+                const max = bounds[i + 1];
+                console.log(max);
```
ZakarFin, line 246: Remove console.log()

ZakarFin:
```suggestion
```

PR #1791, opened 2022-03-25
StatsGrid divided bounds fix

okauppinen:
Don't modify first or last bound on divided points. Fixes issue where min bound was set to base and values under base didn't hit any group (caused JS error and classification plugin crashed).

Range fixes:
Continuous: Don't modify max value if only one value is shown
Discontinuous: use set to detect if group/range has only one value (e.g. [20,20,20] => '20' NOT '20 - 20')

Add default colorset for divided points.

Hide antd sliders dot marks (small blue dots on slider line)


bundles/statistics/statsgrid2016/service/ClassificationService.js
```diff
@@ -231,17 +230,42 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.ClassificationService',
                 bounds = this.getBoundsFallback(manualBounds, count, stats.min(), stats.max());
             }
             const index = bounds.findIndex(bound => bound > base);
+            const lastIndex = bounds.length - 1;
```
ZakarFin, line 233: I tried fiddling with the naming a bit to try get this as readable as possible and came up the one below. createClamp() could be moved outside the Oskari class define/in some helper/util etc. What do you think about making these changes?

```
        /**
         * Generates a function that returns a value based on input but makes sure it is within given range.
         */
        const createClamp = (min, max) => value => {
            if (value < min) {
                return min;
            }
            if (value > max) {
                return max;
            }
            return value;
        };

        getDividedBounds: function (stats, classification) {
            const { method, count, manualBounds, base = 0 } = classification;
            let bounds;
            if (method === 'jenks') {
                bounds = stats.getClassJenks(count);
            } else if (method === 'quantile') {
                bounds = stats.getQuantile(count);
            } else if (method === 'equal') {
                bounds = stats.getEqInterval(count);
            } else if (method === 'manual') {
                bounds = this.getBoundsFallback(manualBounds, count, stats.min(), stats.max());
            }
            const potentialBaseIndex = bounds.findIndex(bound => bound > base);
            const lastIndex = bounds.length - 1;
            if (potentialBaseIndex <= 0) {
                // All bounds are under or over base
                return bounds;
            }
            // clamp function is used to ensure that we don't modify first or last bound
            const clamp = createClamp(1, lastIndex - 1);
            if (count % 2 === 0) {
                const deltaToLowerBound = base - bounds[potentialBaseIndex - 1];
                const deltaToUpperBound = bounds[potentialBaseIndex] - base;
                // by default base index is the next bound over base value
                let indexForBase = potentialBaseIndex;
                if (deltaToLowerBound < deltaToUpperBound) {
                    // but if next bound over base value is further from base than the lower bound, adjust lower instead
                    indexForBase = potentialBaseIndex - 1;
                }
                indexForBase = clamp(indexForBase);
                bounds[indexForBase] = base;
            } else {
                // odd count has 'neutral' group so both bounds have to move to base
                let under = clamp(potentialBaseIndex - 1);
                let over = clamp(potentialBaseIndex);
                if (under === over) {
                    if (over === 1) {
                        over++;
                    } else {
                        under--;
                    }
                }
                bounds[under] = base;
                bounds[over] = base;
            }
            return bounds;
        },

PR #1797, opened 2022-04-05
Feature/mydata tabs react

Creaticide:
Personal data flyout content to React.


bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { AccountTab } from './view/Account/AccountTab';
+import { Tabs } from 'antd';
+import { MyDataHandler } from './handler/MyDataHandler';
+import { MyViewsTab } from './view/MyViews//MyViewsTab';
```
ZakarFin, line 6:
```suggestion
import { MyViewsTab } from './view/MyViews/MyViewsTab';
```

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,229 @@
...
+        this.instance = instance;
+        this.container = null;
+        this.state = null;
+
+        this.template = null;
+        this.templateTabHeader = null;
+        this.templateTabContent = null;
+        this.uiHandler = new MyDataHandler(() => this.update());
+        this.publishedMapsHander = new PublishedMapsHandler(() => this.update(), this.instance);
+        this.myViewsHandler = new MyViewsHandler(() => this.update());
```
ZakarFin, line 34: Remove the listener param from these as they are added through `addTab()` that already calls `handler.addStateListener(() => this.update());`

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,229 @@
...
+                return false;
+            }); */
+
+            this.addTab(
+                'account',
+                Oskari.getMsg('PersonalData', 'tabs.account.title'),
+                AccountTab,
+                {
+                    getState: () => ({ user: Oskari.user() }),
+                    getController: () => ({ changeInfoUrl: () => Oskari.urls.getLocation('profile') }),
```
ZakarFin, line 81: We should pass both user and url in state to emphasize getters are not provided by controllers. Controller can be a no-op function here:
```suggestion
                    getState: () => ({
                        user: Oskari.user(),
                        changeInfoUrl: Oskari.urls.getLocation('profile')
                     }),
                    getController: () => null),
```

bundles/framework/mydata/Flyout.js
ZakarFin: Actually we don't really need or want references to these handlers here so we could generate them in `startPlugin()` on the `addTab()` calls?

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,229 @@
...
+                    addStateListener: () => null
+                }
+            );
+            this.addTab(
+                'myviews',
+                Oskari.getMsg('PersonalData', 'tabs.myviews.title'),
+                MyViewsTab,
+                this.myViewsHandler
```
ZakarFin, line 89: Handlers could be moved here and only passed to addTab() without saving references to `this`?

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,229 @@
...
+            );
+            this.addTab(
+                'publishedmaps',
+                Oskari.getMsg('PersonalData', 'tabs.publishedmaps.title'),
+                PublishedMapsTab,
+                this.publishedMapsHander
+            );
+
+            this.myViewsHandler.refreshViewsList();
+            this.publishedMapsHander.refreshViewsList();
```
ZakarFin, line 99: These can be moved to the contructors of the handlers.

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,229 @@
...
+                registerLink = <a href={registerUrl}>{this.instance.getLocalization('register')}</a>;
+            }
+
+            return (
+                <div>
+                    <p>
+                        {notLoggedInText}
+                    </p>
+                    <p>
+                        {loginLink}
```
ZakarFin, line 202:
```suggestion
                        {loginUrl && <a href={loginUrl}><Message messageKey={'notLoggedInText'} /></a>}
```

ZakarFin: ^ Same for the rest of these

ZakarFin: + move this to FlyoutContent.jsx instead of passing the function for it.

bundles/framework/mydata/resources/scss/personaldata.scss
```diff
@@ -0,0 +1,91 @@
+/* Account tab
```
ZakarFin, line 1: We should get rid of this scss and move the styling that is still required to the components with styled-components.

bundles/framework/mydata/events/PersonaldataLoadedEvent.js
```diff
@@ -0,0 +1,27 @@
+/**
```
ZakarFin, line 1: I think we can remove this event. Bundles can listen to Oskari.app `on('start')` or `on('app.start')` etc instead.

bundles/framework/mydata/handler/MyDataHandler.js
```diff
@@ -0,0 +1,60 @@
+import { StateHandler, controllerMixin } from 'oskari-ui/util';
+
+class UIHandler extends StateHandler {
+    constructor (consumer) {
+        super();
+        this.sandbox = Oskari.getSandbox();
+        this.viewService = Oskari.clazz.create('Oskari.mapframework.bundle.personaldata.service.ViewService', Oskari.urls.getRoute());
```
ZakarFin, line 7: ViewService isn't used and shouldn't referenced here anymore.
```suggestionOskari.urls.getRoute());
```

bundles/framework/mydata/handler/MyDataHandler.js
```diff
@@ -0,0 +1,60 @@
...
+        this.setState({
+            tabs: []
+        });
+        this.popupControls = null;
+        this.loc = Oskari.getMsg.bind(null, 'PersonalData');
```
ZakarFin, line 12: this.loc isn't used?
```suggestion
```

bundles/framework/mydata/handler/MyDataHandler.js
```diff
@@ -0,0 +1,60 @@
...
+            return;
+        }
+
+        return handler.apply(this, [e]);
+    }
+}
+
+const wrapped = controllerMixin(UIHandler, []);
+
+export { wrapped as MyDataHandler };
```
ZakarFin, line 60: Cleanup: 
```suggestion
}

export { UIHandler as MyDataHandler };
```

bundles/framework/mydata/request/AddTabRequest.js
```diff
@@ -0,0 +1,63 @@
+/**
```
ZakarFin, line 1: We should get rid of the request probably if we _don't_ try having a backwards compatible functionality that supports jQuery tabs. I don't think we should. Instead we should try to make myplaces etc external bundles pushing data to personaldata work with both the new "mydata" and the current "personaldata" bundles.

bundles/framework/mydata/service/ViewService.js
```diff
@@ -0,0 +1,245 @@
+/**
+ * @class Oskari.mapframework.bundle.personaldata.service.ViewService
+ *
+ * Requests for a search to be made with the given query and provides
+ * callbacks
+ */
+Oskari.clazz.define('Oskari.mapframework.bundle.personaldata.service.ViewService',
```
ZakarFin, line 7: These should probably be in the MyViewsHandler.js instead.

bundles/framework/mydata/view/MyViews/MyViewsTab.jsx
```diff
@@ -0,0 +1,35 @@
...
+const ButtonContainer = styled.div`
+    margin: 10px 0 10px 0;
+    display: flex;
+    justify-content: flex-end;
+`;
+
+export const MyViewsTab = ({ controller, state }) => {
+
+    return (
+        <>
```
ZakarFin, line 18: Use `<React.Fragment>` instead of `<>`. It's a convention we've used that's probably not documented anywhere.

bundles/statistics/statsgrid2016/instance.js
```diff
@@ -137,10 +142,7 @@ Oskari.clazz.define(
         _addIndicatorsTabToPersonalData: function (sandbox) {
             var reqBuilder = Oskari.requestBuilder('PersonalData.AddTabRequest');
             if (typeof reqBuilder === 'function') {
-                var tab = Oskari.clazz.create('Oskari.statistics.statsgrid.MyIndicatorsTab', this);
```
ZakarFin, line 140: Restore the old MyIndicatorsTab and make it ReactDOM.render() the same list in the tab content that the new `MyIndicatorsTab.jsx` uses so we can have a bit of backwards compatibility with personaldata and mydata bundles:

        <MyIndicatorsList
            controller={controller}
            data={state.data}
        />

bundles/framework/mydata/instance.js
```diff
@@ -0,0 +1,329 @@
...
+
+            // draw ui
+            me.createUi();
+
+            // Flyout exists and we can hook up the request handler
+            this.requestHandlers = {
+                addTabRequestHandler: Oskari.clazz.create('Oskari.mapframework.bundle.personaldata.request.AddTabRequestHandler', sandbox, this.plugins['Oskari.userinterface.Flyout'])
+            };
+            sandbox.requestHandler('PersonalData.AddTabRequest', this.requestHandlers.addTabRequestHandler);
+
```
ZakarFin, line 128: Let's remove the request from the mydata bundle. The problem with the request is that everything in a request should be serializable to JSON and the current jQuery and mydata React refs + handlers are not.

```suggestion
            // Flyout exists and we can hook up the request handler
            this.requestHandlers = {
                addTabRequestHandler: Oskari.clazz.create('Oskari.mapframework.bundle.personaldata.request.AddTabRequestHandler', sandbox, this.plugins['Oskari.userinterface.Flyout'])
            };
            sandbox.requestHandler('PersonalData.AddTabRequest', this.requestHandlers.addTabRequestHandler);

```

bundles/framework/mydata/instance.js
```diff
@@ -0,0 +1,329 @@
...
+            // Personaldata available
+            var eventBuilder = Oskari.eventBuilder('Personaldata.PersonaldataLoadedEvent');
+            var event = eventBuilder();
+            sandbox.notifyAll(event);
```
ZakarFin, line 132: The event can be removed as well
```suggestion
```

bundles/framework/mydata/instance.js
ZakarFin: Instead lets add a service that is registered to sandbox. The service should provide a way for bundles to call the `addTab()` function. Other bundles could then register their tabs with `sandbox.getService('Oskari.mydata.MydataService').addTab()`

bundles/framework/mydata/service/MyDataService.js
```diff
@@ -0,0 +1,46 @@
...
+    __name: 'MyDataService',
+    /**
+     * @method getName
+     * @return {String} service name
+     */
+    getName: function () {
+        return this.__name;
+    },
+    addTab: function (id, title, component, handler) {
+        handler.addStateListener(this._updater);
```
ZakarFin, line 37: We could remove the updater function and just propagate the event to notify uiHandler statelisteners, triggering an update call that is defined for uiHandler state changes:
```suggestion
        handler.addStateListener(this._uiHandler.notify);
```

bundles/framework/mydata/service/ViewService.js
```diff
@@ -0,0 +1,245 @@
...
+         * or error(false). IF success, second parameter has the response.
+         */
+        loadViews: function (type, callback) {
+            var me = this;
+            if (!type) {
+                type = 'USER';
+            }
+
+            jQuery.ajax({
+                url: me._viewUrl + 'action_route=GetViews',
```
ZakarFin, line 55: Lets drop the configurable base url and just use:
```suggestion
                url: Oskari.urls.getRoute('GetViews'),
```

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,202 @@
...
+            const loginUrl = Oskari.getLocalized(conf.logInUrl) || Oskari.urls.getLocation('login');
+
+            const registerUrl = Oskari.urls.getLocation('register');
+
+            return {
+                loggedIn: Oskari.user().isLoggedIn(),
+                notLoggedInText,
+                loginUrl,
+                registerUrl
+            };
```
ZakarFin, line 194:
```suggestion
            const conf = this.instance.conf || {};
            return {
                loggedIn: Oskari.user().isLoggedIn(),
                loginUrl: Oskari.getLocalized(conf.logInUrl) || Oskari.urls.getLocation('login'),
                registerUrl: Oskari.urls.getLocation('register')
            };
```

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,207 @@
...
+        },
+        /**
+         * @method getLoginUrl
+         * @returns {String}
+         */
+        getLoginStatus: function () {
+            const conf = this.instance.conf || {};
+            return {
+                loggedIn: Oskari.user().isLoggedIn(),
+                notLoggedInText: this.instance.getLocalization('notLoggedIn'),
```
ZakarFin, line 196:
```suggestion
```

bundles/framework/mydata/instance.js
```diff
@@ -0,0 +1,312 @@
...
+            me.started = true;
+
+            var conf = this.conf || {};
+            this.conf = conf;
+            var sandboxName = (conf ? conf.sandbox : null) || 'sandbox';
+            var sandbox = Oskari.getSandbox(sandboxName),
+                p;
+
+            me.sandbox = sandbox;
+            this.viewService = Oskari.clazz.create('Oskari.mapframework.bundle.personaldata.service.ViewService', Oskari.urls.getRoute());
```
ZakarFin, line 100: `PublishedMapsHandler` and `MyViewsHandler` both create their own viewService instance now. I don't think we need this here anymore but if we do we should probably use one or the other from the handlers.

ZakarFin:
It's starting look really good. Lets rename everything inside `bundles/framework/mydata` from `personaldata` -> `mydata` and if you can remove the couple of changes under `myplaces3` from this pr I think we could merge this after that.

bundles/framework/mydata/instance.js
Creaticide: Changed handlers to use the same instance with existing getViewService. Is this ok?

bundles/framework/mydata/Flyout.js
```diff
@@ -0,0 +1,209 @@
...
+     */
+
+    function (instance) {
+        this.instance = instance;
+        this.container = null;
+        this.state = null;
+
+        this.template = null;
+        this.templateTabHeader = null;
+        this.templateTabContent = null;
```
ZakarFin, line 33: These are not referenced anymore in the code and could be removed?
```suggestion
```

bundles/framework/mydata/handler/PublishedMapsHandler.js
```diff
@@ -0,0 +1,292 @@
...
+        const supported = this.isCurrentProjectionSupportedForView(data);
+        if (!supported) {
+            this.createProjectionChangeDialog(() => {
+                window.location.href = this.constructUrlWithUuid(srs, null, data);
+            });
+        }
+
+        this.setMapState(data, false, () => {
+            this.setMapState(data, true);
+        });
```
ZakarFin, line 221: Should this be in else?

```suggestion
        } else {
            this.setMapState(data, false, () => {
                this.setMapState(data, true);
            });
        }
```

bundles/framework/mydata/instance.js
```diff
@@ -0,0 +1,312 @@
...
+    /**
+     * @method create called automatically on construction
+     * @static
+     */
+
+    function () {
+        this.core = null;
+        this.sandbox = null;
+        this.started = false;
+        this.template = null;
```
ZakarFin, line 23: Remove unused variables:
```suggestion
        this.sandbox = null;
        this.started = false;
```

bundles/framework/mydata/instance.js
```diff
@@ -0,0 +1,312 @@
...
+            me.started = true;
+
+            var conf = this.conf || {};
+            this.conf = conf;
+            var sandboxName = (conf ? conf.sandbox : null) || 'sandbox';
+            var sandbox = Oskari.getSandbox(sandboxName),
+                p;
+
+            me.sandbox = sandbox;
+            this.viewService = Oskari.clazz.create('Oskari.mapframework.bundle.mydata.service.ViewService', Oskari.urls.getRoute());
```
ZakarFin, line 100: ViewService isn't using the constructor param:
```suggestion
            this.viewService = Oskari.clazz.create('Oskari.mapframework.bundle.mydata.service.ViewService');
```

bundles/framework/mydata/service/MyDataService.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * @class Oskari.mapframework.bundle.mydata.service.MyDataService
+ *
+ * Requests for a search to be made with the given query and provides
+ * callbacks
```
ZakarFin, line 5:
```suggestion
 * Provides addTab() for other bundles/functionalities to add a tab panel on mydata UI.
```

bundles/framework/mydata/service/MyDataService.js
```diff
@@ -0,0 +1,48 @@
...
```
ZakarFin, line 5:
```suggestion
```

PR #1799, opened 2022-04-12
Internal vector layer handling

okauppinen:
Remove z index from hover layer and add/move hover layer one step above hovered layer.

Add addInternalLayer method for mapmodule which sets z index for layer to keep internal layer above 'normal' layers. Mapmodule has method for ordering layers by z index which is used AfterRearrangeSelectedMapLayerEvent. 
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/mapping/mapmodule/MapModuleClass.ol.js#L971

Don't know if z indexes were used but at least didn't find current usage. Maybe it could be removed or order layers also on addLayer. Layers could be also added directly to ol map which isn't handled by mapmodule. So the order in map's layer collection/array may vary (even mapmodule addLayer sorts array by z index).
"At rendering time, the layers will be ordered, first by Z-index and then by position"

Removed bringToTop from vectorlayerplugin because it was used only after addLayer which already adds it on top.

### DIFF
--- a/bundles/mapping/mapmodule/MapModuleClass.ol.js
+++ b/bundles/mapping/mapmodule/MapModuleClass.ol.js
@@ -30,6 +30,7 @@ import './AbstractMapModule';
 import './plugin/BasicMapModulePlugin';
 
 const AbstractMapModule = Oskari.clazz.get('Oskari.mapping.mapmodule.AbstractMapModule');
+const INTERNAL_LAYER_Z_INDEX = 1;
 
 if (!window.proj4) {
     window.proj4 = proj4;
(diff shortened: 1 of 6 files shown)


bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -1138,6 +1139,19 @@ export class MapModule extends AbstractMapModule {
         }
     }
 
+    /**
+     * Adds internal layer to map. Internal layers aren't listed in selected layers and are always above those.
+     * @method bringToTop
+     * @param {ol/layer/Layer} layer ol specific!
+     */
+    addInternalLayer (layerImpl) {
```
ZakarFin, line 1147: Would something like "addOverlayLayer" be more descriptive name? It's not perfect either since there's already a concept of an "overlay" in OpenLayers but I think it could be better.

ZakarFin:
Even with this it's a matter of functionality starting order that changes markers/draw layers/runtime vector layers relative placement to each other. But this removes the need to bump them to the top every time the layer order changes since layers with z-index are always on top of "regular" layers. I think having "bringToTop" is good to have, but I'm not sure if we should clarify what it does (bringing a layer to top of regular layers with these "internal/overlay" layers still on top of that?)

PR #1800, opened 2022-04-12
Announcement bundle localization

osqq:
- Integrate localizationComponent to announcements bundle
- New announcement model, "title" is now removed and renamed as "name".
- Instead of separate title and content, these are now localized and are included in "locale" json:
{ fi: { name: "Name", content: "Content"}, en: { } ...}
- Added text to announcements flyout when there are no active announcements

* Also includes fix to announcement modales not being shown

PR #1804, opened 2022-04-20
Announcements bundle text editor

osqq:
Added Draft.js text editor for announcements bundle, replacing text area content. This will have to be merged after https://github.com/oskariorg/oskari-frontend/pull/1800

Text editor contains inline styles:

Bold
Italic
Underline
Monospace
And block styles:

H1
H2
H3
Blockquote
UL
OL
Styled content is shown in all announcement bundles' parts. All styles are specific to localization and it's stored in the database in html syntax. draft-js-export-html and html-to-draftjs are included to translate Draft.js editorstate to HTML and back as Draft.js doesn't have that functionality yet.

!HOX! There is still a problem with using + and - on the editor as it currently zooms the map in and out!


src/react/components/TextEditor/InlineStyleControls.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { StyleButton } from './index';
```
ZakarFin, line 2:
```suggestion
import { StyleButton } from './StyleButton';
```

package.json
```diff
@@ -47,6 +47,8 @@
     "core-js": "3.3.5",
     "css-loader": "3.2.0",
     "dompurify": "2.3.6",
+    "draft-js": "^0.11.7",
+    "draft-js-export-html": "^1.4.1",
```
ZakarFin, line 51:
```suggestion
    "draft-js": "0.11.7",
    "draft-js-export-html": "1.4.1",
```

src/react/components/RichTextEditor.jsx
```diff
@@ -0,0 +1,171 @@
+import React from 'react';
+import {Editor, EditorState, ContentState, RichUtils, getDefaultKeyBinding, Modifier, SelectionState} from 'draft-js';
+import htmlToDraft from 'html-to-draftjs';
+import './TextEditor/RichText.css';
+import 'draft-js/dist/Draft.css';
+
+import { InlineStyleControls, BlockStyleControls } from './TextEditor/';
+
+      class RichEditor extends React.Component {
```
ZakarFin, line 9: Intendation is funky here + could this be migrated to a function component for consistency (all the others are)?

ZakarFin: Also everything else is using named exports so lets continue doing that in here as well instead of default export.

src/react/components/TextEditor/StyleButton.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export class StyleButton extends React.Component {
```
ZakarFin, line 3: Use funtion components for consistency. Everything else under src/react/components are functional components.

osqq:
I still need help with figuring out how to stop the zooming of the map when trying to write '+' or '-' to the text editor.

ZakarFin:
> I still need help with figuring out how to stop the zooming of the map when trying to write '+' or '-' to the text editor.

I think the proper fix would be to change this in OpenLayers https://github.com/openlayers/openlayers/blob/v6.14.1/src/ol/events/condition.js#L253 by adding 
`&& !originalEvent.target.isContentEditable`.

bundles/admin/admin-announcements/view/AnnouncementsForm.jsx
```diff
@@ -38,6 +40,11 @@ const AnnouncementsForm = ({controller, key, announcement, bundleKey, index}) =>
     // Should format date value before submit.
     const rangeValue = fieldsValue["range_picker"];
     let content = {...locales};
+    for (const lang in content) {
+      if (content[lang].content && content[lang].content.editorstate) {
+        content[lang].content = stateToHTML(content[lang].content.editorstate.getCurrentContent());
```
ZakarFin, line 45: Could we move `stateToHTML()` handling under `RichEditor`? So we would always get HTML and could input HTML to the editor without needing to know how it goes about it's internal state? Would be a lot nicer to use and we could change the editor implementation easier that way without touching other parts of the code.

ZakarFin: I think this is good for merging after this change.

PR #1807, opened 2022-04-27
Feature/mydata date fields

Creaticide:
Date fields to mydata tables.

### DIFF
--- a/bundles/framework/myplaces3/service/MyPlacesService.js
+++ b/bundles/framework/myplaces3/service/MyPlacesService.js
@@ -391,8 +391,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.service.MyPlacesServic
             place.setLink(Oskari.util.sanitize(properties.link));
             place.setImageLink(Oskari.util.sanitize(properties.image_url));
             place.setCategoryId(properties.category_id);
-            place.setCreateDate(properties.created, false);
-            place.setUpdateDate(properties.updated, false);
+            place.setCreateDate(properties.created);
+            place.setUpdateDate(properties.updated);
             place.setGeometry(geometry);
             const drawMode = this.getDrawModeFromGeometry(geometry);
             const measurement = this.mapmodule.formatMeasurementResult(this.mapmodule.getMeasurementResult(geometry), drawMode);
--- a/bundles/framework/myplaces3/service/MyPlacesService.js
+++ b/bundles/framework/myplaces3/service/MyPlacesService.js
@@ -391,8 +391,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.service.MyPlacesServic
             place.setLink(Oskari.util.sanitize(properties.link));
             place.setImageLink(Oskari.util.sanitize(properties.image_url));
             place.setCategoryId(properties.category_id);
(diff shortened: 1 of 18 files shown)


bundles/framework/myplaces3/service/MyPlacesService.js
```diff
@@ -406,8 +391,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.service.MyPlacesServic
             place.setLink(Oskari.util.sanitize(properties.link));
             place.setImageLink(Oskari.util.sanitize(properties.image_url));
             place.setCategoryId(properties.category_id);
-            place.setCreateDate(this._formatDate(properties.created));
-            place.setUpdateDate(this._formatDate(properties.updated));
+            place.setCreateDate(Oskari.util.formatDate(properties.created, false));
```
ZakarFin, line 394: I think we should have the date as `Date` in the model and only format when showing it (in the table renderer).

bundles/framework/myplacesimport/UserLayersList.jsx
```diff
@@ -46,6 +47,12 @@ export const UserLayersList = ({ data = [], controller, loading }) => {
             dataIndex: 'description',
             sorter: getSorterFor('description')
         },
+        {
+            align: 'left',
+            title: <Message messageKey='tab.grid.createDate' />,
+            dataIndex: 'created',
+            sorter: getSorterFor('created')
```
ZakarFin, line 54: We should use the render function for dates on the table instead of modifying the incoming list. If we ever do paging this reduces the amount of mapping we need to do and should make sorting more predictable (though `getSorterFor()` might not work with dates so we might need to change that).
```suggestion
            sorter: getSorterFor('created'),
            render: text => Oskari.util.formatDate(text)
```

bundles/framework/myplacesimport/UserLayersList.jsx
```diff
@@ -90,7 +97,8 @@ export const UserLayersList = ({ data = [], controller, loading }) => {
                 key: item.getId(),
                 name: Oskari.util.sanitize(item.getName()),
                 description: Oskari.util.sanitize(item.getDescription()),
-                source: Oskari.util.sanitize(item.getSource())
+                source: Oskari.util.sanitize(item.getSource()),
+                created: Oskari.util.formatDate(item.getCreated())
```
ZakarFin, line 101: See comment about render function:
```suggestion
                created: item.getCreated()
```

bundles/framework/myplaces3/service/MyPlacesService.js
```diff
@@ -406,8 +391,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplaces3.service.MyPlacesServic
             place.setLink(Oskari.util.sanitize(properties.link));
             place.setImageLink(Oskari.util.sanitize(properties.image_url));
             place.setCategoryId(properties.category_id);
-            place.setCreateDate(this._formatDate(properties.created));
-            place.setUpdateDate(this._formatDate(properties.updated));
+            place.setCreateDate(properties.created, false);
```
ZakarFin, line 394: There's no boolean value for the setters: https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/framework/myplaces3/model/MyPlace.js#L162

PR #1808, opened 2022-04-28
Feature/mydata react improvements

Creaticide:
Table improvements.


bundles/framework/myplaces3/MyPlacesLayerControls.jsx
```diff
@@ -2,26 +2,26 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Message, Tooltip } from 'oskari-ui';
 import styled from 'styled-components';
-import { Select } from 'antd';
+import { Select, Button } from 'antd';
```
ZakarFin, line 5: Import Select and Button from oskari-ui instead of directly from antd.

PR #1820, opened 2022-05-18
Announcements popup

okauppinen:
Refactor announcement to use new json model (endData, beginDate, title, link, options). 
Refactor to render oskari popup instead of antd modal.
Use pagination if more than one announcement.


bundles/framework/announcements/service/util.js
```diff
@@ -0,0 +1,27 @@
+const RANGE_SEPARATOR = '\u2013';
+const TIME_OPTIONS = {
+    hour: '2-digit',
+    minute: '2-digit'
+};
+
+export const formatDate = (isoDateTime) => {
```
ZakarFin, line 7: Should we add this for Oskari.util.formatDate() instead of https://github.com/oskariorg/oskari-frontend/blob/1441ad3158b3780de05c171971a9d329ae0aa080/src/util.js#L775-L799

bundles/framework/announcements/view/AnnouncementsContent.jsx
```diff
@@ -0,0 +1,36 @@
...
+import PropTypes from 'prop-types';
+import { Message } from 'oskari-ui';
+import styled from 'styled-components';
+import { SelectOutlined } from '@ant-design/icons';
+
+const LinkIcon = styled(SelectOutlined)`
+    margin-left: 6px;
+`;
+
+const Link = ({ url }) => (
```
ZakarFin, line 11: Should we add this as Link-component under oskari-ui? With something like "external" boolean prop (default to true) to set target="_blank"?

PR #1822, opened 2022-05-19
Fix indicator dataform parsing when importing data from clipboard

willberthos:
Removed the comma from the regex parser so that the data is now separated with just a semicolon or a tabulator.
Added an integer check for checking the value of an area. Non numeral values will now be shown as an empty value in the indicator form.


bundles/statistics/statsgrid2016/components/IndicatorDataForm.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-var */
```
ZakarFin, line 1: Don't disable it when you can change vars to const/let here.

PR #1827, opened 2022-05-25
Admin announcements tools

okauppinen:
Add admin tools via announcements service to flyout. Remove handler, tile and flyout from admin. Add tooled flyout content and collapse tools.

Store announcements to service and fetch from backend only once or if forced.

Filter all announcements to announcements (active), outdated and upcoming in handler.

Copied tooling from layerlist. Not sure if same tool `Oskari.mapframework.domain.Tool'` could be used. So created similar Tool under admin-announcements. Tools in layerlist handle clicks only. So didn't find any nice solution for DeleteButton + onConfirm where id could be used. Now collapse tools has hard coded DeleteButton.


bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -1,91 +1,72 @@
...
-                });
-            }
-        }.bind(this));
-    }
+    onFetch () {
+        const allAnnouncements = this.service.getAnnouncements();
+        const dontShowAgain = this.service.getIdsFromLocalStorage();
+
+        // Admin gets all announcements
+        const announcements = allAnnouncements.filter(a => isActive(a));
```
ZakarFin, line 19: I would rename some variables here to make it more clear for reading but it's no biggie (I had to scroll up to see if announcements was everything or the active ones :)):
- announcements -> activeAnnouncements
- allAnnouncements -> announcements

bundles/framework/announcements/view/AnnouncementsHandler.js
```diff
@@ -1,91 +1,72 @@
...
+        if (!this.state.popupShown) {
+            const ann = announcements.filter(ann => ann.options.showAsPopup && !dontShowAgain.includes(ann.id));
+            newState.showAsPopup = ann;
+            newState.popupShown = !ann.length;
+        }
+        if (!this.state.bannerShown) {
+            const ann = announcements.filter(ann => !ann.options.showAsPopup && !dontShowAgain.includes(ann.id));
+            newState.showAsBanner = ann;
+            newState.bannerShown = !ann.length;
+        }
```
ZakarFin, line 37: I would not include the boolean that can be easily checked from the array lenght, but I'm not sure if there is an alternative use for it? The fetch is only done once so the if will always be true?

I would rename `showAsPopup` as `popupAnnouncements` and `showAsBanner` `bannerAnnouncements`
```suggestion
        if (!this.state.popupShown) {
            const ann = announcements.filter(ann => ann.options.showAsPopup && !dontShowAgain.includes(ann.id));
            newState.popupAnnouncements= ann;
            newState.popupShown = !ann.length;
        }
        if (!this.state.bannerShown) {
            const ann = announcements.filter(ann => !ann.options.showAsPopup && !dontShowAgain.includes(ann.id));
            newState.bannerAnnouncements= ann;
            newState.bannerShown = !ann.length;
        }
```

ZakarFin:
The tools handling feels a bit heavy for what it is, but at least it's now as extendable as with layers :) Just minor renamings that for the announcements but everything looks ok for merging.

bundles/framework/announcements/view/AnnouncementsHandler.js
okauppinen: announcements -> active (like outdated and upcoming)

bundles/framework/announcements/view/AnnouncementsHandler.js
okauppinen: For admin users fetch is used after add and update.

okauppinen: Removed boolean. If state doesn't have array populates array on fetch (one time only) and empty array means that popup is shown and closed.

bundles/admin/admin-announcements/instance.js
```diff
@@ -1,42 +1,94 @@
...
-        'extend': ['Oskari.userinterface.extension.DefaultExtension']
+        setupAdminTooling () {
+            if (!this.service) {
+                return;
+            }
+            const editTool = new Tool('announcements-edit', EditOutlined);
+            editTool.setTooltip(this.loc('tools.edit'));
+            editTool.setTypes(['announcement']);
+            editTool.setCallback((id) => this.showEditPopup(id));
+            this.service.addTool(editTool);
```
ZakarFin, line 79: The Tool class should be provided by announcement bundle since it's the one that uses it/needs to understand it's usage. I would refactor this to just admin bundle service.addTool('edit', tooltip, ['announcement'], callback). Then we wouldn't have to import the tool class here.

ZakarFin:
Lets merge this so we get the service API changes to develop so we can make progress on other things. The discussed changes will be fixed in another PR.

PR #1828, opened 2022-05-25
React popup cleanup

okauppinen:
Clean open/active popups and flyouts on UIChangeEvent. Added validation for options to be sure that popup/flyout has unique id.

Not sure how event listener should be registered for this. Tested with 'Try this out in console!'.


src/sandbox/sandbox-state-methods.js
```diff
@@ -123,6 +123,8 @@ Oskari.clazz.category('Oskari.Sandbox', 'state-methods', {
             // reset to the default state
             bundle.setState(bundleState);
         }
+        // trigger an event letting bundles know that the whole UI has changed to clean functionalities
+        this.notifyAll(Oskari.eventBuilder('UIChangeEvent')('sandbox'));
```
ZakarFin, line 127: This should be sent before calling setState() on bundles since it's mainly for cleanup. This way if setState() causes a popup to open it's not cleared right away because of this.

PR #1831, opened 2022-06-02
Proxy url contains already id parameter

kuosman:
Remove unnecessary ID parameter.

* Proxy url already contains id parameter --> not necessary
* Normal non proxy WMS url not know what to do with id parameter


ZakarFin:
This resulted in the url having both id (provided by server in url) and ID (added by frontend) parameters in the url. This removes the ID and assumes server provides the id.

PR #1832, opened 2022-06-02
Myplaces refactor

okauppinen:
Use react popup instead of jQuery popup for PlaceForm. Changed PlaceForm to use LabeledInputs and URLInput and added button for adding category/layer.

Not sure if URLInput looks nice in form but at least it handles protocol more properly than:
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/myplaces3/view/PlaceForm.js#L179-L181

Use antd confirm and popover (if category has places) for delete category.

MyPlacesHandler, CategoryHandler and MyPlacesService handled things with callbacks. Now all content is react rendered so there is no need for callbacks. Only category delete with places really needs callback. Removed callbacks and CategoryHandler and added helpers for places and layers.

For now:
- only ButtonHandler/tools has jQuery rendered content
- all notifications are made via Messaging API
- only category delete uses callback
- user can add category from PlaceForm
- user can select layer for moved places when category is deleted
- added category is selected in MyData tab
- drawing is handled in one place
- some code is moved to helpers and constants
- localization keys are named and ordered more properly

Not sure if requests and handlers are needed
https://github.com/oskariorg/oskari-frontend/tree/master/bundles/framework/myplaces3/request
because MyPlacesHandler does the same things and MyData tab has confirm functionalities.
Not sure if edit place request is needed for embedded maps etc.

Also other than fi,en,sv,ru,is and fr localization files should be removed because them are outdated and nobody maintains them.


ZakarFin:
Too many changes to have a meaningful review. Merging blindly so we can have this on test env.

PR #1837, opened 2022-06-03
Feature/announcement banner

Creaticide:
Announcements banner.


bundles/framework/announcements/view/AnnouncementsBanner.jsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import { showBanner } from 'oskari-ui/components/window';
+import { Checkbox, Message, Pagination, Link } from 'oskari-ui';
+import { BUNDLE_KEY } from '..//constants';
```
ZakarFin, line 4:
```suggestion
import { BUNDLE_KEY } from '../constants';
```

src/react/components/window/index.js
```diff
@@ -185,3 +186,41 @@ export const showFlyout = (title, content, onClose, options = {}) => {
...
+ * @returns 
+ */
+export const showBanner = (type, onClose, props) => {
+    const element = createTmpContainer();
+    const removeWindow = createRemoveFn(element, onClose);
+    const bringToTop = createBringToTop(element);
+
+    const render = (props) => {
+        ReactDOM.render(
+            <Alert
```
ZakarFin, line 205: Does Alert bring something to the table here or could we just use some stylized div?

ZakarFin: Didn't think of this before, but it might be a very good idea to have this banner-functionality here as a common component 👍

Creaticide: Alert seemed like a good idea at first, but div would probably be better.

PR #1862, opened 2022-06-09
Show/hide mydata tabs with config.

Creaticide:
Show/hide mydata tabs depending on config.


ZakarFin:
Link for server-side counterpart: oskariorg/oskari-server#862

PR #1878, opened 2022-06-15
Statsgrid own indicators fix

okauppinen:
Split data (saveIndicatorData) and indicator (saveIndicator) validation. This allows user to update inicator (name, desc, source) without editing/adding data.

Add DatasourceEvent listeners to update ui (selected indicators, classification, table, disagram) when own indicator is updated.

Fix showAddDatasetForm (this => me). Caused js error and didn't show add dataset form (year, regionset, add button).

Use existing StatisticService in MyIndicatorsHandler (don't create new). Double service caused unwanted behavior as service triggers events and creates cache.


bundles/statistics/statsgrid2016/view/IndicatorFormFlyout.js
```diff
@@ -276,17 +268,18 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.IndicatorFormFlyout', func
...
             return false;
         }
-
-        if (typeof regionValues.values === 'undefined' || regionValues.values.length === 0) {
-            return false;
-        }
-
+        return true;
+    },
+    validateIndicatorData: function (regionValues) {
```
ZakarFin, line 279: Doesn't this mean that you need to fill in data for all regions for validation to pass?

ZakarFin: Or does this not include the regions that user didn't fill in?

okauppinen: Didn't change data validation. Only removed length check because saveIndicatorData is called only if  if valuesForm.values.length. 

Values are modified/checked in different places. Validation shouldn't get empty values:
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/statistics/statsgrid2016/view/IndicatorFormFlyout.js#L178
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/statistics/statsgrid2016/components/IndicatorDataForm.js#L131
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/statistics/statsgrid2016/view/IndicatorFormFlyout.js#L336

PR #1884, opened 2022-06-16
Admin layereditor show coverage

okauppinen:


ZakarFin:
This is a good feature but lets add a button that admin actively clicks to show the coverage instead of automatically adding it. We can build on that with capabilties diffs to show if coverage changed.

PR #1895, opened 2022-06-22
Feature/published info

Creaticide:
Publisher general info react form.

### DIFF
--- a/bundles/framework/publisher2/view/form/GeneralInfoForm.jsx
+++ b/bundles/framework/publisher2/view/form/GeneralInfoForm.jsx
@@ -1,0 +1,74 @@
+import React, { useState } from 'react';
+import { LabeledInput, Select, Option, Message, Label } from 'oskari-ui';
+import styled from 'styled-components';
+import { InfoIcon } from 'oskari-ui/components/icons';
+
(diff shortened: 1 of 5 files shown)


bundles/framework/publisher2/view/form/GeneralInfoForm.jsx
```diff
@@ -0,0 +1,76 @@
...
+            <FieldWithInfo>
+                <LabeledInput
+                    type='text'
+                    label={<Message messageKey='BasicView.name.label' bundleKey={BUNDLE_KEY} />}
+                    name='name'
+                    value={name}
+                    mandatory={true}
+                    onChange={(e) => {
+                        setName(e.target.value)
+                        onChange(e)
```
ZakarFin, line 34: Let's not expose the event outside this component. Instead onChange() could receive an object that holds all of the field values. Something like: 
```
{
    name: 'asdf,
    domain: 'adsGsa",
    lang: 'fi'
}

ZakarFin: It might read better if we just had ^ this as "state" object for the component and field updates would call setState() and onChange() with the same value.

PR #1897, opened 2022-06-23
React use common DeleteButton

okauppinen:
Change to use common DeleteButton. 

Add hover effect for DeleteIcon (not sure if theme color should be used but at least it's in one place now)
Add DeleteIcon to get same style (color, hover) than DeleteButton uses.
Add index.js for theme.

### DIFF
--- a/bundles/admin/admin-layeranalytics/LayerAnalyticsDetails.jsx
+++ b/bundles/admin/admin-layeranalytics/LayerAnalyticsDetails.jsx
@@ -1,8 +1,9 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { Table } from 'oskari-ui/components/Table';
-import { Button, Confirm, Message, Space, Spin, Tooltip } from 'oskari-ui';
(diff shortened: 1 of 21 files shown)


bundles/admin/admin-layeranalytics/LayerAnalyticsDetails.jsx
```diff
@@ -1,8 +1,9 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { Table } from 'oskari-ui/components/Table';
-import { Button, Confirm, Message, Space, Spin, Tooltip } from 'oskari-ui';
-import { ArrowLeftOutlined, DeleteOutlined } from '@ant-design/icons';
+import { Button, Message, Space, Spin, Tooltip } from 'oskari-ui';
+import { DeleteButton } from 'oskari-ui/components/buttons';
+import { ArrowLeftOutlined } from '@ant-design/icons';
 import { red } from '@ant-design/colors'
```
ZakarFin, line 7: Can we remove this import as well?

PR #1901, opened 2022-06-28
Feature/os 1792 published metadata

Creaticide:
Published maps layer selection popup in React + metadata links for layers.

### DIFF
--- a/bundles/framework/publisher2/tools/LayerSelectionTool.js
+++ b/bundles/framework/publisher2/tools/LayerSelectionTool.js
@@ -14,6 +14,9 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.LayerSelectionTool',
         _templates: {
             extraOptions: jQuery(`
             <div class="publisher2 background-layer-selector tool-options">
+                <div class="metadata-selection">
+                    <label><input type="checkbox"/><span></span></label>
+                </div>
                 <div class="style-selection">
                     <label><input type="checkbox"/><span></span></label>
                 </div>
@@ -113,10 +116,16 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.LayerSelectionTool',
                 const isAllowStyleChange = initialConf && initialConf.config && initialConf.config.isStyleSelectable;
                 const extraOptions = me._templates.extraOptions.clone();
                 extraOptions.find('.style-selection label span').append(me.__loc.layerselection.allowStyleChange);
+                extraOptions.find('.metadata-selection label span').append(me.__loc.layerselection.showMetadata);
                 const allowCheckbox = extraOptions.find('.style-selection label input')
                     .on('change', function () {
                         const isChecked = jQuery(this).is(':checked');
                         me.__plugin.setStyleSelectable(isChecked);
+                    });
+                extraOptions.find('.metadata-selection label input')
+                    .on('change', function () {
+                        const isChecked = jQuery(this).is(':checked');
+                        me.__plugin.setShowMetadata(isChecked);
                     });
                 if (isAllowStyleChange) {
                     allowCheckbox.prop('checked', true).change();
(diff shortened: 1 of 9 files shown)


bundles/framework/publisher2/tools/LayerSelectionTool.js
```diff
@@ -128,10 +137,22 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.LayerSelectionTool',
                     var layer = layers[i];
                     me._addLayer(layer);
                 }
+                me._metadataAvailable(layers);
```
ZakarFin, line 140: Add `_metadataAvailable()` calls to other places where `_checkCanChangeStyle()` is called (on maplayer add/remove atleast). The init-function should also check if the checkbox is checked when editing a map. Looks like it doesn't for styleselection either, but it's done here: https://github.com/oskariorg/oskari-frontend/blob/3c333284e1c498a7d614bf085daa1dc61ae5b2c4/bundles/framework/publisher2/tools/LayerSelectionTool.js#L130-L132. We can do the same thing here.

bundles/mapping/mapmodule/plugin/layers/LayerSelectionPopup.js
```diff
@@ -0,0 +1,123 @@
...
+                return (
+                    <LayerRow key={layer.getId()}>
+                        <div>
+                            <StyledCheckbox
+                                checked={layer.isVisible()}
+                                onChange={e => setLayerVisibility(layer, e.target.checked, false)}
+                            />
+                            <span>{layer.getName()}</span>
+                        </div>
+                        {showMetadata && (<MetadataIcon metadataId={layer.getMetadataIdentifier()} />)}
```
ZakarFin, line 78: Just noticed that this new implementation is missing the styleChange option. We need to implement that as well.

bundles/mapping/mapmodule/plugin/layers/LayerSelectionPlugin.js
```diff
@@ -745,18 +776,29 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.LayerSelectionP
...
+                    me.popupCleanup();
                 } else {
-                    me.openSelection();
+                    const mapModule = me.getMapModule();
+                    me.popupControls = showLayerSelectionPopup(
+                        me._baseLayers,
+                        me._layers,
+                        () => me.popupCleanup(),
+                        me._showMetadata,
+                        (layer, visible, isBaseLayer) => me._setLayerVisible(layer, visible, isBaseLayer),
```
ZakarFin, line 788: We need to pass `this._styleSelectable` here as well to see if style selection should be shown for user.

src/react/components/window/Popup.jsx
```diff
@@ -135,8 +135,14 @@ export const Popup = ThemeConsumer(( {title = '', children, onClose, bringToTop,
     </div>
     */
 
-    const headerTheme = getHeaderTheme(theme);
-    return (<Container {...containerProps}>
+    let headerTheme;
+    if (options.customTheme) {
+        headerTheme = getHeaderTheme(options.customTheme);
```
ZakarFin, line 140: This feels a bit clunky, but right now I don't have a better suggestion how to do it. I'll have to think on it a bit more. At least headerTheme shouldn't be involved with the custom theme in my opinion for the style we are using here. Maybe just allow passing  a style-object directly instead? I'll try to think of a better way.

src/react/components/window/index.js
```diff
@@ -125,6 +125,7 @@ const createBringToTop = (element) => {
  */
 export const showPopup = (title, content, onClose, options = {}) => {
     validate(options, TYPE.POPUP);
+
```
ZakarFin, line 128: Handle theme here instead by passing `options.theme` to ThemeProvider. If it's not defined it will default to the current theme and if it is defined the given one will be used.

src/react/theme/ThemeHelper.js
```diff
@@ -10,7 +10,8 @@ export const getHeaderTheme = (theme) => {
         getBgBorderBottomColor: () => getColorEffect(theme.color.accent, 20),
         getTextColor: () => theme.color.header?.text || headerTextColor,
         getToolColor: () => theme.color.header?.icon || funcs.getTextColor(),
-        getToolHoverColor: () => theme.color.accent
+        getToolHoverColor: () => theme.color.accent,
+        getOtherStyles: () => theme.otherStyles
```
ZakarFin, line 14: We can remove getOtherStyles() if we pass the theme through showPopup() options and add another helper for the font. Something like this:

```
export const getFont = (theme) => {
    if (theme.font === 'arial') return 'Arial, Helvetica, sans-serif';
    if (theme.font === 'georgia') return 'Georgia, Times, "Times New Roman"';
    return theme.font ||  'Arial, Helvetica, sans-serif';
}
```

PR #1914, opened 2022-08-03
Notify user when layer is not visible.

Creaticide:
Notify user when selected layer is not visible.

### DIFF
--- a/bundles/mapping/mapmodule/plugin/layers/LayersPluginClass.ol.js
+++ b/bundles/mapping/mapmodule/plugin/layers/LayersPluginClass.ol.js
@@ -1,4 +1,5 @@
 import olFormatWKT from 'ol/format/WKT';
+import { Messaging } from 'oskari-ui/util';
 
 const WKT_READER = new olFormatWKT();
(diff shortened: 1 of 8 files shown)


bundles/mapping/mapmodule/plugin/layers/LayersPluginClass.ol.js
```diff
@@ -321,6 +322,16 @@ export class LayersPlugin extends AbstractMapModulePlugin {
     }
     notifyLayerVisibilityChanged (layer, inScale, geometryMatch) {
         var event = Oskari.eventBuilder('MapLayerVisibilityChangedEvent')(layer, inScale, geometryMatch);
+
+        if (layer.isVisible()) {
+            if (!inScale) {
+                Messaging.notify(Oskari.getMsg('MapModule', 'layerVisibility.notInScale', { name: layer.getName() }));
```
ZakarFin, line 328: This will affect embedded maps which might use runtime vector layers that don’t have names or at least user-friendly names. I don’t have a good suggestion for this now but I’m very hesitant for merging this as is. I think we need some other solution for this or handle some edge cases at least.

PR #1917, opened 2022-08-05
Show indicators without data.

Creaticide:
Show indicators without data.

### DIFF
--- a/bundles/statistics/statsgrid2016/view/IndicatorFormFlyout.js
+++ b/bundles/statistics/statsgrid2016/view/IndicatorFormFlyout.js
@@ -89,16 +89,19 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.IndicatorFormFlyout', func
             const datasets = [];
             ind.selectors.forEach(function (sel) {
                 ind.regionsets.forEach(function (regionset) {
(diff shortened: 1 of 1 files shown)


bundles/statistics/statsgrid2016/view/IndicatorFormFlyout.js
```diff
@@ -89,16 +89,19 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.view.IndicatorFormFlyout', func
...
-                        if (typeof value === 'object') {
-                            data[sel.id] = value.id;
-                        } else {
-                            data[sel.id] = value;
-                        }
-                        data.regionset = regionset;
-                        datasets.push(data);
-                    });
+                    // Has data
+                    if (regionset !== 0) {
```
ZakarFin, line 93: What does 0 mean here? I think all the regionset ids are positive integers?

Creaticide: Regionset id was 0 if data doesn't exist, if I remember correctly.

ZakarFin:
New version in #1957 that removes the dummy region set before UI knows about it.

PR #1919, opened 2022-08-10
Disable gfi popup if layer is hidden by swipe tool.

Creaticide:
Disable gfi popup if layer is hidden by swipe tool.

### DIFF
--- a/bundles/mapping/layerswipe/instance.js
+++ b/bundles/mapping/layerswipe/instance.js
@@ -16,6 +16,7 @@ Oskari.clazz.define(
         this.cropSize = null;
         this.mapModule = null;
         this.layer = null; // ol layer
+        this.oskariLayer = null;
         this.popupService = null;
         this.popup = null;
         this.loc = Oskari.getMsg.bind(null, 'LayerSwipe');
@@ -71,6 +72,7 @@ Oskari.clazz.define(
                 this.unregisterEventListeners();
                 this.hideSplitter();
                 Oskari.getSandbox().getService('Oskari.mapframework.service.VectorFeatureService').setHoverEnabled(true);
+                this.mapModule.setSwipeStatus(false, null, null);
             }
             this.active = active;
             this.mapModule.getMap().render();
@@ -78,7 +80,13 @@ Oskari.clazz.define(
 
         updateSwipeLayer: function () {
             this.unregisterEventListeners();
-            this.layer = this.getTopmostLayer();
+            const topLayer = this.getTopmostLayer();
+            this.layer = topLayer.ol;
+            this.oskariLayer = topLayer.layerId;
+
+            if (this.oskariLayer !== null) {
+                this.mapModule.setSwipeStatus(true, this.cropSize, this.oskariLayer);
+            }
 
             if (this.alertTimer) {
                 clearTimeout(this.alertTimer);
@@ -153,7 +161,10 @@ Oskari.clazz.define(
                 this.showNotVisibleAlert(layerId, !isInGeometry);
             }
             const olLayers = this.mapModule.getOLMapLayers(layerId);
-            return olLayers.length !== 0 ? olLayers[0] : null;
+            return {
+                ol: olLayers.length !== 0 ? olLayers[0] : null,
+                layerId: layerId
+            };
         },
 
         registerEventListeners: function () {
@@ -223,6 +234,7 @@ Oskari.clazz.define(
             const splitterOffset = this.getSplitterElement().offset();
             this.cropSize = splitterOffset.left - mapOffset.left + this.splitterWidth / 2;
             this.mapModule.getMap().render();
+            this.mapModule.setSwipeStatus(true, this.cropSize, this.oskariLayer);
         },
 
         showSplitter: function () {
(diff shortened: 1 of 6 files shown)


bundles/mapping/layerswipe/instance.js
```diff
@@ -71,14 +72,21 @@ Oskari.clazz.define(
...
 
         updateSwipeLayer: function () {
             this.unregisterEventListeners();
-            this.layer = this.getTopmostLayer();
+            const topLayer = this.getTopmostLayer();
+            this.layer = topLayer.ol;
+            this.oskariLayer = topLayer.layerId;
+
+            if (this.oskariLayer !== null) {
+                this.mapModule.setSwipeStatus(true, this.cropSize, this.oskariLayer);
```
ZakarFin, line 88: Since the map module doesn't need this status for itself and it's already quite big and complex, I would rather store this on the layer itself (`oskariLayer`). Then we only need to store the `cropSize`. It's currently named as `x` on the status, but I'm thinking it might be more clear if we call it something like something like partiallyHidden or "partial visibility" with an object with keys `right` or `left` or something instead of "swipe". 

Also I would rather have the value in one place instead of having both "cropSize" on the layerswipe AND whatever it ends up on the layer/mapmodule. With this the same value is stored in two places which is error prone. We could remove cropSize from the swipe tool and store it on the layer itself as well.

Any thoughts on this?

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -300,10 +300,16 @@ Oskari.clazz.define(
             const me = this;
             const dteMs = (new Date()).getTime();
             const requestedLayers = layers || me.getSandbox().findAllSelectedMapLayers();
-            const layerIds = me._buildLayerIdList(requestedLayers);
+            let layerIds = me._buildLayerIdList(requestedLayers);
```
ZakarFin, line 303: To make it easier to process here we might want to return the layers instead of their ids so we can filter again with the swipestatus/partial visibility info from the layer.

bundles/mapping/layerswipe/instance.js
ZakarFin: Another thought is that we could save the "swipe status" on the gfi-plugin and skip setting it on the map module. Then it would be almost identical but it would make more sense to store it at the gfi-plugin. Storing the status on the layers would make it harder to tear down the status I suppose since when layers are switched you should remove the partial visibility from one and add it to another.

Creaticide: Swipe status moved to the plugin now.

bundles/mapping/mapmodule/plugin/getinfo/request/SwipeStatusRequest.js
```diff
@@ -0,0 +1,59 @@
...
+         * @method create called automatically on construction
+         * @static
+         *
+         * @param {Number} layerId layer ID
+         * @param {Boolean} active swipe active
+         * @param {Number} cropX swipe tool X position
+         */
+        function (layerId, active, cropX) {
+            this._layerId = layerId;
+            this._active = active;
```
ZakarFin, line 20: I don't think we need the boolean here. If a layerId is present we can deduce that it's active?

bundles/mapping/layerswipe/instance.js
```diff
@@ -71,14 +72,26 @@ Oskari.clazz.define(
                 this.unregisterEventListeners();
                 this.hideSplitter();
                 Oskari.getSandbox().getService('Oskari.mapframework.service.VectorFeatureService').setHoverEnabled(true);
+                this.setSwipeStatus(null, false, null);
             }
             this.active = active;
             this.mapModule.getMap().render();
         },
 
+        setSwipeStatus: function (layerId, active, cropX) {
```
ZakarFin, line 81: Move layer variable setting here:
```suggestion
        setSwipeStatus: function (layerId, active, cropX) {
            this.oskariLayerId = layerId;
```

bundles/mapping/layerswipe/instance.js
```diff
@@ -71,14 +72,26 @@ Oskari.clazz.define(
...
         updateSwipeLayer: function () {
             this.unregisterEventListeners();
-            this.layer = this.getTopmostLayer();
+            const topLayer = this.getTopmostLayer();
+            this.layer = topLayer.ol;
+            this.oskariLayer = topLayer.layerId;
+
+            if (this.oskariLayer !== null) {
+                this.setSwipeStatus(this.oskariLayer, true, this.cropSize);
+            }
```
ZakarFin, line 94:
```suggestion
            if (topLayer.layerId !== null) {
                this.setSwipeStatus(topLayer.layerId, true, this.cropSize);
            }
```

bundles/mapping/layerswipe/instance.js
```diff
@@ -16,6 +16,7 @@ Oskari.clazz.define(
         this.cropSize = null;
         this.mapModule = null;
         this.layer = null; // ol layer
+        this.oskariLayer = null;
```
ZakarFin, line 19: Rename so it's instantly clear it holds the id and not the layer object:
```suggestion
        this.oskariLayerId = null;
```

PR #1922, opened 2022-08-18
Allow editing style/layer for user generated content through "selected layers" in layerlist

Creaticide:
Edit places & userlayers throught map layer list (feature/OS-1848).

### DIFF
--- a/bundles/framework/myplacesimport/request/ShowUserLayerDialogRequest.js
+++ b/bundles/framework/myplacesimport/request/ShowUserLayerDialogRequest.js
@@ -9,10 +9,10 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplacesimport.request.ShowUserL
     /**
      * @method create called automatically on construction
      * @static
-     * @param {Object} values User layer values
+     * @param {Object} id User layer id
      */
-    function (values) {
-        this._values = values;
+    function (id) {
+        this._id = id;
     }, {
         /** @static @property __name request name */
         __name: 'MyPlacesImport.ShowUserLayerDialogRequest',
@@ -27,8 +27,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.myplacesimport.request.ShowUserL
          * @method getId
          * @return {String} tab title
          */
-        getValues: function () {
-            return this._values;
+        getId: function () {
+            return this._id;
         }
     }, {
         /**
--- a/bundles/framework/myplacesimport/request/ShowUserLayerDialogRequest.js
+++ b/bundles/framework/myplacesimport/request/ShowUserLayerDialogRequest.js
@@ -1,0 +1,39 @@
+/**
+ * @class Oskari.mapframework.bundle.myplacesimport.request.ShowUserLayerDialogRequest
+ * Requests tab to be added
+ *
+ * Requests are build and sent through Oskari.Sandbox.
+ * Oskari.mapframework.request.Request superclass documents how to send one.
+ */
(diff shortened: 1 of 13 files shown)


bundles/mapping/userstyle/request/ShowUserStylesRequest.js
```diff
@@ -15,7 +15,7 @@ Oskari.clazz
      * @param {Boolean} showStyle flag indicating that visualization form should be opened to create/edit style
      * @param {String} styleName style identifier so we can initialize visualization form with correct style, if undefined new style is created
      */
-        function (layerId, showStyle, styleName) {
+        function (layerId, showStyle, styleName, layerType) {
```
ZakarFin, line 18: The layerType isn't use so it can be removed.

bundles/mapping/mapuserlayers/domain/UserLayerModelBuilder.js
```diff
@@ -22,6 +22,20 @@ Oskari.clazz.define(
             // call parent parseLayerData
             this.wfsBuilder.parseLayerData(layer, mapLayerJson, maplayerService);
             layer.setLocale(mapLayerJson.locale);
+
+            const values = {
+                locale: layer.getLocale(),
+                style: layer.getCurrentStyle().getFeatureStyle(),
+                id: layer.getId()
+            };
```
ZakarFin, line 30: We should only send layer id in the request since at least the locale can change after the model builder has been run and this would create an old layer name be shown when activating this tool.

bundles/framework/myplacesimport/request/ShowUserLayerDialogRequest.js
```diff
@@ -0,0 +1,39 @@
...
+ * Oskari.mapframework.request.Request superclass documents how to send one.
+ */
+Oskari.clazz.define('Oskari.mapframework.bundle.myplacesimport.request.ShowUserLayerDialogRequest',
+    /**
+     * @method create called automatically on construction
+     * @static
+     * @param {Object} values User layer values
+     */
+    function (values) {
+        this._values = values;
```
ZakarFin, line 15: Use only layer id instead of values object.

bundles/framework/myplacesimport/request/ShowUserLayerDialogRequest.js
```diff
@@ -0,0 +1,39 @@
...
+         * @return {String} request name
+         */
+        getName: function () {
+            return this.__name;
+        },
+        /**
+         * @method getId
+         * @return {String} tab title
+         */
+        getValues: function () {
```
ZakarFin, line 30: Use only layer id instead of values object.

bundles/framework/myplacesimport/request/ShowUserLayerDialogRequestHandler.js
```diff
@@ -0,0 +1,30 @@
...
+    }, {
+    /**
+     * @method handleRequest
+     * @param {Oskari.mapframework.core.Core} core
+     *      reference to the application core (reference sandbox core.getSandbox())
+     * @param {Oskari.mapframework.bundle.myplacesimport.request.ShowUserLayerDialogRequestHandler} request
+     *      request to handle
+     */
+        handleRequest: function (core, request) {
+            this.instance.openLayerDialog(request.getValues());
```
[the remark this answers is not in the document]
ZakarFin: You could add another function to instance that can generate the values object and call openLayerDialog() but only takes the id as parameter.

PR #1927, opened 2022-08-31
Feature/os 1086 guided tour react

Creaticide:
Guided tour popup to React.

### DIFF
--- a/bundles/framework/guidedtour/handler/GuidedTourHandler.js
+++ b/bundles/framework/guidedtour/handler/GuidedTourHandler.js
@@ -1,0 +1,192 @@
+import { StateHandler, controllerMixin } from 'oskari-ui/util';
+import { showGuidedTourPopup } from '../view/GuidedTourPopup';
+
+class TourHandler extends StateHandler {
+    constructor (instance) {
+        super();
+        this.instance = instance;
+        this.sandbox = Oskari.getSandbox();
+        this.setState({
+            step: 0,
+            steps: [],
+            position: null,
+            loading: false
+        });
+        this.popupControls = null;
+        this._localization = Oskari.getLocalization('GuidedTour');
+        this.initSteps();
+    };
+
+    getName () {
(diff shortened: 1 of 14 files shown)


bundles/framework/guidedtour/handler/GuidedTourHandler.js
```diff
@@ -0,0 +1,192 @@
...
+    getName () {
+        return 'GuidedTourHandler';
+    }
+
+    initSteps () {
+        const delegate = {
+            bundleName: this.instance.getName(),
+            priority: 0,
+            getTitle: () => {
+                return this._localization.page1.title;
```
ZakarFin, line 29: Use Oskari.getMessage() instead.

bundles/framework/guidedtour/handler/GuidedTourHandler.js
```diff
@@ -0,0 +1,192 @@
...
+                        if (success) {
+                            delegate.getContent = () => { return response.body; };
+                            delegate.getTitle = () => { return response.title; };
+                        } else {
+                            Oskari.log(this.instance.getName()).error('Failed to load guided tour content for step "' + stepSpec[index].bundleName + '" with tags: ' + content);
+                        }
+                    });
+                }
+            }
+        }
```
ZakarFin, line 67: Move this block to it's own function that takes the delegate and the instance.conf.steps as params and returns a new delegate to be used for the rest of the function.

bundles/framework/guidedtour/handler/GuidedTourHandler.js
```diff
@@ -0,0 +1,192 @@
...
+        if (typeof delegate.priority === 'number') {
+            const priorities = this.state.steps.map((d) => d.priority);
+            const insertLocation = _.sortedIndex(priorities, delegate.priority);
```
ZakarFin, line 70: Replace lodash with something else since we are trying to get rid of it at some point.

PR #1928, opened 2022-09-01
Fix map state reset.

Creaticide:
Fix map state reset.


bundles/mapping/mapmodule/plugin/panbuttons/PanButtons.js
```diff
@@ -145,7 +145,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.PanButtons',
                     'StateHandler.SetStateRequest'
                 );
                 if (requestBuilder) {
```
ZakarFin, line 147: I think you should be able to use this instead to clean up the code above a bit:
```suggestion
                if (sandbox.hasHander('StateHandler.SetStateRequest')) {
```

PR #1929, opened 2022-09-01
Disable notifications of layer not in scale/geometry.

Creaticide:


ZakarFin:
The notifications added in #1914 added a bit too much noise for map usage and was decided to be removed after testing it out. This removes the most annoying notifications.

PR #1930, opened 2022-09-01
Feature/os 1975

Creaticide:
Missing translation.

### DIFF
--- a/bundles/mapping/mapuserlayers/domain/UserLayerModelBuilder.js
+++ b/bundles/mapping/mapuserlayers/domain/UserLayerModelBuilder.js
@@ -23,11 +23,12 @@ Oskari.clazz.define(
             this.wfsBuilder.parseLayerData(layer, mapLayerJson, maplayerService);
             layer.setLocale(mapLayerJson.locale);
 
-            const loc = Oskari.getLocalization('MapWfs2')['own-style'];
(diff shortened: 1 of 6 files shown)


bundles/mapping/mapuserlayers/domain/UserLayerModelBuilder.js
```diff
@@ -23,11 +23,12 @@ Oskari.clazz.define(
             this.wfsBuilder.parseLayerData(layer, mapLayerJson, maplayerService);
             layer.setLocale(mapLayerJson.locale);
 
-            const loc = Oskari.getLocalization('MapWfs2')['own-style'];
+            const toolName = Oskari.getMsg('MapMyPlaces', 'editLayer');
```
ZakarFin, line 26: We'll need to add this to userlayer localization as well and use it from there since it could be that userlayers are used on an instance and mylayers are not.

ZakarFin: The previous ones using from MapWfs2 was ok since both expect it to be present to work properly. But it's fine for now if we just duplicate the localization for both.

ZakarFin: Lets use  MapWfs2 for locale

PR #1935, opened 2022-09-05
Declutter option to vector tile layers.

Creaticide:
Declutter option switch to vector tile layers.

### DIFF
--- a/bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/Declutter.jsx
+++ b/bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/Declutter.jsx
@@ -1,20 +1,16 @@
-import React, { useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { Controller } from 'oskari-ui/util';
 import { Message, Switch } from 'oskari-ui';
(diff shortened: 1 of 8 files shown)


bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/Declutter.jsx
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Controller } from 'oskari-ui/util';
+import { Message, Switch } from 'oskari-ui';
+import { StyledFormField, SpacedLabel } from '../styled';
+
+export const Declutter = ({ layer, controller }) => {
+    const [checked, setChecked] = useState(layer.options && layer.options.declutter ? layer.options.declutter : false);
```
ZakarFin, line 8: I don't think we need to use state here. It's more like this one: https://github.com/oskariorg/oskari-frontend/blob/0b14e61788e49ffdb4a65a0cdfac5afb55607941/bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/WfsRenderMode.jsx

We always get the current value from the `layer` prop.

PR #1942, opened 2022-09-08
Drawtools intersection fix

okauppinen:
Fix intersection check for given geojson and modify end.

There was intersection check for multi geometry (eg edit myplace).
And measurement + overlay was added in add modify interaction.
=> moved these after geojson parsing (measurements are shown for given features also if modify isn't allowed)

Call pointerMoveHandler on modifyend to show warning message.

### DIFF
--- a/bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
+++ b/bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
@@ -382,7 +382,6 @@ Oskari.clazz.define(
             me.removeInteractions(me._draw, id);
             me.removeInteractions(me._modify, id);
             me._cleanupInternalState();
-            me.getMap().un('pointermove', me.pointerMoveHandler, me);
             // enable gfi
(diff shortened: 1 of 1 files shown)


ZakarFin:
Should we just replace all calls to `pointerMoveHandler()` to `updateDrawingTooltip()` since that's the only thing it seems to do anymore.

```
        pointerMoveHandler: function (evt) {
            if (!this._sketch || !this.getOpts('showMeasureOnMap')) {
                // if no drawing of we don't want to show it on map -> skip
                return;
            }
            this.updateDrawingTooltip(this._sketch);
        },
// ^remove this and call updateDrawingTooltip directly using default value with the check added at the start?
        updateDrawingTooltip: function (feature = this._sketch) {
            if (!feature || !this.getOpts('showMeasureOnMap')) {
                // if no drawing of we don't want to show it on map -> skip
                return;
            }
            const id = feature.getId();
            ...
```

PR #1983, opened 2022-10-07
Feature/os 1936 maprotator timecontrol

Creaticide:
Time control & map rotator buttons.


bundles/mapping/mapmodule/MapModuleButton.jsx
```diff
@@ -1,5 +1,13 @@
 import React from 'react';
 import { MapButton } from 'oskari-ui/components/buttons';
+import styled from 'styled-components';
+
+const Container = styled('div')`
+    margin: 0 0 10px 30px;
+    width: 32px;
+    height: 32px;
+    z-index: 15000;
```
ZakarFin, line 9: Do we need z-index here?

bundles/framework/maplegend/resources/scss/style.scss
```diff
@@ -5,6 +5,12 @@ $darkIcon: url("../images/Question-mark-dark.png");
 $lightIcon: url("../images/Question-mark-light.png");
 
 .maplegend .maplegend {
+  width: 32px;
+  height: 32px;
+  cursor: pointer;
+  position: relative;
+  z-index: 15000;
+  margin-bottom: 10px !important;
```
ZakarFin, line 13: Do we need these here or why does this tool need these instead of ones in MapModuleButton.jsx?

PR #1988, opened 2022-10-12
Zoom slider

Creaticide:
Zoom slider


bundles/mapping/mapmodule/plugin/zoombar/ZoomSlider.jsx
```diff
@@ -0,0 +1,183 @@
+import React from 'react';
+import { MapModuleButton } from '../../MapModuleButton';
+import styled from 'styled-components';
+import { Slider } from 'antd';
```
ZakarFin, line 4: Use https://github.com/oskariorg/oskari-frontend/blob/master/src/react/components/Slider.jsx instead if possible

bundles/framework/layerlist/view/LayerViewTabs/SelectedLayers/LayerBox/Footer/OpacitySlider.jsx
```diff
@@ -14,6 +14,19 @@ const Border = styled('div')`
...
+    .ant-slider-handle {
+        border: #0091ff solid 2px;
+        margin-top: -6px;
+    }
+    &:hover .ant-slider-track {
+        background-color: #003fc3 !important;
+    }
+    &:hover .ant-slider-handle {
+        border: #003fc3 solid 2px !important;
+    }
```
ZakarFin, line 29: Could we use src/react/components/Opacity.jsx here instead so we don't have to assign style twice (here and in src/react/components/Opacity.jsx)? It's currently used in bundles\admin\admin-layereditor\view\AdminLayerForm\VisualizationTabPane\Opacity.jsx but since it's a common component we should be able to use it here as well.

PR #1990, opened 2022-10-14
New icons, styles & fixes.

Creaticide:
New icons, styles & fixes.


src/react/components/icons/Backward.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export const Backward = () => (
+    <svg id="Layer_2" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 11.19 8.1">
+        <g id="Layer_1-2">
```
ZakarFin, line 5: Lets remove the ids from all of these SVGs

PR #1992, opened 2022-10-25
Search bar in React.

Creaticide:
Search bar in React.


bundles/mapping/mapmodule/plugin/search/SearchBar.jsx
```diff
@@ -0,0 +1,111 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Input } from 'antd';
+import { Button } from 'oskari-ui';
```
ZakarFin, line 4: Can we use the TextInput from oskari-ui instead?
```suggestion
import { TextInput, Button } from 'oskari-ui';
```

PR #1993, opened 2022-10-26
Fix/opacity slider style

Creaticide:
Opacity slider styling.


src/react/components/Opacity.jsx
```diff
@@ -61,23 +73,25 @@ export class Opacity extends React.Component {
         const { opacity } = this.state;
         return (
             <Container>
-                <StyledSlider>
+                <StyledSlider bordered={this.props.bordered}>
```
ZakarFin, line 76: We should declare bordered in Opacity.propTypes below.

PR #2001, opened 2022-11-01
Remove mobile toolbar & mobile button implementations

Creaticide:
Removed mobile toolbar & mobile buttons from map module plugins.

### DIFF
--- a/bundles/mapping/mapmodule/plugin/publishertoolbar/PublisherToolbarPlugin.js
+++ b/bundles/mapping/mapmodule/plugin/publishertoolbar/PublisherToolbarPlugin.js
@@ -25,7 +25,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.PublisherToolba
         me._defaultLocation = 'top left';
         me._index = 0;
         me._name = 'PublisherToolbarPlugin';
-        me._toolButtons = [];
+        me._toolButtons = conf.buttons || [];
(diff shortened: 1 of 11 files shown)


bundles/mapping/mapmodule/plugin/publishertoolbar/PublisherToolbarPlugin.js
```diff
@@ -25,50 +25,15 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.PublisherToolba
...
-                            me._openToolsPopup();
-                        }
-                    }
-                }
-            },
-            buttonGroup: 'mobile-toolbar'
-        };
-
-        me._buttons = conf.buttons || [];
+        me._toolButtons = [];
```
ZakarFin, line 28: We'll probably need to initialize toolButtons like the buttons array was before so we can restore the selections on embedded map. Otherwise this only works on the publisher functionality itself?

bundles/mapping/mapmodule/plugin/search/SearchPlugin.js
```diff
@@ -503,38 +505,21 @@ Oskari.clazz.define(
 
             // remove old element
             this.teardownUI();
+            this.inMobileMode = isMobile;
+            me._element.removeClass('mobilesearch');
```
ZakarFin, line 509: Does something add this mobilesearch class if we always just remove it?

PR #2012, opened 2022-11-08
Fix feature data button styles.

Creaticide:


ZakarFin:
We should remove these too:
https://github.com/oskariorg/oskari-frontend/blob/4ae0da2d3bbc0e99b189cdebb31ca433fe63c4f0/bundles/framework/featuredata2/resources/scss/style.scss#L157-L159
https://github.com/oskariorg/oskari-frontend/blob/4ae0da2d3bbc0e99b189cdebb31ca433fe63c4f0/bundles/framework/featuredata2/resources/images/ajax-loader.gif

PR #2013, opened 2022-11-09
Better tooltip positions.

Creaticide:


bundles/mapping/mapmodule/MapModuleButton.jsx
```diff
@@ -97,6 +97,7 @@ export const MapModuleButton = ({ styleName, title, icon, onClick, size = '32px'
                 iconSize={iconSize}
                 className={className}
                 disabled={disabled}
+                position={position}
```
ZakarFin, line 100: Do we still need toolbarDirection prop here or could we deduce it from position?

Creaticide: Not really, but a separate prop would allow a different direction if somebody wants to use that later.. Maybe even up/down, which is not implemented at the moment though. But can be removed if we don't want that.

ZakarFin: I guess we can keep it but maybe default it based on position?

PR #2020, opened 2022-11-14
Disable style select if layer has 1 style.

Creaticide:
Disable layer style selection if layer has only 1 style.


ZakarFin:
Cherry-picking the commit to develop as this isn't directly related to the work on map-buttons branch

PR #2031, opened 2022-11-22
Remove mapplugin location restrictions.

Creaticide:
Removed restrictions from map plugin layout drag and drop locations.


ZakarFin:
Did you try assigning allowedLocations: ['*'] and allowedSiblings: ['*'] as defaults in `AbstractPluginTool` instead of defining them in all tools?

PR #2044, opened 2022-11-28
Option to show pan tool and/or reset button.

Creaticide:
Separate options to show reset button / pan buttons.

### DIFF
--- a/bundles/framework/publisher2/tools/PanButtonsTool.js
+++ b/bundles/framework/publisher2/tools/PanButtonsTool.js
@@ -4,29 +4,39 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.PanButtonsTool',
         index: 2,
         lefthanded: 'top left',
         righthanded: 'top right',
-
         groupedSiblings: true,
+        _templates: {
+            extraOptions: jQuery(`
+                <div class="publisher2 panbutton-options tool-options">
+                    <div class="arrows-selection">
+                        <label><input type="checkbox"/><span></span></label>
+                    </div>
+                </div>
+            `)
+        },
+        extraOptions: null,
 
         /**
-    * Get tool object.
-    * @method getTool
-    *
-    * @returns {Object} tool description
-    */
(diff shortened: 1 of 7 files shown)


bundles/framework/publisher2/tools/PanButtonsTool.js
```diff
@@ -43,6 +53,63 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.PanButtonsTool',
...
+            const { conf = {} } = mapfull;
+            const { plugins = [] } = conf;
+            // data.configuration.mapfull.conf.plugins
+            return plugins.find(plug => plug.id === 'Oskari.mapframework.bundle.mapmodule.plugin.PanButtons');
+        },
+        init: function (data) {
+            var me = this;
+            me.data = data;
+
+            if (data.configuration && data.configuration.mapfull && data.configuration.mapfull.conf && data.configuration.mapfull.conf.plugins) {
```
ZakarFin, line 104: You could use `_getToolPluginMapfullConf()` to make this a bit cleaner as it does the same thing

bundles/framework/publisher2/tools/PanButtonsTool.js
```diff
@@ -43,6 +53,63 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.PanButtonsTool',
...
+                const plugins = data.configuration.mapfull.conf.plugins;
+                plugins.forEach(function (plugin) {
+                    if (me.getTool().id !== plugin.id) {
+                        return;
+                    }
+                    me.setEnabled(true);
```
ZakarFin, line 110: Does this work when editing a previously published map so that it restores the arrow options from previous state or do you need to do it here? It might work since you are using the plugin config, but just want to know if it's been tested as it's one of the most common problems with publisher tools :)

Creaticide: Should work correctly.

PR #2057, opened 2022-12-05
Map reset popup in React.

Creaticide:
Map reset popup in React.


ZakarFin:
We might want to move this to core functionality or to separate bundle, but for now it's good as is.

PR #2060, opened 2022-12-08
Fix feature data button duplicating.

Creaticide:
Fix feature data button duplicating.


ZakarFin:
This solution won't work as the button is not updated when layers on the map change.

PR #2062, opened 2022-12-09
Drawtools refactoring

okauppinen:
Remove id or shape specific draw layers and interactions. Use same functions for new, updated and finished features => code is clearer to read and buffer, measurements etc works for every shape types.

Add buffer for every feature (not just for sketch) if buffer is requested.

`figure out why we have some variables that are "globally reset" and some that are functionality id specific.
As some are "global"/shared between functionalities resuming a previous id will probably NOT work the way expected
and storing these for each id is probably not needed.`

`some of the variables used are shared between functionalities anyway. Consider using just "currentLayer"`


bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
```diff
@@ -33,7 +33,8 @@ const OPTIONS = {
     allowMultipleDrawing: true,
     showMeasureOnMap: false,
     buffer: 0,
-    bufferAccuracy: 10 // is number of line segments used to represent a quadrant circle
+    bufferAccuracy: 10, // is number of line segments used to represent a quadrant circle
+    sidesForCircle: 50 // is number of sides/vertices used to reprsesnt circle as polygon
```
ZakarFin, line 37:
```suggestion
    sidesForCircle: 50 // is number of sides/vertices used to represent circle as polygon
```

bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
```diff
@@ -123,83 +141,101 @@ Oskari.clazz.define(
...
-            if (modifyControl) {
-                this.addModifyInteraction(this.getCurrentLayerId(), shape);
+            let featuresFromJson = geoJsonFormatter.readFeatures(geojson);
+            // parse multi geometries to single geometries
+            if (this.getOpts('allowMultipleDrawing') === 'multiGeom') {
+                featuresFromJson = this.parseMultiGeometries(featuresFromJson);
             }
+            featuresFromJson.forEach(f => {
+                this.onNewFeature(f);
+                this.onFinishedFeature(f, true);
```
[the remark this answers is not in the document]
ZakarFin: Ah, nevermind. It has the suppressEvent true so event is not triggered.

bundles/mapping/drawtools/plugin/DrawPlugin.ol.js
```diff
@@ -998,56 +808,70 @@ Oskari.clazz.define(
...
-            }
-            me.modifyStartEvent(shape);
-            me.getMap().addInteraction(me._modify[me._id]);
+        addModifyInteraction: function () {
+            const isLimited = isModifyLimited(this.getShape());
+            this._modify = new olInteractionModify({
+                source: this.getDrawSource(),
+                style: this.getPointerStyleFunction(),
+                insertVertexCondition: isLimited ? olEventsCondition.never : olEventsCondition.always,
+                deleteCondition: isLimited ? olEventsCondition.never : event => olEventsCondition.shiftKeyOnly(event) && olEventsCondition.singleClick(event)
```
ZakarFin, line 817: I haven't noticed this deleteCondition config before so I tested it out with current version. Drawing a line with start and having shift pressed while clicking the next makes the measurement tooltip stay on screen while removing the line. Should we have this olEventsCondition.never in all conditions instead?

ZakarFin: drawwithshift

Here on the first line I shift click the last point. On second line I finish it with double click. On third using shift click again. For some reason shift with dragging goes to the "freeform" draw, but just clicking with shift pressed does this.

ZakarFin: We should test if having `deleteCondition` always on `olEventsCondition.never` fixes this and maybe leave it on never.

PR #2065, opened 2022-12-19
Coordinate plugin popup in React.

Creaticide:
TODO; 
- Reverse-geocoding needs more testing


bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -0,0 +1,625 @@
...
+        } else {
+            this.updateState({
+                emergencyInfo: null
+            });
+        }
+        this.updatePopup();
+    }
+
+    formatEmergencyCallMessage (data) {
+        const degmin = this.coordinateTransformationExtension._formatDegrees(data.lonlat.lon, data.lonlat.lat, 'min');
```
ZakarFin, line 496: Could we use this from Oskari.util instead? https://github.com/oskariorg/oskari-frontend/blob/develop/src/util.js#L507

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -0,0 +1,625 @@
...
+    'showPopup',
+    'markersSupported',
+    'setMarker',
+    'centerMap',
+    'setLat',
+    'setLon',
+    'getCrsText',
+    'getPreciseTransform',
+    'setLoading',
+    'getSupportedProjections',
```
ZakarFin, line 615: Controller shouldn't have getters. We should pass getSupportedProjections() etc as props to the React-component.

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -0,0 +1,625 @@
...
+            showMouseCoordinates: false,
+            selectedProjection: this.mapModule.getProjection(),
+            originalProjection: this.mapModule.getProjection(),
+            emergencyInfo: null,
+            showReverseGeoCode: false
+        });
+        this.popupControls = null;
+        this.eventHandlers = this.createEventHandlers();
+        this.coordinateTransformationExtension =
+            Oskari.clazz.create('Oskari.mapframework.bundle.coordinatetool.plugin.CoordinateTransformationExtension',
```
ZakarFin, line 30: We probably don't need CoordinateTransformationExtension as is anymore. Maybe pick out some functions that are still used from that and add them to a helper.js that is imported with ES imports or something like that.

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -0,0 +1,625 @@
...
+
+    setSelectedProjection (projection) {
+        this.updateState({
+            selectedProjection: projection
+        });
+        this.updateLonLat();
+    }
+
+    showErrorMessage (message, title) {
+        const dialog = Oskari.clazz.create('Oskari.userinterface.component.Popup');
```
ZakarFin, line 206: Let's move these error messages from jQuery popup to using Messaging.error() or showing the error message next/close to the field (whichever makes more sense for the error).

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -0,0 +1,626 @@
...
+                            emergencyInfo: emergencyData
+                        });
+                    });
+            }
+        } else {
+            this.updateState({
+                emergencyInfo: null
+            });
+        }
+        this.updatePopup();
```
ZakarFin, line 498: Config shouldn't change at runtime so we shouldn't need to updateState() on else nor updatePopup() if `showEmergencyCallMessage` is not configured.

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -0,0 +1,626 @@
...
+        this.mapModule = mapModule;
+        this.config = config;
+        this.setState({
+            xy: {},
+            displayXy: {},
+            loading: false,
+            popupControls: null,
+            showMouseCoordinates: false,
+            selectedProjection: this.mapModule.getProjection(),
+            originalProjection: this.mapModule.getProjection(),
```
ZakarFin, line 24: Original projection shouldn't change at runtime so we could move it out from state.

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -0,0 +1,626 @@
+import { StateHandler, controllerMixin, Messaging } from 'oskari-ui/util';
+import { showCoordinatePopup } from './CoordinatePopup';
+import { PLACEMENTS } from 'oskari-ui/components/window';
+import { getTransformedCoordinatesFromServer, transformCoordinates, formatDegrees } from './helper';
```
ZakarFin, line 4: getTransformedCoordinatesFromServer is implemented in this file and the imported one isn't used?

bundles/framework/coordinatetool/plugin/helper.js
```diff
@@ -0,0 +1,114 @@
+/**
+ * Transforms the given coordinates using action_route=Coordinates and updates coordinates to the UI
+ * @method getTransformedCoordinatesFromServer
+ * @param {Object} data: {lonlat: lat: '', lon: ''} coordinates to be transformed
+ * @param {String} srs: projection for given lonlat params like "EPSG:4326"
+ * @param {String} targetSRS: projection to transform to like "EPSG:4326"
+ * @param {Function} successCb success callback
+ * @param {Function} errorCb error callback
+ */
+export const getTransformedCoordinatesFromServer = async (mapModule, data, srs, targetSRS) => {
```
ZakarFin, line 10: Looks like this isn't used anywhere, but if it is we should pass the mapModule.getProjection() as param instead of the whole mapModule since it's the only thing that is used from mapModule here.

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
Creaticide: This is called from a function in handler with identical name, which is a bit confusing. Should probably change it to something else.

PR #2066, opened 2022-12-19
Fix feature data button refresh.

Creaticide:
Fixes featuredata button duplicating & error on click.


bundles/framework/featuredata2/plugin/FeaturedataPlugin.js
```diff
@@ -221,6 +221,20 @@ Oskari.clazz.define('Oskari.mapframework.bundle.featuredata2.plugin.FeaturedataP
...
+                 *
+                 * Calls flyouts layerAdded() method
+                 */
+                'AfterMapLayerAddEvent': function (event) {
+                    if (event.getMapLayer().hasFeatureData()) {
+                        this.refresh();
+                    }
+                },
+                'AfterMapLayerRemoveEvent': function (event) {
+                    this.refresh();
```
ZakarFin, line 237: Do we need to refresh here if layer doesn't have featuredata? Or could we add the same if here as in add?

PR #2067, opened 2022-12-20
Search plugin index.

Creaticide:
Search plugin order on UI.


ZakarFin:
^ Search is wider than the usual button-plugins so it looks nicer if it's on top of the page instead of between buttons.

PR #2070, opened 2022-12-23
Get logo from server.

Creaticide:


ZakarFin:
Related to oskariorg/oskari-server#901. Allows instance logo to be configured on the server and both the frontend and the server printing use that same image.

PR #2072, opened 2023-01-04
Mapstyle style editor

Creaticide:
Style editor + making theme work on zoom slider, feature data & toolbar button.

### DIFF
--- a/bundles/framework/featuredata2/plugin/FeatureDataButton.jsx
+++ b/bundles/framework/featuredata2/plugin/FeatureDataButton.jsx
@@ -1,72 +1,53 @@
 import React from 'react';
 import { Button, Tooltip } from 'oskari-ui';
 import styled from 'styled-components';
+import { ThemeConsumer } from '../../../../src/react/util';
+import { getNavigationTheme } from '../../../../src/react/theme';
 
 const StyledButton = styled(Button)`
     height: 32px;
     font-size: 14px;
     border: none;
-    border-radius: ${props => props.$rounding};
-    color: ${props => props.$active ? '#ffd400' : props.$theme.accent};
-    background: ${props => props.$theme.primary};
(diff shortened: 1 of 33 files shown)


bundles/framework/featuredata2/plugin/FeatureDataButton.jsx
```diff
@@ -1,72 +1,53 @@
 import React from 'react';
 import { Button, Tooltip } from 'oskari-ui';
 import styled from 'styled-components';
+import { ThemeConsumer } from '../../../../src/react/util';
+import { getNavigationTheme } from '../../../../src/react/theme';
```
ZakarFin, line 5:
```suggestion
import { ThemeConsumer } from 'oskari-ui/util';
import { getNavigationTheme } from 'oskari-ui/theme';
```

bundles/framework/featuredata2/plugin/FeaturedataPlugin.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { ThemeProvider } from '../../../../src/react/util';
```
ZakarFin, line 3:
```suggestion
import { ThemeProvider } from 'oskari-ui/util';
```

bundles/mapping/mapmodule/MapModuleButton.jsx
```diff
@@ -15,7 +15,6 @@ const Container = styled('div')`
 `;
 
 const StyledButton = styled(MapButton)`
-    opacity: 0.8;
```
ZakarFin, line 18: The current buttons are a bit transparent. So we should have this here. It has been moved to toolbar button instead. Should we have a MapButton with opacity if it doesn't have a toolbar and a toolbar one without opacity?

PR #2091, opened 2023-01-11
Set fixed font for xy button.

Creaticide:


ZakarFin:
So theme doesn't affect it as it's more of an icon than text for UI-purposes.

PR #2098, opened 2023-01-18
Coordinate tool fixes & refactoring.

Creaticide:
Fixes to coordinate transformation & some refactoring.

### DIFF
--- a/bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
+++ b/bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
@@ -33,7 +33,6 @@ class UIHandler extends StateHandler {
         this.decimalSeparator = Oskari.getDecimalSeparator();
         this.preciseTransform = Array.isArray(this.config.supportedProjections);
         this.updateLonLat(this.getMapXY(), true, true, true);
-        this.getEmergencyCallInfo(this.state.xy);
         this.popupListeners = [];
     };
 
@@ -82,21 +81,21 @@ class UIHandler extends StateHandler {
         this.updatePopup();
     }
 
-    setLon (value) {
+    setLonInputValue (value) {
         this.updateState({
             lonField: value
         });
         this.updatePopup();
     }
 
-    setLat (value) {
+    setLatInputValue (value) {
         this.updateState({
             latField: value
         });
         this.updatePopup();
     }
 
-    async blur () {
+    async useUserDefinedCoordinates () {
         let data = {
             lonlat: {
                 lon: this.state.lonField,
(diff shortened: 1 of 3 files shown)


bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -19,27 +17,30 @@ class UIHandler extends StateHandler {
...
+            showReverseGeoCode: false,
+            approxValue: false
         });
         this.popupControls = null;
         this.eventHandlers = this.createEventHandlers();
         this.decimalSeparator = Oskari.getDecimalSeparator();
         this.preciseTransform = Array.isArray(this.config.supportedProjections);
-        this.updateLonLat();
+        this.updateLonLat(this.getMapXY(), true, true, true);
         this.getEmergencyCallInfo(this.state.xy);
```
ZakarFin, line 36: updateLonLat() seems to call getEmergencyCallInfo() so we don't need another call here.

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -83,28 +84,29 @@ class UIHandler extends StateHandler {
...
-            displayXy: {
-                lonlat: {
-                    lat: value,
-                    lon: this.state.displayXy?.lonlat?.lon
-                }
-            }
+            latField: value
         });
         this.updatePopup();
     }
```
ZakarFin, line 97: It is most likely that both lon and lat are changed at the same time so we should optimize for that and combine the setter to a single one. Otherwise on most cases updatePopup() will be called twice when ever the coordinate changes.

ZakarFin: Or how does this compare to updateLonLat()?

ZakarFin: Ah, these are the changelisteners when user writes the coordinates on the field -> popup is refreshed to show the new input value? Maybe we should add some documentation comments here or rename to more descriptive names?

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -83,28 +84,29 @@ class UIHandler extends StateHandler {
...
 
+    async blur () {
```
ZakarFin, line 99: We should rename this from "when this is called" to something more descriptive like "what happens when this is called". Maybe `useUserDefinedCoordinates()`?

bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
Creaticide: Renamed to setXInputValue.

PR #2104, opened 2023-01-23
Side panel component & print panel in React

Creaticide:
-Reusable side panel component.
-Print panel converted to React.


src/react/components/window/SidePanel.jsx
```diff
@@ -0,0 +1,73 @@
...
+    width: 252px;
+    display: flex;
+    flex-direction: column;
+
+    div.header {
+        background-color: #FDF8D9;
+        padding: 5px 10px;
+        div.icon-close {
+            float: right;
+        }
```
ZakarFin, line 25: This can be removed as we are using the AntD CloseCircleFilled icon instead of the older png close icon.

src/react/components/window/SidePanel.jsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return (
+        <div className="header">
+            <FloatingIcon>
+                <Confirm
+                    disabled={!confirmExit}
+                    title={<Message messageKey='ContentEditorView.exitConfirm' bundleKey={BUNDLE_KEY} />}
+                    onConfirm={onClose}
+                    okText={<Message messageKey='ContentEditorView.buttons.yes' bundleKey={BUNDLE_KEY} />}
+                    cancelText={<Message messageKey='ContentEditorView.buttons.no' bundleKey={BUNDLE_KEY} />}>
```
ZakarFin, line 52: We shouldn't use ContentEditorView keys here (but for now confirmExit is always false anyway so this is dead code).

bundles/framework/printout/PrintoutHandler.js
```diff
@@ -0,0 +1,317 @@
...
+                <PrintoutPanel
+                    controller={this.controller}
+                    state={this.state}
+                    scaleSelection={this.instance.conf.scaleSelection}
+                    scaleOptions={this.scaleOptions}
+                    isTimeSeries={this.isTimeSeriesActive()}
+                />
+            );
+        }
+    }
```
ZakarFin, line 63: Try refactoring this in a way that there is no JSX used in the handler -> move it to instance and trigger calls to update with state change listeners provided by the StateHandler base class or something like that.

bundles/framework/printout/PrintoutHandler.js
```diff
@@ -0,0 +1,317 @@
...
+            baseLayer.getCurrentStyle().getName();
+        }
+
+        let url = Oskari.urls.getRoute('GetPrint') +
+            '&format=image/png' +
+            '&pageSize=' + pageSize +
+            '&resolution=' + map.getResolution() +
+            '&srs=' + map.getSrsName() +
+            '&coord=' + map.getX() + '_' + map.getY() +
+            '&mapLayers=' + mapLayers;
```
ZakarFin, line 239: You can pass params to getRoute as an object:
```suggestion
        let url = Oskari.urls.getRoute('GetPrint', {
         format: 'image/png',
         ...etc})
```

bundles/framework/printout/instance.js
```diff
@@ -180,6 +191,37 @@ Oskari.clazz.define('Oskari.mapframework.bundle.printout.PrintoutBundleInstance'
         getService: function () {
             return this.printService;
         },
+        showPanel: function () {
+            const controller = this.handler?.getController();
+            controller.showPanel(
```
ZakarFin, line 196: Optional chaining in `this.handler?.getController()` doesn't really make sense here since controller.showPanel() is called on the next line.

PR #2105, opened 2023-01-24
Coordinates to print

Creaticide:
Additional coordinate options to print panel.

Based on https://github.com/oskariorg/oskari-frontend/pull/2104

### DIFF
--- a/bundles/framework/printout/PrintoutHandler.js
+++ b/bundles/framework/printout/PrintoutHandler.js
@@ -19,6 +19,7 @@ class UIHandler extends StateHandler {
             scale: null,
             showTimeSeriesDate: false,
             isMapStateChanged: false,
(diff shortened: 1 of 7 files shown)


ZakarFin:
This look good for merging, but it needs to wait for the server side implementation to be done.

bundles/framework/printout/PrintoutHandler.js
```diff
@@ -18,7 +18,10 @@ class UIHandler extends StateHandler {
             scaleType: 'map',
             scale: null,
             showTimeSeriesDate: false,
-            isMapStateChanged: false
+            isMapStateChanged: false,
+            showCoordinates: false,
```
ZakarFin, line 22:
```suggestion
            // impl for toggling coordinates is commented out in JSX. This will be enabled on future release.
            showCoordinates: false,
```

PR #2126, opened 2023-02-14
Enable user to select search channels

ZakarFin:
This adds a configuration button for search on map that allows users to select which data sources are used for searches. This requires explicit config for the SearchPlugin with:
```
{
  id: 'Oskari.mapframework.bundle.mapmodule.plugin.SearchPlugin',
  config: {
     allowOptions: true
  }
}
```

Works nicely with WFS-search channels, but needs more testing for merging. Also we might want to disable the choice for the user by configuration.

Selecting channels:
image

Toggles for non-default channels are colored differently and the "options" button is highlighted when search is done using non-default channels:
image


ZakarFin:
With non-wfs search channels if currently looks like this: 
image

Tooltips are empty and everychannel is "default" channel (even the ones not in use). Also translations should be added since there is no UI for these.

PR #2130, opened 2023-02-16
Map legend rewrite on publisher

Creaticide:
Map legend tool in React + moved to map layers panel.

### DIFF
--- a/bundles/framework/maplegend/publisher/MapLegendHandler.js
+++ b/bundles/framework/maplegend/publisher/MapLegendHandler.js
@@ -9,6 +9,7 @@ class UIHandler extends StateHandler {
         this.setState({
             showLegends: initialData || false
         });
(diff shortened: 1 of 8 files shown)


bundles/framework/maplegend/publisher/MapLegendHandler.js
```diff
@@ -0,0 +1,68 @@
...
+
+    setShowLegends (value) {
+        this.updateState({
+            showLegends: value
+        });
+        if (value === false) {
+            this.tool.setEnabled(false);
+        } else {
+            this.tool.setEnabled(true);
+        }
```
ZakarFin, line 31: This feels wrong having the state in two places. I'm not sure if we could do something about it since it's bound to go out of sync at some point. We could atleast lose the if and just call `this.tool.setEnabled(!!value);`

PR #2138, opened 2023-02-23
Layer visibility info

okauppinen:
Gather visibilityInfo to layer on visibility change and check unsupported reason on start. Selected layer can use whole visibleInfo and layerlist can use unsupported on render (no need to check or get values). 

layer has methods:
- isVisible() => user 
- isVisibleOnMap() => maplayer is visible on map (visible, geom, scale, supported, opacity != 0)
- const {unsupported} = layer.getVisibleInfo() => most severe UnsupportedReason

Changes:
- update visibility info and unsupported reason on layers plugin. remove override from 3D/cesium plugin
- selected layers uses visibilityInfo
- layerlist uses unsupported to show most severe unsupported reason
- mapfull gather params uses isVisibleOnMap if optimized
- publisher shows most severe unsupported reason (earlier listed all unsupported reasons)
- deprecate mapmodule isLayerVisible


ZakarFin:
Looks good to me

PR #2139, opened 2023-02-27
Thematic map controls in react

Creaticide:
Thematic map controls to react + adding drag 'n drop.

### DIFF
--- a/bundles/statistics/statsgrid2016/plugin/TogglePlugin.js
+++ b/bundles/statistics/statsgrid2016/plugin/TogglePlugin.js
@@ -1,3 +1,9 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { PluginHandler } from "../handler/PluginHandler";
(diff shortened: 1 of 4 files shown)


bundles/statistics/statsgrid2016/plugin/TogglePlugin.js
```diff
@@ -6,75 +12,79 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.TogglePlugin', function (flyout
...
-        } else {
-            tools.removeClass('tool-drag-disabled');
-        }
-    },
+    _setLayerToolsEditModeImpl: function () {},
     getElement: function () {
         return this.element;
     },
+    hasUI: function () {
+        this.changeToolStyle();
```
ZakarFin, line 30: hasUI() should return a boolean value. Publisher uses this to check if the plugin should be shutdown when entering publisher.

PR #2141, opened 2023-03-01
Rpc panel to publisher

Creaticide:
New rpc panel with metadata & feedback tools.


ZakarFin:
We need the tools presented in this panel to be added by the corresponding tools. That is the reason why these tools were under metadata catalogue and feedback service. When they are provided by non-publisher bundles the user will only see these choices if those bundles are used by the Oskari application. If we bundle them inside the publisher the user will be presented with choices that might not be supported by the Oskari instance.

bundles/catalogue/metadatacatalogue/publisher/MetaDataSearchPublisherTool.jsx
```diff
@@ -0,0 +1,20 @@
...
+    }
+`;
+
+export const MetadataSearchPublisherTool = ({ state, controller }) => {
+    return (
+        <StyledCheckbox
+            checked={state.allowMetadata}
+            onChange={(e) => controller.setAllowMetadata(e.target.checked)}
+        >
+            <Message bundleKey='Publisher2' messageKey='BasicView.maptools.MetadataSearchTool' />
```
ZakarFin, line 17: We should move this localization from Publisher2 locales to metadatasearch locale while we are at it.

bundles/catalogue/metadatacatalogue/publisher/MetadataSearchTool.js
```diff
@@ -50,13 +59,11 @@ Oskari.clazz.define('Oskari.mapframework.publisher.tool.MetadataSearchTool',
         * @returns {Object} tool value object
         */
         getValues: function () {
-            const me = this;
-
-            if (me.state.enabled) {
-                var json = {
+            if (this.handler?.getState()?.allowMetadata) {
+                let json = {
```
ZakarFin, line 63: `const` should work here too since the object is modified without reassigning the variable itself.

bundles/framework/feedbackService/publisher/FeedbackServicePublisherTool.jsx
```diff
@@ -0,0 +1,77 @@
...
+    flex-direction: column;
+`;
+
+const StyledCheckbox = styled(Checkbox)`
+    + .ant-checkbox-wrapper {
+        margin-left: 0;
+    }
+`;
+
+export const FeedbackServicePublisherTool = ({ state, controller }) => {
```
ZakarFin, line 21: Maybe rename this to FeedbackServiceForm or similar to make it easier to recognize the React-component for the form from the actual publisher tool js-file.

Also add a comment why it's using both `useState()` AND `controller.updateField()` with something like: 
```
controller.updateField() will re-render the form which makes the field that is being edited loose focus after each keypress -> using useState() to workaround this while waiting for the whole publisher to be written in React.
 ```

PR #2155, opened 2023-03-07
Change tooltip container only for map layers.

Creaticide:


ZakarFin:
Turned out the generic component change broke tooltips in more places than it fixed (it created scrollbars for smaller containers with elements that have tooltips for example dropdown options):
image

This reverts #2132 and fixes the issue with dragging selected layers whe tooltips tend to stay on screen after the flyout is closed.

PR #2156, opened 2023-03-07
Missing layer tool drag handles.

Creaticide:
Works for map legend tool, missing for layer selection.

### DIFF
--- a/bundles/framework/publisher2/view/PanelToolLayout.js
+++ b/bundles/framework/publisher2/view/PanelToolLayout.js
@@ -159,7 +159,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelToolLayout'
             var me = this;
             this.tools.forEach(tool => {
                 // don't call for tools that already have been set enabled (=plugin has already been created.)
-                if (tool.isDisplayed(me.data) && !tool.isShownInToolsPanel() && !tool.state.enabled) {
+                if (tool.isDisplayed(me.data) && !tool.isShownInToolsPanel() && (tool.state && !tool.state.enabled)) {
                     tool.setEnabled(true);
                 }
             });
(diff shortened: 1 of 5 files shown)


bundles/framework/publisher2/view/PanelToolLayout.js
```diff
@@ -386,7 +386,13 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelToolLayout'
             if (!elem || this._addedDraggables.includes(elem)) {
                 return;
             }
-            const enabled = tool.state.enabled === true;
+            let enabled = false;
+            if (tool.state) {
+                enabled = tool.state.enabled === true;
+            } else if (typeof tool.handler !== 'undefined') {
+                enabled = true;
+            }
```
ZakarFin, line 394: AbstractPluginTool has isEnabled() that we could use here to make it less hacky: 

https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/publisher2/tools/AbstractPluginTool.js#L119-L121

This PR also adds isEnabled() to MapLegend.js LayerTool so we could just implement isEnabled() for other LayerTools and use isEnabled() calls here to get rid of this if/else.

bundles/framework/publisher2/view/PanelToolLayout.js
```diff
@@ -159,7 +159,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PanelToolLayout'
             var me = this;
             this.tools.forEach(tool => {
                 // don't call for tools that already have been set enabled (=plugin has already been created.)
-                if (tool.isDisplayed(me.data) && !tool.isShownInToolsPanel() && !tool.state.enabled) {
+                if (tool.isDisplayed(me.data) && !tool.isShownInToolsPanel() && (tool.state && !tool.state.enabled)) {
                     tool.setEnabled(true);
                 }
```
ZakarFin, line 164: Maybe we should try if we can just remove this loop completely. It feels like it's used to fix something that is out of sync by calling setEnabled() for tools that are already "enabled". Let's fix the broken tools instead and get rid of this loop.

PR #2165, opened 2023-03-10
Upgrade to AntD 4.24.8.

Creaticide:
Upgrade to AntD 4.24.8.
- Switched moment.js to day.js
- Tabs.TabPane -> items 
- Menu.Item -> items
- Use dropdownRender on Dropdown


ZakarFin:
Also momentjs -> dayjs

PR #2170, opened 2023-03-13
Vector style tab

okauppinen:


bundles/mapping/userstyle/resources/locale/fi.js
```diff
@@ -6,11 +6,33 @@ Oskari.registerLocalization(
...
+        },
+        "success": {
+            "post": "Oma tyyli tallennettu",
+            "put": "Oma tyyli päivitetty",
+            "delete": "Oma tyyli poisttettu"
+        },
+        "error": {
+            "get": "Omien tyylien haku epäonnistui",
+            "post": "Oman tyylin tallennus epäonnistui",
+            "put": "Oman tyylin päivitys epäonnisui",
```
ZakarFin, line 34:
```suggestion
            "put": "Oman tyylin päivitys epäonnistui",
```
Or maybe we could use the one from post for put as well to reduce translations?

PR #2172, opened 2023-03-13
DescribeLayer parse vector styles

okauppinen:
- Moved DescribeLayer handling to map.layer.handler. Handles adding layer and has access to state and service.
- Moved DescribeLayer to maplayer service
-> layer is added and event fired after DescribeLayer
- Added describeLayerStatus to AbstractLayer
- Added handleDescribeLayer to AbstractLayer and override in AbstractVectorLayer to set styles from info
- Added handleDescribeLayer and impl to layer plugins (abstract and draft for wmts)
- Added handleDescribeLayer to LayersPlugin which handles visibilityInfo and parses coverage

Not sure if getDescribeLayerStatus is really needed as handleDescribeLayer functions should do everything that is needed before adding layer to map. Could be also {layerId: loaded} in maplayer state/service.
=> removed userstyle would use status to figure out if style has to be removed from layer. if describelayer isn't loaded, nothing have to do (updated stuff will be loaded before use)

VectorStyle extends Style which causes:
- id => name
- name => title
- abstractlayer style selection assumes that all styles are available on create layer
- selectStyle (name -> first -> create empty)
- getcurrentstyle -> creates empty VectorStyle on create layer
=> selectStyle stores requested id/name, describelayer sets styles (doesn't use addstyle)
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapmodule/domain/AbstractLayer.js#L704-L748
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapmodule/service/map.layer.js#L1398
https://github.com/oskariorg/oskari-frontend/blob/master/bundles/framework/mapfull/instance.js#L324

TODO:
- ~have to check if selected wfs layers are handled properly on startup~
- analysis setStylesFromOptions

UPDATE (accidently pushed changes for this branch):
- refactored toggle editor/style list
- changed to store added/updated style from response instead of fetching all styles. 
- For now storing and adding style to actual layer works same way to guest (sync) and registered user (async). 
- bug fixes: storeStyle -> !layerId, deleteStyle id was missing

Requires: https://github.com/oskariorg/oskari-server/pull/942


ZakarFin:
We should try to call DescribeLayer before `AfterMapLayerAddEvent` is sent so we can minimize compatibility issues. Most functionality check what is inside the OskariLayer object only after it has been added to the map and if we can load DescribeLayer before that the other functionalities could stay unchanged. But as to where to call it, I'm not sure. I was thinking we could have some flag on OskariLayer to tell functionalities if it has all the data like you track the status here for each layer. Loading and adding that information could be in MapLayerService perhaps? A function that would take a layer id and a callback that would be called with the filled in OskariLayer maybe? Then the data could be stored on the layer and callback could be called right away if the layer was already filled with DescribeLayer data.

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -1262,19 +1264,31 @@ Oskari.clazz.define(
...
             return this.getOpacity() !== 0 && !unsupported &&
                 Object.values(booleans).every(b => b === true);
+        },
+        getDescribeLayerStatus: function () {
+            return this._describeLayerStatus;
+        },
+        setDescribeLayerStatus: function (status) {
+            this._describeLayerStatus = status;
+        },
+        requiresDescripeLayer: function () {
```
ZakarFin, line 1287:
```suggestion
        requiresDescribeLayer: function () {
```

bundles/mapping/mapmodule/request/map.layer.handler.js
```diff
@@ -64,6 +67,41 @@ Oskari.clazz.define('map.layer.handler',
...
+            }
+            const layerId = layer.getId();
+            if (isNaN(layerId)) {
+                // only layers that have numeric ids can have reasonable response for DescribeLayer
+                addLayer();
+                return;
+            }
+            const status = layer.getDescribeLayerStatus();
+            if (status === DESCRIBE_LAYER.LOADED || status === DESCRIBE_LAYER.PENDING) {
+                return;
```
ZakarFin, line 85: Shouldn't we call addLayer() when `status === LOADED` since there isn't any pending calls waiting for network call for DescribeLayer  to finish? And only return when `status === PENDING` since the handler that is waiting on the describe will call addLayer once the info has been loaded?

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -1262,19 +1264,31 @@ Oskari.clazz.define(
...
+            return false;
+        },
+        handleDescribeLayer: function (info) {
```
ZakarFin, line 1290: I guess this could replace the modelbuilders that we have for layers currently 🤔 Need to think if we should just get rid of one or the other.

bundles/mapping/mapmodule/request/map.layer.handler.js
```diff
@@ -64,6 +67,41 @@ Oskari.clazz.define('map.layer.handler',
...
+                    if (layer.requiresDescripeLayer()) {
+                        // notify user here or in implementations
+                        this.log.error('Attempt to add layer that requires more info. Skipping id: ' + layerId);
+                        return;
+                    }
+                }
+                layer.getDescribeLayerStatus(DESCRIBE_LAYER.LOADED);
+                const sandbox = this.layerService.getSandbox();
+                layer.handleDescribeLayer(info);
+                sandbox.findRegisteredModuleInstance('MainMapModule').handleDescribeLayer(layer, info);
```
ZakarFin, line 100: I think we should move `layer.handleDescribeLayer(info)` to `mapmodule.handleDescribeLayer()` so all the setting of it happens in one place. Or maybe we should set the info to layer here and not pass it to mapmodule with the layer 🤔 

I think this handler class is created in mapmodule so we could also pass a reference to mapmodule in constructor instead of getting it through sandbox, but I'll tinker with it a bit to see what feels best.

bundles/mapping/mapwmts/domain/WmtsLayer.js
```diff
@@ -6,6 +6,10 @@
 Oskari.clazz.define('Oskari.mapframework.wmts.domain.WmtsLayer', function () {
     this._layerType = 'WMTS';
 }, {
+    requiresDescripeLayer: function () {
```
ZakarFin, line 9:
```suggestion
    requiresDescribeLayer: function () {
```

bundles/mapping/mapwmts/plugin/WmtsLayerPlugin.ol.js
```diff
@@ -58,6 +68,8 @@ Oskari.clazz.define('Oskari.mapframework.wmts.mapmodule.plugin.WmtsLayerPlugin',
             if (!this.isLayerSupported(layer)) {
                 return;
             }
+            // TODO: after DescripeLayer action route returns capa info
```
ZakarFin, line 71:
```suggestion
            // TODO: after DescribeLayer action route returns capa info
```

bundles/mapping/mapmodule/domain/AbstractLayer.js
okauppinen: fixed

bundles/mapping/mapmodule/request/map.layer.handler.js
```diff
@@ -64,6 +67,41 @@ Oskari.clazz.define('map.layer.handler',
...
+            }
+            const status = layer.getDescribeLayerStatus();
+            if (status === DESCRIBE_LAYER.LOADED || status === DESCRIBE_LAYER.PENDING) {
+                return;
+            }
+            layer.getDescribeLayerStatus(DESCRIBE_LAYER.PENDING);
+            this.layerService.getDescribeLayer(layerId, info => {
+                if (!info) {
+                    layer.setDescribeLayerStatus(DESCRIBE_LAYER.ERROR);
+                    if (layer.requiresDescripeLayer()) {
```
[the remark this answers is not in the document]
okauppinen: fixed

bundles/mapping/mapwmts/domain/WmtsLayer.js
okauppinen: fixed

bundles/mapping/mapwmts/plugin/WmtsLayerPlugin.ol.js
okauppinen: made changes to another branch

bundles/mapping/mapmodule/request/map.layer.handler.js
okauppinen: fixed

bundles/mapping/mapmodule/domain/AbstractLayer.js
okauppinen: Didn't have good feelings how I solved describe layer info handling. Had to add handling in too many places. Layer, builder and plugins may all require info. One solution is to remove builders since them doesn't have much work to do nowadays.

PR #2188, opened 2023-03-20
Admin user control in React.

Creaticide:
Admin user & roles control flyout content written in React.


bundles/framework/admin-users/view/RolesTab.jsx
```diff
@@ -0,0 +1,81 @@
...
+    align-items: center;
+`;
+
+const AddButton = styled(PrimaryButton)`
+    width: 50px;
+    margin-left: 15px;
+    align-self: flex-end;
+`;
+
+const DeleteButton = styled(Button)`
```
ZakarFin, line 26: We have a delete button with confirm wrapping in oskari-ui buttons already. We should use that instead since it's nice to have a confirm on delete and it adds a lot of noise that we don't need here if we use the common component.

bundles/framework/admin-users/view/UserForm.jsx
```diff
@@ -0,0 +1,154 @@
...
+                />
+            </LabelledField>
+            <Buttons>
+                <RightButtons>
+                    <SaveButton
+                        type='save'
+                        onClick={() => controller.saveUser()}
+                    />
+                    {(state.editingUserId && !isExternal) && (
+                        <Confirm
```
ZakarFin, line 135: Let's use the common delete button in oskari-ui here as well if there's not some reason that we can't.

bundles/framework/admin-users/view/UsersTab.jsx
```diff
@@ -0,0 +1,110 @@
...
+`;
+
+
+export const UsersTab = ({ state, controller, isExternal }) => {
+    const [filter, setFilter] = useState('');
+    const hasFilter = state.userFilter !== null && state.userFilter !== undefined && state.userFilter.length > 0;
+    let users = !hasFilter ? state.users : state.users.filter(user => 
+        (user.firstName && user.firstName.toLowerCase().indexOf(state.userFilter.toLowerCase()) > -1) ||
+        (user.lastName && user.lastName.toLowerCase().indexOf(state.userFilter.toLowerCase()) > -1) ||
+        (user.user && user.user.toLowerCase().indexOf(state.userFilter.toLowerCase()) > -1));
```
ZakarFin, line 59: We already have a controller/handler. Let's refactor this so the filtering is inside the handler.

bundles/framework/admin-users/view/UsersTab.jsx
```diff
@@ -0,0 +1,110 @@
...
+                    value={filter}
+                    onChange={(e) => setFilter(e.target.value)}
+                />
+                <PrimaryButton
+                    type='search'
+                    onClick={() => controller.setUserFilter(filter)}
+                />
+            </SearchContainer>
+            {!isExternal && (
+                <AddButton
```
ZakarFin, line 79: We have an add button in oskari-ui. Could we use that instead?

bundles/framework/admin-users/view/UsersTab.jsx
```diff
@@ -0,0 +1,110 @@
...
+                            <span>{user.user} ({user.firstName} {user.lastName})</span>
+                            <EditButton
+                                onClick={() => controller.setEditingUserId(user.id)}
+                            >
+                                <EditOutlined />
+                            </EditButton>
+                        </>
+                    )}
+                </UserBlock>
+            ))}
```
ZakarFin, line 107: Maybe we should have the listing in another component/view and the editing "mode" in another? Or is the editing done directly on the listing? I think I need to take a look at the UI that this creates first :)

PR #2209, opened 2023-03-28
Fix announcements form crashing and styles.

Creaticide:
Fixes form crashing when editing dates and styling on small window sizes.


bundles/framework/announcements/view/AnnouncementsBanner.jsx
```diff
@@ -65,21 +75,25 @@ const getContent = (state, controller, renderDescriptionPopup) => {
     }
     return (
         <Fragment>
```
ZakarFin, line 77: Fragment here isn't really needed since Row is now the root tag for this.

PR #2213, opened 2023-04-03
Timeseries slider component

Creaticide:
New slider component for timeseries, currently works for year mode & year range mode.


ZakarFin:
The fields on range mode go a bit off from the container
image

PR #2220, opened 2023-04-06
Use theme for timeseries slider + styling changes.

Creaticide:
Use theme, styling + added markers on the rail.

### DIFF
--- a/src/react/components/TimeSeries/TimeSeriesSlider.jsx
+++ b/src/react/components/TimeSeries/TimeSeriesSlider.jsx
@@ -1,7 +1,9 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { playbackSpeedOptions, sliderTypes, timeUnits } from './util/constants';
+import { sliderTypes, timeUnits } from './util/constants';
 import { getDifferenceCalculator, calculateSvgX } from './util/calculation';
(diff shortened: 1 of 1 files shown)


src/react/components/TimeSeries/TimeSeriesSlider.jsx
```diff
@@ -1,21 +1,24 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { playbackSpeedOptions, sliderTypes, timeUnits } from './util/constants';
+import { sliderTypes, timeUnits } from './util/constants';
 import { getDifferenceCalculator, calculateSvgX } from './util/calculation';
+import { getHeaderTheme } from 'oskari-ui/theme/ThemeHelper';
```
ZakarFin, line 5: I think we want to use the `getNavigationTheme()` instead as it it uses the button colors that matches the current colors on this controls more closely than the "popup colors" that `getHeaderTheme()` is used for.

ZakarFin:
You can use the same kind of code as here https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/mapping/mapmodule/MapModuleButton.jsx#L43-L45 to pass the theme:

```
    const mapModule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
    return (
        <ThemeProvider value={mapModule.getMapTheme()}>
         ...
```

PR #2232, opened 2023-04-25
Disable system role editing.

Creaticide:
Hide guest role from admin list & hide delete button for other system roles.

### DIFF
--- a/bundles/framework/admin-users/view/RolesTab.jsx
+++ b/bundles/framework/admin-users/view/RolesTab.jsx
@@ -1,7 +1,8 @@
 import React from 'react';
-import { TextInput, Confirm, Message } from 'oskari-ui';
-import { PrimaryButton, DeleteButton } from 'oskari-ui/components/buttons';
+import { TextInput, Message, Button } from 'oskari-ui';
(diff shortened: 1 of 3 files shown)


bundles/framework/admin-users/view/RolesTab.jsx
```diff
@@ -57,12 +63,43 @@ export const RolesTab = ({ state, controller }) => {
...
+                    {state.editingRole && state.editingRole.id === role.id ? (
+                        <StyledInput
+                            value={state.editingRole.name}
+                            onChange={(e) => controller.updateEditingRole(e.target.value)}
+                            status={state.editingRoleError ? 'error' : null}
+                        />
+                    ) : (
+                        <span>{role.name}</span>
+                    )}
+                    {state.systemRoles.findIndex(r => r === role.name) < 0 && (
```
ZakarFin, line 75: Can we use includes() here? 
```suggestion
                    {!state.systemRoles.includes(r => r === role.name) && (
```
Or maybe even inject a boolean property to roles on the list so we could just check some role.editableRole here?

PR #2243, opened 2023-05-02
Map link popup in React.

Creaticide:


bundles/mapping/toolbar/view/MapLinkPopup.jsx
```diff
@@ -0,0 +1,125 @@
...
+        </StyledContent>
+    );
+};
+
+export const showMapLinkPopup = (guidedTour, baseUrl, onClose) => {
+    const dimensions = getNavigationDimensions();
+    let placement = PLACEMENTS.BL;
+    if (dimensions?.placement === 'right') {
+        placement = PLACEMENTS.BR;
+    }
```
ZakarFin, line 110: We can always show this in the middle of the screen.

PR #2244, opened 2023-05-03
Featuredata conversion to react feature branch

DenverCoder544:


bundles/framework/featuredata/plugin/FeatureDataFlyout.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { FeatureDataContainer } from '../view/FeatureDataContainer';
+import { showFlyout } from '../../../../src/react/components/window';
```
ZakarFin, line 3:
```suggestion
import { showFlyout } from 'oskari-ui/window';
```

bundles/framework/featuredata/plugin/FeaturedataPlugin.js
```diff
@@ -0,0 +1,149 @@
...
+        getMapStatusChanged: function () {
+            return this._mapStatusChanged;
+        },
+        /**
+         * @method refresh
+         * Updates the plugins interface (hides if no featuredata layer selected)
+         */
+        refresh: function () {
+            if (!this.handler) {
+                this.handler = new FeatureDataPluginHandler(this, this.getMapModule(), this.getConfig());
```
ZakarFin, line 67: Nothing changes `this._flyoutOpen` here so the button never has "active" color. We could add a state listener here for it to make it easier to read when the renderButton() is called:
```
this.handler.addStateListener(state => {
  // toggle  _flyoutOpen and/or loading
  // based on state and call renderButton() 
})
```

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -0,0 +1,72 @@
...
+    updateStateAfterMapEvent () {
+        const featureDataLayers = this.getFeatureDataLayers() || [];
+        const layerId = featureDataLayers?.find(layer => layer.getId() === this.state.activeLayerId) ? this.state.activeLayerId : featureDataLayers.map(layer => layer.getId())[0] || null;
+        const features = layerId ? this.getFeaturesByLayerId(layerId) : [];
+        this.updateState({
+            activeLayerId: layerId,
+            layers: featureDataLayers,
+            activeLayerFeatures: features
+        });
+    }
```
ZakarFin, line 42: The one-liner is a bit hard to read, but this is the intention?
```suggestion
    updateStateAfterMapEvent () {
        const featureDataLayers = this.getFeatureDataLayers() || [];
        // make sure our active layer in state is still valid
        let currentLayer = featureDataLayers.find(layer => layer.getId() === this.state.activeLayerId);
        if (!currentLayer && featureDataLayers.length) {
            // layer removed/need to update active layer to a new one
            currentLayer = featureDataLayers[0];
        }
        const features = currentLayer ? this.getFeaturesByLayerId(currentLayer.getId()) : [];
        this.updateState({
            activeLayerId: currentLayer.getId(),
            layers: featureDataLayers,
            activeLayerFeatures: features
        });
    }
```

bundles/framework/featuredata/plugin/FeaturedataPlugin.js
```diff
@@ -0,0 +1,149 @@
...
+                        visible={this._hasFeaturedataLayers()}
+                        icon={<Message messageKey='title' bundleKey='FeatureData'/>}
+                        onClick={() => this.handler.openFlyout()}
+                        active={this._flyoutOpen}
+                        loading={loading}
+                        position={this.getLocation()}
+                    />
+                </ThemeProvider>,
+                el[0]
+            );
```
ZakarFin, line 93:
```suggestion
            const { layers } = this.handler.getState();
            ReactDOM.render(
                <ThemeProvider value={this.getMapModule().getMapTheme()}>
                    <FeatureDataButton
                        visible={layers.length > 0}
                        icon={<Message messageKey='title' bundleKey='FeatureData'/>}
                        onClick={() => this.handler.openFlyout()}
                        active={this._flyoutOpen}
                        loading={loading}
                        position={this.getLocation()}
                    />
                </ThemeProvider>,
                el[0]
            );
```

bundles/framework/featuredata/plugin/FeaturedataPlugin.js
```diff
@@ -0,0 +1,149 @@
...
+                        this.refresh();
+                    }
+                },
+                AfterMapLayerRemoveEvent: function (event) {
+                    if (event.getMapLayer().hasFeatureData()) {
+                        this.refresh();
+                    }
+                },
+                MapLayerVisibilityChangedEvent: function () {
+                    this.refresh();
```
ZakarFin, line 138: I would just call `this.handler.updateStateAfterMapEvent();` directly here since that should be the only thing we need. If there's no longer layers with featuredata -> state will be updated -> state change should trigger renderButton().

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -0,0 +1,72 @@
...
+    openFlyout () {
+        if (this.flyoutController) {
+            this.closeFlyout();
+            return;
+        }
+        const layers = this.getFeatureDataLayers() || null;
+        const activeLayerId = layers && layers.length ? layers[0].getId() : null;
+        const activeLayerFeatures = activeLayerId ? this.getFeaturesByLayerId(activeLayerId) : null;
+        this.state = { layers, activeLayerId, activeLayerFeatures };
+        this.flyoutController = showFeatureDataFlyout(this.state, this.getController());
```
ZakarFin, line 53: If we don't preload features to state on every map move even if the flyout is closed we should get them here. 

See if this kind of code works ok. It might not since the last line assumes that the updateState() has added the features BUT since in constructor the state changes trigger a call to updateFlyout() it might be ok :)

```suggestion
        const { layers, activeLayerId, activeLayerFeatures } = this.getState();
        if (!activeLayerFeatures) {
            // not empty features, but missing completely
            // empty should mean there is no features on viewport to list
            this.updateState({
                activeLayerFeatures: this.getFeaturesByLayerId(activeLayerId)
            });
        }
        this.flyoutController = showFeatureDataFlyout(this.getState(), this.getController());
```

bundles/framework/featuredata/plugin/FeaturedataPlugin.js
DenverCoder544: Removed the old (copy&paste)stuff concerning flyoutOpen and set a flag in state instead.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -0,0 +1,72 @@
+import { StateHandler, controllerMixin } from 'oskari-ui/util';
+import { showFeatureDataFlyout } from './FeatureDataFlyout';
+
+class FeatureDataPluginUIHandler extends StateHandler {
+    constructor (plugin, mapModule, config) {
```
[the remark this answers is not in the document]
DenverCoder544: Neither was config. Removed both.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
DenverCoder544: Refactored the oneliner to a more human-readable form

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -0,0 +1,72 @@
...
+        this.updateState({
+            activeLayerId: layerId,
+            activeLayerFeatures: features
+        });
+    }
+
+    updateStateAfterMapEvent () {
+        const featureDataLayers = this.getFeatureDataLayers() || [];
+        const layerId = featureDataLayers?.find(layer => layer.getId() === this.state.activeLayerId) ? this.state.activeLayerId : featureDataLayers.map(layer => layer.getId())[0] || null;
+        const features = layerId ? this.getFeaturesByLayerId(layerId) : [];
```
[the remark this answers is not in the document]
DenverCoder544: You're right. I refactored this a bit (a bunch) and now the features are updated only when needed.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
DenverCoder544: Didn't work out of the box but I implemented the basic idea like this.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -0,0 +1,104 @@
...
+    updateStateAfterMapEvent () {
+        const featureDataLayers = this.getFeatureDataLayers() || [];
+        if (!featureDataLayers || !featureDataLayers.length) {
+            this.closeFlyout();
+            return;
+        }
+
+        const activeLayerId = this.determineActiveLayerId(featureDataLayers);
+        let activeLayerFeatures = null;
+        if (activeLayerId && this.state.flyoutOpen) {
```
ZakarFin, line 47:
```suggestion
        if (activeLayerId && this.getState().flyoutOpen) {
```

ZakarFin:
This will not be Ready(tm)/completed for the next release, but we can still merge it to develop since this is a parallel implementation of the current "featuredata2" bundle. This helps with development and testing as it can be used next to the current impl, but the sample/template app won't be migrated to use this yet.

PR #2252, opened 2023-05-05
Copy to clipboard components.

Creaticide:
Style fix + common components for "copy to clipboard" situations.


bundles/framework/mydata/resources/locale/en.js
```diff
@@ -93,8 +93,7 @@ Oskari.registerLocalization(
                 },
                 "published": {
                     "desc": "Copy the HTML code below to your html code and the map will be visible on your website:",
-                    "copy": "Copy to the clipboard",
-                    "copied": "Copied"
+                    "copy": "Copy to clipboard"
```
ZakarFin, line 96: Do we still need these under mydata when these are in oskariui localizations?

PR #2262, opened 2023-05-10
Featuredata react conversion loading indicator

DenverCoder544:


bundles/framework/featuredata/view/TabStatusIndicator.jsx
```diff
@@ -0,0 +1,42 @@
...
+    padding-left: 1.5em;
+`;
+
+export const TabTitle = styled('div')`
+    color: ${theme.getTextColor()}
+`;
+
+export const TabLoadingTitle = (props) => {
+    const { layer } = props;
+    console.log('theme ', theme);
```
ZakarFin, line 23:
```suggestion
```

bundles/framework/featuredata/view/FeatureDataContainer.jsx
```diff
@@ -93,11 +95,19 @@ const createDatasourceFromFeatures = (features) => {
...
-const createLayerTabs = (layerId, layers, features, selectedFeatureIds, showSelectedFirst, sorting, controller) => {
+const createLayerTabs = (layerId, layers, features, selectedFeatureIds, showSelectedFirst, sorting, loadingStatus, controller) => {
     const tabs = layers?.map(layer => {
+        const status = loadingStatus[layer.getId()];
+        let title;
+        if (!status) {
+            title = <TabTitle>{layer.getName()}</TabTitle>;
+        } else {
+            title = status === FEATUREDATA_WFS_STATUS.error ? <TabErrorTitle>{layer.getName()}</TabErrorTitle> : <TabLoadingTitle layer={layer}/>;
+        }
```
ZakarFin, line 106: We could clean this up a bit if we just passed loading status to a Title component and it would decide how to render the title. Basically moving this if/else if /else structure inside a single component.

ZakarFin: The TabLoadingTitle also uses only the name of the layer so we could just pass the name to the Title component for all of these cases OR reference it by using the chidlren prop.

PR #2265, opened 2023-05-12
Change d3 to imports.

Creaticide:


bundles/framework/layerlist/model/Scale.js
```diff
@@ -1,3 +1,5 @@
+import * as d3 from 'd3';
```
ZakarFin, line 1: We should just remove d3 from layerlist. It's only used to calculate throttling (for search) based on how long the query is.

ZakarFin: Then again it's used on VectorLayerPlugin as well to calculate some colors. If we need to have it always present, we can use it here as well. I'll take a closer look at this.

ZakarFin:
Let's merge this for now as it is already a better situation than before and figure out later where we want to use D3 (namely can layerlist and VectorLayerPlugin work without it or should we embrace it as part of the "core libraries" like React and jQuery).

PR #2274, opened 2023-05-19
Featuredata react conversion select features by properties

DenverCoder544:


bundles/framework/featuredata/view/SelectByProperties.jsx
```diff
@@ -0,0 +1,266 @@
...
+const StyledButton = styled(Button)`
+    margin: 0 .5em 0 .5em
+`;
+
+const Buttons = (props) => {
+    const { closePopup, removeFilter, applyFilters, hasErrors } = props;
+    return <ButtonsContainer>
+        <StyledButton onClick={() => { closePopup(); }}>
+            <Message bundleKey={FEATUREDATA_BUNDLE_ID} messageKey={'selectByPropertiesPopup.buttons.cancel'}/>
+        </StyledButton>
```
ZakarFin, line 91: Try replacing this with https://github.com/oskariorg/oskari-frontend/blob/develop/src/react/components/buttons/SecondaryButton.jsx with type `cancel` so we don't need to have the localization in multiple places.

Maybe we could even replace the clearAll with `<SecondaryButton type='reset' />`?

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -152,12 +159,28 @@ class FeatureDataPluginUIHandler extends StateHandler {
...
 
+    openSelectByPropertiesPopup () {
+        if (this.popupController) {
+            this.closePopup();
+            return;
+        }
+        this.popupController = showSelectByPropertiesPopup(this.getState(), this.getController());
+    }
+
+    closePopup () {
```
ZakarFin, line 177: I'm not sure if there will be other popups here, but we should rename this to match the opener-function/something like closeSelectByPropertiesPopup() to make it immediately clear what popups it is closing.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -170,18 +193,148 @@ class FeatureDataPluginUIHandler extends StateHandler {
...
         if (this.flyoutController) {
             this.flyoutController.close();
             this.flyoutController = null;
         }
-        this.updateState({ flyoutOpen: false, layers: resetLayers ? null : this.getState().layers });
+
+        if (this.popupController) {
+            this.popupController.close();
+            this.popupController = null;
+        }
```
ZakarFin, line 214:
```suggestion
       this.closePopup();
```
Or the renamed version.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -170,18 +193,148 @@ class FeatureDataPluginUIHandler extends StateHandler {
...
+                filteredList = filterFeaturesByAttribute(filteredList, filter);
+            });
+            filteredList
+                .map((feature) => feature.id)
+                .filter(id => !!id)
+                .forEach(id => filteredIds.add(id));
+        });
+
+        this.selectionService.setSelectedFeatureIds(activeLayerId, [...filteredIds]);
+        this.updateState({ selection: [...filteredIds] });
```
ZakarFin, line 286: We should remove updateState() from here and do it when we receive the event about selected features. That way we don't need to care how the features got selected and selecting from map works as well as selecting through this tool.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
DenverCoder544: This was on my todo list as an item to be forgotten. Fixed.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
DenverCoder544: Removed excessive state update.

bundles/framework/featuredata/view/SelectByProperties.jsx
```diff
@@ -0,0 +1,266 @@
...
+                <StyledSelectSmall
+                    options={generateLogicalOperatorOptions()}
+                    value={filter.logicalOperator}
+                    onChange={(value) => { filter.logicalOperator = value; updateFilters(index, filter); }}
+                />
+            }
+            { showAddRemove &&
+                <>
+                    <Button onClick={(() => addFilter())}><PlusCircleTwoTone/></Button>
+                    { showRemove && <Button onClick={(() => removeFilter(index))}><MinusCircleTwoTone/></Button> }
```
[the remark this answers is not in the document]
DenverCoder544: changed the buttons to the look of those in myplaces. IMO some kind of confirmation dialog would be more annoying than helpful, after all we're not resetting anything _that_ significant here

PR #2275, opened 2023-05-19
Warn if layer min/max scale is not defined.

Creaticide:


bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/Scale.jsx
```diff
@@ -72,6 +74,17 @@ const Scale = ({ layer, scales = [], controller, getMessage }) => {
         // if max zoom is undefined the slider needs to be at the max value
         layerMaxZoom = maxZoomUnrestrictedValue;
     }
+
+    useEffect(() => {
+        if (layer && !layerChecked) {
+            setLayerChecked(true);
+            const {minscale, maxscale} = layer;
+            if (!minscale || !maxscale || minscale === -1 || maxscale === -1) {
```
ZakarFin, line 82: We don't need to notify for value `-1` or missing value. `-1` is used for "no value" in either the service doesn't give any restrictions on capabilities or it couldn't be parsed. I guess the best check we could do for notifying when min == max and the value is not -1. But do we need to do it in useEffect?

ZakarFin: Or do we know how the slider behaves when minscale and maxscale is 0? Can we just notify when the slider would show "everything is ok" when in fact no zoom-levels match the scale limits?

PR #2287, opened 2023-05-25
Featuredata react conversion toolbar button

DenverCoder544:


packages/framework/bundle/featuredata/bundle.js
```diff
@@ -41,6 +41,12 @@ Oskari.clazz.define("Oskari.mapframework.bundle.featuredata.FeatureDataBundle",
         }, {
             "type": "text/javascript",
             "src": "../../../../bundles/framework/featuredata/plugin/FeatureDataPlugin.js"
+        }, {
+            "type": "text/javascript",
+            "src": "../../../../bundles/framework/featuredata/PopupHandler.js"
+        }, {
+            "type": "text/javascript",
+            "src": "../../../../bundles/framework/featuredata/plugin/MapSelectionPlugin.js"
```
ZakarFin, line 49: We should just import these on the impl files instead.

bundles/framework/featuredata/PopupHandler.js
```diff
@@ -0,0 +1,120 @@
+import { FEATUREDATA_BUNDLE_ID } from './view/FeatureDataContainer';
+import { showSelectToolPopup } from './view/SelectToolPopup';
+/**
+ * @class Oskari.mapframework.bundle.featuredata.PopupHandler
+ *
+ * Handles map selection popup functionality.
+ */
+Oskari.clazz.define('Oskari.mapframework.bundle.featuredata.PopupHandler',
```
ZakarFin, line 8: This doesn't need to be defined as Oskari.clazz. We should just create something that exports the functions that we can import where needed.

bundles/framework/featuredata/plugin/MapSelectionPlugin.js
```diff
@@ -0,0 +1,146 @@
...
+                circle: function () {
+                    sb.postRequestByName('DrawTools.StartDrawingRequest', [
+                        me.DRAW_REQUEST_ID,
+                        'Circle'
+                    ]);
+                }
+            };
+        }
+    }, {
+        extend: ['Oskari.mapping.mapmodule.plugin.AbstractMapModulePlugin'],
```
ZakarFin, line 137: This doesn't really reference the map in any way so we could implement this as a regular file/part of the "PopupHandler" instead. This used to call the map directly to activate drawing tools etc, but now it just uses the requests without needing direct reference to the mapmodule.

PR #2310, opened 2023-06-05
Classification popup styles.

Creaticide:


ZakarFin:
Makes classification popup have a scrollbar if it would be larger than the screen.

PR #2333, opened 2023-06-15
AdminLayerEditor scales fix

okauppinen:
LayerHelper removes -1 values and scale checks uses -1 => Default to -1
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/admin/admin-layereditor/view/LayerHelper.js#L44-L49

### DIFF
--- a/bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/Scale.jsx
+++ b/bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/Scale.jsx
@@ -78,7 +78,7 @@ const Scale = ({ layer, scales = [], controller, getMessage }) => {
     useEffect(() => {
         if (layer && !layerChecked) {
             setLayerChecked(true);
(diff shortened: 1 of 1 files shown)


bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/Scale.jsx
```diff
@@ -78,7 +78,7 @@ const Scale = ({ layer, scales = [], controller, getMessage }) => {
     useEffect(() => {
         if (layer && !layerChecked) {
             setLayerChecked(true);
-            const {minscale, maxscale} = layer;
+            const {minscale = 1, maxscale = -1} = layer;
```
ZakarFin, line 81: is minScale 1 and not -1 on purpose here?

okauppinen: no. made some other changes and reverted them with this mistake. fixed.

PR #2337, opened 2023-06-19
Replace enzyme with react testing lib

DenverCoder544:


bundles/framework/layerlist/view/LayerViewTabs/LayerList/Filter/Filter.test.js
```diff
@@ -1,49 +1,32 @@
...
+import { render } from '@testing-library/react';
+import { beforeEach } from '@jest/globals';
+import '@testing-library/jest-dom';
 import { Filter } from './Filter';
 import { testFilters } from './Filter.test.util';
 import { Controller, LocaleProvider } from 'oskari-ui/util';
-import { Option } from 'oskari-ui';
 import { getBundleInstance } from '../../test.util';
-
 // from https://github.com/ant-design/ant-design/blob/4.3.1/components/select/__tests__/index.test.js
```
ZakarFin, line 9: This comment probably refers to the removed toggleOpen() function so it should be removed as well.

PR #2341, opened 2023-06-22
Layer permissions confirms, filters & select all

Creaticide:
"Select all" checkboxes, confirm unsaved changes, filter by name.

### DIFF
--- a/bundles/admin/admin-permissions/handler/layerRightsHandler.js
+++ b/bundles/admin/admin-permissions/handler/layerRightsHandler.js
@@ -7,9 +7,16 @@ class UIHandler extends StateHandler {
         this.setState({
             roles: [],
             permissions: [],
+            resources: [],
             selectedRole: 0,
             loading: false,
-            changedIds: new Set()
+            changedIds: new Set(),
+            checkAllCheckboxes: {},
+            pagination: {
+                pageSize: 50,
+                page: 1,
+                filter: ''
+            }
         });
         this.addStateListener(consumer);
         this.fetchRoles();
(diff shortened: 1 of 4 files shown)


bundles/admin/admin-permissions/view/LayerRightsTable.jsx
```diff
@@ -46,66 +39,26 @@ const getPermissionTableHeader = (permission) => {
...
         state.permissions.names.forEach((name, index) => {
             columnSettings.push({
                 align: 'left',
-                title: getPermissionTableHeader(name),
+                title: () => (
+                    <HeaderCell>
+                        {getPermissionTableHeader(name)}
+                        <CheckAllCheckbox
+                            checked={state.checkAllCheckboxes[state.pagination.page]?.[name.id]}
+                            onChange={() => controller.setCheckAllForPermission(name.id)}
```
ZakarFin, line 58: What if we didn't store the state of this "check all" but calculated it on the fly using helpers like:
```
const hasPermission = (layer, permissionType) => {
    return layer.permissions.includes(p => p === permissionType)
};
const allChecked = (permissionType) => {
    state.resources.every(layer => hasPermission(layer, permissionType))
};
```
So this could be something like (sending boolean param to setter to see if we should remove the permission or add it):
```
const permissionType = name.id;
const allCurrentLayersHavePermission = allChecked(permissionType);

                        <CheckAllCheckbox
                            checked={allCurrentLayersHavePermission}
                            onChange={() => controller.setPermissionForCurrentLayers(permissionType, !allCurrentLayersHavePermission)}
```

Creaticide: Done with small changes. Array.every didn't work here for some reason.

bundles/admin/admin-permissions/handler/layerRightsHandler.js
```diff
@@ -39,6 +45,57 @@ class UIHandler extends StateHandler {
...
+                page: page
+            }
+        });
+    }
+
+    search (searchText) {
+        const permissions = structuredClone(this.state.permissions?.resource?.filter(r => r.name.toLowerCase().includes(searchText.toLowerCase())));
+        this.updateState({
+            resources: permissions,
+            changedIds: new Set(),
```
ZakarFin, line 78: Why does searching and clearing search change the changedIds? Aren't those the layers that have permissions changed after saving or what is tracked with it?

Creaticide: That tracks which layers have been edited and need to be saved, and search clears unsaved changes.

PR #2345, opened 2023-07-13
Metadata search react

DenverCoder544:
Made some changes compared to the old implementation, which seemed to make sense:
 
-removed info-icon link in the search result. The metadataflyout now opens simply by clicking the name.
-removed the delete - button, since it seemed kind of useless anyway.
-wasn't sure whether the "additional" action links - functionality is still valid or in use somewhere so that's not implemented (yet). If it's still valid/neede we might want to implement that in another pr.  

image

image


bundles/catalogue/metadatasearch/MetadataStateHandler.js
```diff
@@ -0,0 +1,260 @@
...
+                    formdata[key] = advancedSearchValues[key];
+                }
+            }
+        });
+
+        this.searchService.doSearch(formdata, (results) => this.updateSearchResults(results));
+    }
+
+    updateSearchResults (json) {
+        json?.results?.forEach(result => this.updateLayerInfo(result));
```
ZakarFin, line 81: Maybe map() instead to return a modified result so it's easier to follow the code.

bundles/catalogue/metadatasearch/instance.js
```diff
@@ -0,0 +1,403 @@
...
+
+    /**
+     * @method create called automatically on construction
+     * @static
+     */
+    function () {
+        const sandboxName = (this.conf ? this.conf.sandbox : null) || 'sandbox';
+        this.sandbox = Oskari.getSandbox(sandboxName);
+        this.sandbox.register(this);
+        this.started = false;
```
ZakarFin, line 35: This is usually done on start() instead and I'm not sure if started variable is actually needed, but I recognize it from other bundles so it's probably copy-pasted here.

bundles/catalogue/metadatasearch/view/advanced-search/AdvancedSearchDropdown.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { AdvancedSearchInputLabel, AdvancedSearchRowContainer, AdvancedSearchSelect } from './AdvancedSearchStyledComponents';
+import { Option } from 'oskari-ui';
+import PropTypes from 'prop-types';
+export const AdvancedSearchDropdown = (props) => {
+    const { title, options, onChange, selected } = props;
+    const hasOptions = options && options?.values?.length && options.values.length > 0;
+    return <AdvancedSearchRowContainer>
```
ZakarFin, line 8: I think we can just return null and not render the label nor dropdown if we don't have any values to show for the user.

bundles/catalogue/metadatasearch/view/advanced-search/AdvancedSearchMulti.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { AdvancedSearchCheckboxGroupContainer, AdvancedSearchInputLabel, AdvancedSearchRowContainer } from './AdvancedSearchStyledComponents';
+import { METADATA_BUNDLE_LOCALIZATION_ID } from '../../instance';
+import { PropTypes } from 'prop-types';
+import { Checkbox } from 'oskari-ui';
+
+export const AdvancedSearchMulti = (props) => {
+    const { title, options, onChange, selected } = props;
+    const hasOptions = options && options?.values?.length && options.values.length > 0;
```
ZakarFin, line 9: Same here, we don't need to render if it there wouldn't be anything the choose.

bundles/catalogue/metadatasearch/view/MetadataSearchContainer.jsx
```diff
@@ -0,0 +1,92 @@
...
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import { METADATA_BUNDLE_LOCALIZATION_ID } from '../instance';
+import { SearchInput, Spin } from 'oskari-ui';
+import { AdvancedSearchContainer } from './advanced-search/AdvancedSearchContainer';
+import { MetadataSearchResultListContainer } from './resultlist/MetadataSearchResultListContainer';
+import { FlexRowCentered } from './advanced-search/AdvancedSearchStyledComponents';
+
+const Description = () => {
+    return <div>{Oskari.getMsg(METADATA_BUNDLE_LOCALIZATION_ID, 'metadataSearchDescription')}</div>;
```
ZakarFin, line 11: We could use `<Message />` here, but since this is the only instance to use it AND we need to use Oskari.getMsg() anyway for the placeholder, I think it's ok to not use the Message-tag here.

bundles/catalogue/metadatasearch/view/MetadataSearchContainer.jsx
```diff
@@ -0,0 +1,92 @@
...
+export const renderMetadataSearchContainer = (state, controller, element) => {
+    const render = (state, controller) => {
+        ReactDOM.render(<MetadataSearchContainer state={state} controller={controller} />, element);
+    };
+
+    render(state, controller);
+    return {
+        update: render
+    };
+};
```
ZakarFin, line 92: This is unusual, but I guess it's needed because the container where it's rendered can be either on a jQuery based tab or jQuery based flyout?

packages/catalogue/metadatasearch/bundle.js
```diff
@@ -0,0 +1,81 @@
...
+
+		"scripts" : [
+			/* service */
+			{
+	            "type": "text/javascript",
+	            "src": "../../../bundles/catalogue/metadatasearch/service/MetadataOptionService.js"
+	        }, {
+	            "type": "text/javascript",
+	            "src": "../../../bundles/catalogue/metadatasearch/service/MetadataSearchService.js"
+	        },
```
ZakarFin, line 33: These could be just imported on instance.js

ZakarFin: Or rather they already are imported in the handler but both are ES classes so referencing them here probably does nothing but add dead code :)

ZakarFin:
Looks pretty good. Added some comments but lets wait with merging until 2.13

bundles/catalogue/metadatasearch/MetadataStateHandler.js
DenverCoder544: yeah, definitely easier to read that way

bundles/catalogue/metadatasearch/instance.js
DenverCoder544: Yeah, it's copypasted alright. And it is set to true at start method unless it's already set to true, in which case the start() - method does nothing. Wasn't sure of the side-effects this might have so I left it as it was.

bundles/catalogue/metadatasearch/view/advanced-search/AdvancedSearchDropdown.jsx
DenverCoder544: true

bundles/catalogue/metadatasearch/view/MetadataSearchContainer.jsx
DenverCoder544: Changed this. I agree it's cleaner that way even if we still gotta use the getMsg later on.

bundles/catalogue/metadatasearch/view/MetadataSearchContainer.jsx
DenverCoder544: I thought so but I'm no quite sure if that's true... The whole bit seems a bit shady now, gotta dig a bit deeper into this.

DenverCoder544: Turns out it really didn't have to be that complex. Refactored this bit and it still should work in botha cases - when the search bundle is available and when this is standalone.

bundles/catalogue/metadatasearch/instance.js
```diff
@@ -0,0 +1,411 @@
...
+        * Add search result action.
+        * @public
+        * @param {jQuery} actionElement jQuery action element
+        * @param {Function} callback the callback function
+        * @param {String} bindCallbackTo the jQuery selector where to bind click operation
+        * @param {String} actionTextElement action text jQuery selector. If it's null then text showed on main element
+        * @param {String} actionText action text
+        * @param {Function} showAction function. If return true then shows action text. Optional.
+        */
+        addSearchResultAction: function (actionElement, actionTextElement, callback, bindCallbackTo, actionText, showAction) {
```
ZakarFin, line 384: I think we can just remove everything regarding search result action in this implementation. If someone is using them we can add them here later. If anyone is using them they will probably need some changes on the usage (jQuery -> React)  as well even if we include it directly.

ZakarFin: -> Lets remove these.

DenverCoder544: done & done.

PR #2346, opened 2023-07-14
Add "no UI" to announcements plugin

osqq:
Add "noUI" checkbox option to Announcements tool/plugin so we don't have to always use the provided UI on apps using published maps. Also added a function that provides the app the selected Announcements so we can access just the certain Announcements that were selected with the publisher tool.

function: _getSelectedAnnouncements()_

### DIFF
--- a/bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
+++ b/bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
@@ -35,12 +35,16 @@ Oskari.clazz.define('Oskari.admin.bundle.admin-announcements.publisher.Announcem
                 '</div>'),
             inputCheckbox: jQuery('<div><input type="checkbox" name="announcement"/><label></label></div>')
         };
+        this.noUI = null;
+        this.noUiIsCheckedInModifyMode = false;
     }, {
 
         init: function (data) {
             const me = this;
             me.data = data;
             me.selectedAnnouncements = [];
+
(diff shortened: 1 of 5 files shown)


bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -35,13 +35,17 @@ Oskari.clazz.define('Oskari.admin.bundle.admin-announcements.publisher.Announcem
                 '</div>'),
             inputCheckbox: jQuery('<div><input type="checkbox" name="announcement"/><label></label></div>')
         };
+        this.noUI = true;
```
ZakarFin, line 38: Default to showing the UI instead?
```suggestion
        this.noUI = false;
```

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -125,6 +129,34 @@ Oskari.clazz.define('Oskari.admin.bundle.admin-announcements.publisher.Announcem
...
+                'Oskari.userinterface.component.CheckboxInput'
+            );
+
+            input.setTitle(labelNoUI);
+            input.setHandler(function (checked) {
+                if (checked === 'on') {
+                    me.noUI = true;
+                    me.getPlugin().teardownUI();
+                } else {
+                    me.noUI = null;
```
ZakarFin, line 144:
```suggestion
                    me.noUI = false;
```

PR #2348, opened 2023-07-20
add featuredata tool to vector layers

DenverCoder544:
Enable opening (the new react implementation of) featuredata from the maplayer list's three dot menu.

image

### DIFF
--- a/bundles/framework/featuredata/instance.js
+++ b/bundles/framework/featuredata/instance.js
@@ -30,7 +30,7 @@ Oskari.clazz.define('Oskari.mapframework.bundle.featuredata.FeatureDataBundleIns
          * Event is handled forwarded to correct #eventHandlers if found or discarded if not.
          */
         onEvent: function (event) {
-            var handler = this.eventHandlers[event.getName()];
+            const handler = this.eventHandlers[event.getName()];
             if (!handler) {
                 return;
             }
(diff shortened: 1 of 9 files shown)


bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -63,6 +63,12 @@ class FeatureDataPluginUIHandler extends StateHandler {
     }
 
     setActiveTab (layerId) {
+        // when clicking on a tab directly, layerId will be a number.
+        // When clicked in the "show more" menu of the tablist, it will be a String and no features are found.
+        // so we gotta make a conversion for a numeric layerid just in case. :(
+        if (Oskari.util.isNumber(layerId)) {
+            layerId = +layerId;
```
ZakarFin, line 70: Good commenting above 👍 Maybe use parseInt() instead to make it easier to understand?

bundles/framework/featuredata/instance.js
```diff
@@ -81,6 +86,45 @@ Oskari.clazz.define('Oskari.mapframework.bundle.featuredata.FeatureDataBundleIns
...
+            }
+            // add feature data tool for layer
+            const tool = Oskari.clazz.create('Oskari.mapframework.domain.Tool');
+            const label = Oskari.getMsg(FEATUREDATA_BUNDLE_ID, 'layer.featureData');
+
+            tool.setName('featureData');
+            tool.setIconCls('show-featuredata-tool');
+            tool.setTitle(label);
+            tool.setTooltip(label);
+            tool.setCallback(() => this.sandbox.postRequestByName('Featuredata.ShowFeatureDataRequest', [layer.getId()]));
```
ZakarFin, line 124: We don't really need to use a request here since we can call featuredata bundles own function in this callback, but the request might come in handy in the future for opening the table programmatically through RPC.

PR #2359, opened 2023-08-03
Allow decimal coordinate input for ETRS89 & KKJ projections.

Creaticide:


bundles/framework/coordinatetool/plugin/CoordinatePluginHandler.js
```diff
@@ -103,6 +103,14 @@ class UIHandler extends StateHandler {
                 lat: this.formatNumber(this.state.latField, '.')
             }
         };
+
+        // Special treatment for these degree format projections to allow coordinates in decimal format as well
+        if (this.state.selectedProjection === 'EPSG:4258' || this.state.selectedProjection === 'LATLON:kkj') {
```
ZakarFin, line 108: Instead of hardcoding the projections, lets use the config to detect if selectedProjection _should be_ degrees with something like this:
```
conf.projectionShowFormat[this.state.selectedProjection]?.format === 'degrees'
```
Note that there might not be projectionShowFormat object at all so we should check that as well AND use the default from `projectionShowFormat.format` as fallback.

I think the regexp is not needed at all here. I guess we need to append the '°' IF it's not there already (need to check that as well) if we use `Oskari.util.coordinateIsDegrees()` in the next function.

The config for coordinatetool in for example Paikkatietoikkuna is this:
```
{
  "reverseGeocodingIds": "TM35LEHTIJAKO_CHANNEL",
  "supportedProjections": [
    "EPSG:3067",
    "NLSFI:etrs_gk",
    "NLSFI:kkj",
    "NLSFI:ykj",
    "EPSG:4258",
    "LATLON:kkj",
    "EPSG:3046",
    "EPSG:3048",
    "EPSG:3873",
    "EPSG:3874",
    "EPSG:3875",
    "EPSG:3876",
    "EPSG:3877",
    "EPSG:3878",
    "EPSG:3879",
    "EPSG:3880",
    "EPSG:3881",
    "EPSG:3882",
    "EPSG:3883",
    "EPSG:3884",
    "EPSG:3885"
  ],
  "showEmergencyCallMessage": true,
  "projectionShowFormat": {
    "decimals": 3,
    "format": "metric",
    "EPSG:4258": {
      "decimals": 3,
      "format": "degrees"
    },
    "LATLON:kkj": {
      "decimals": 3,
      "format": "degrees"
    },
    "EPSG:3067": {
      "decimals": 0
    }
  }
}

PR #2362, opened 2023-08-08
Iconbtn colors

okauppinen:
Hover color didn't work if icon has inline style color
-  Added color prop for IconButton
- Remove antd color import and copied values to constant. Use predifened colors for delete, info, accept, reject
- Add active (sticky) prop for IconButton like mapplugin buttons have
- Remove DeleteIcon.jsx and use IconButton for DeleteButton icon types
- Move disablewrapper to themebutton (also button without tooltip gets wrappper)
- Refactor admin announcements and layeranalytics to use IconButton

`<IconButton type='delete' bordered=true/false>` is almost same than `<DeleteButton type='icon || button'>`
For IconButton confirm text is always same.


src/react/components/buttons/IconButton.jsx
```diff
@@ -87,31 +103,45 @@ const getConfirmProps = (type) => {
...
+        className = `${className} t_${type}`;
     }
-    if (!color) {
-        color = '#ffd400';
-    } else if (Oskari.util.isDarkColor(color)) {
-        color = getColorEffect(color, EFFECT.LIGHTEN);
+    let hover = theme?.color?.accent;
+    if (hover && Oskari.util.isDarkColor(hover)) {
+        hover = theme?.color?.primary;
     }
```
ZakarFin, line 124: I don't quite understand this logic. If hover (accent color) is dark -> hover is changed to primary color? Is this correct? Then it's checked again for being dark in the next else if and a lighten effect is applied?

okauppinen: Changed for this pr only variable name (color -> hover). I think that the goal is to get light hover color based on theme. Not sure if hover should be lighten accent if both colors are dark (now it's lighten primary).
`
let hover = theme?.color?.accent || theme?.color?.primary || '#ffd400';
if (Oskari.util.isDarkColor(color)) {
    color = getColorEffect(color, EFFECT.LIGHTEN);
}
`
Not sure if only accent color is used in theme. If it's both or none, could be:
`
let hover = theme?.color?.accent;
if (hover && Oskari.util.isDarkColor(hover)) {
    hover = getColorEffect(hover, EFFECT.LIGHTEN);
} else if (!hover ) {
    hover = '#ffd400'
}
`
First contains useless isDark check for #ffd400 if theme isn't defined

okauppinen: Checked how MapButton uses hover and found this:
https://github.com/oskariorg/oskari-frontend/blob/master/src/react/theme/ThemeHelper.js#L36

PR #2363, opened 2023-08-08
Markers refactor

okauppinen:
Purpose was to refactor marker handling but expanded to customization handle also colors and default vector style. The reason was that generate blank style uses fill color from theme which isn't available on predefined colors and using own color set is easier without hard coded list. Also default vector style is stored in db which isn't always same that hard coded style in frontend.

Removed svg wrappers from StyleEditor preview and ol generator.
- StyleEditor: used css to center img tag and marker svg as src. Size is set to img tag (32 x scale) => same size in map (Point preview is little smaller than before)
- ol generator: use offsets to adjust positioning and scale to adjust size

Note! not sure what kind of external svg icons should be supported. Only {data: 'svg', offsetX, offsetY} and should give svg contain stroke and fill already. Now left as is. Also there mapmodiule has wellknownStyle which could be removed.

Note! sizePx isn't calculated for oskari markers 
not sure how it should work because previously had svg wrapper with 64px viewBox and only width and height varied by size or sizePx:
Size: 1 => 50px
image
Size: 5 => 90px
image
So should it scale 32 size marker to given sizePx or use same logic than size

Not sure why size 1-5 is 50-90 width/height scaled to 64 viewBox


ZakarFin:
Looks good for merging after the conflicting change has been resolved.

PR #2366, opened 2023-08-08
Rpc specs

AleksiKaa:
Updated and implemented unit tests for the RPC library using jasmine js


tests/rpc/spec/location_spec.js
```diff
@@ -45,18 +45,22 @@ describe('Location', function(){
...
-                expect(result).toContain('id');
-                expect(result).toContain('lat');
-                expect(result).toContain('lon');
-                expect(result).toContain('rank');
-                expect(result).toContain('region');
-                expect(result).toContain('type');
-                expect(result).toContain('village');
-                //expect(result).toContain('zoomScale');
+            expect(result).toContain('zoomScale');
+            expect(result).toContain('paikkatyyppi');
```
ZakarFin, line 53: The other keys should be ok, but `paikkatyyppi` is specific to search backend used in Paikkatietoikkuna. We could do something like add "additionalSearchResultKeys" array of values to parameter file, add `paikkatyyppi` to `paikkatietoikkuna_parameters.js` and use that here to check this.

tests/rpc/spec/map_spec.js
```diff
@@ -1,32 +1,35 @@
...
-        it('Get all layers', function(done) {
-            channel.getAllLayers(function(data) {
+        it('Get all layers', function (done) {
+            channel.getAllLayers(function (data) {
                 // Expect atleast basemap layer to be found.
                 expect(Object.keys(data).length).toBeGreaterThan(0);
-                // Expect basemap layer to be have 4 elements.
-                expect(Object.keys(data[0]).length).toBe(4);
+                // Expect basemap layer to be have at least 4 elements.
+                expect(Object.keys(data[0]).length).toBeGreaterThanOrEqual(4);
```
ZakarFin, line 32: The layer count is something that should come from the params-file.

tests/rpc/spec/location_spec.js
AleksiKaa: Added an array called additionalSearchResultKeys to paikkatietoikkuna_parameters.js. Unit test in question now loops over the entries in said array and expects the values of the array to appear in SearchResultEvent.

PR #2368, opened 2023-08-08
Admin layereditor WFS attributes part 3 - feature filter 

okauppinen:
UI for creating feature filter. Added FeatureFilter.jsx to react compnents as it could be used for user own optional styles. FeatureFilter has helper function to clean filter values to save only valid filters. Uses same modal than properties. Added also select for idProperty (attributes.data.idProperty).

- Only number type has < and > operators and `><` range toggle button (range has 2 operators)
- String (not number) has `Aa` toggle for case (in)sensitive
- `AND` and `OR` operators needs at least 2 property filters (rows) ` {key:${name}, ${operator}: ${value}}` so delete icons is rendered 3 rows
- `in` and `notIn` operators uses array. Tried textarea and row separator but it didn't look nice. For now textinput with placeholder and `;` separator is used.
- active prop is used for toggle button (uses same color than hover if active, like XY-tool)
- filter contains property or AND and/or OR, property is stored only if valid AND < 2 and OR < 2
- for both AND is used: (...AND) AND (...OR...)

https://oskari.org/documentation/examples/oskari-filter
https://github.com/oskariorg/oskari-server/blob/master/service-wfs-client/src/main/java/org/oskari/service/wfs/client/OskariWFSFilter.java#L146-L211

Contains (but not all): https://github.com/oskariorg/oskari-frontend/pull/2367
Changed delete icons to use onClick (not onConfirm) as admin can click cancel (modal or layereditor).

image

image

image


ZakarFin:
https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/framework/featuredata/view/SelectByProperties.jsx looks _very similar_ to FeatureFilter. But merging for now and lets figure out if we can combine these two later.

PR #2374, opened 2023-08-11
WFS not unique feature id workaround

okauppinen:
Use property value as feature id if overriding id is defined.

TODO: rename

### DIFF
--- a/bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/VectorLayerHandler.ol.js
+++ b/bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/VectorLayerHandler.ol.js
@@ -180,7 +180,18 @@ export class VectorLayerHandler extends AbstractLayerHandler {
                         }
                     });
                     const features = source.getFormat().readFeatures(resp);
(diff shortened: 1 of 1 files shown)


bundles/mapping/mapwfs2/plugin/WfsVectorLayerPlugin/impl/VectorLayerHandler.ol.js
```diff
@@ -180,7 +180,18 @@ export class VectorLayerHandler extends AbstractLayerHandler {
...
+                        // workaround for layers which returns generated id for features.
+                        // "The feature id is considered stable and may be used when requesting features or comparing identifiers returned from a remote source."
+                        const idProperty = layer.getIdProperty();
+                        if (idProperty) {
+                            const id = ftr.getProperties()[idProperty];
+                            if (id) {
+                                ftr.setId(id);
+                            }
+                        }
+                        ftr.set(WFS_ID_KEY, ftr.getId())
```
ZakarFin, line 193:
```suggestion
                        ftr.set(WFS_ID_KEY, ftr.getId());
```

PR #2379, opened 2023-08-17
WIP: Fix gfi popup when map is rotated.

Creaticide:


ZakarFin:
Using OpenLayers to generate the parameters so it handles map rotation etc.

PR #2380, opened 2023-08-17
Enable opts config for fixing render order for decluttered labels

ZakarFin:
Related to:
- https://github.com/openlayers/openlayers/issues/10096
- https://github.com/openlayers/openlayers/issues/13235


ZakarFin:
Using this requires updating oskari_maplayer table options field with:
```
{
"className ": "somerandomstring"
}
```
For the layer with label issues.

PR #2389, opened 2023-08-29
Fix popup positioning on mobile

Creaticide:
Always open popups top or bottom on mobile to prevent popup overflowing the screen.


src/react/components/window/Popup.jsx
```diff
@@ -86,8 +87,17 @@ export const Popup = ThemeConsumer(( {title = '', children, onClose, bringToTop,
             return handleUnmounting;
         }
         // center after content has been rendered
+        let placement = options.placement;
+        // Open on bottom/top on mobile to prevent popup overflowing the screen
+        if (Oskari.util.isMobile()) {
+            if (placement && placement.indexOf('bottom') > -1) {
+                placement = PLACEMENTS.BOTTOM;
+            } else {
```
ZakarFin, line 95: This will move all popups that would open at center to top (if they are not configured to open on bottom). But lets see how it works.

PR #2393, opened 2023-08-31
Use theme for info/metadata icon colors.

Creaticide:


src/react/components/icons/Info.jsx
```diff
@@ -24,12 +44,14 @@ const StyledInfoIcon = styled(QuestionCircleOutlined)`
 export const Info = ({ children, title, size = 16, style }) => {
 
     return (
-        <Tooltip title={title || children}>
-            <StyledInfoIcon
-                className='t_icon t_info'
-                style={{ fontSize: `${size}px`, ...style }}
+        <ThemeProvider>
```
ZakarFin, line 47: ThemeProvider should not be in the icon but somewhere where this is used. The theme might be different if the usage is on a map plugin or in a geoportal setting.

src/util.js
```diff
@@ -435,6 +435,13 @@ Oskari.util = (function () {
         return this.getColorBrightness(color) === 'light';
     };
 
+    util.hexToRgb = (hex) => {
```
ZakarFin, line 438: We should add some tests for this as util.js already has a test file.

Creaticide: Didn't notice that util.js already had a hexToRgb function, so this is not needed.

PR #2394, opened 2023-09-01
Admin layereditor vector properties tuning

okauppinen:
- remove add/edit prefix from button labels and use badge
- use badge for filter language options
- group properties and move unique id select to bottom
- fix add/edit detection

For now Badges show how many definitions are made:
- lang based filters count
-  languages for labels count
- formatters for properties count
- 1 for feature filter
- how many properties are selected for language (filter lang select options)

Just noticed that we had already ThemedBadge:
https://github.com/oskariorg/oskari-frontend/blob/master/src/react/components/ThemedBadge.jsx

The difference is that color, showZero and overflowCount can be used. Not sure if is it better to have separate component or use prop as now in Badge -> use ThemedBadge and revert changes for Badge or remove ThemedBadge and change to use <Badge themed/>

image

image


ZakarFin:
I added ThemedBadge some time ago because I thought that by default the badge should use theming, but didn't have time to test if things broke or looked bad if theming was used in all the places badge is used. I still think that useTheme should default to true if we make this change. Then we could remove ThemedBadge.

If we need a non-themed Badge I think we still should have only one Badge and then use `themed=false` on those to make them easier to find.

src/react/components/Badge.jsx
```diff
@@ -2,22 +2,48 @@
...
+}) => {
+    const useTheme = themed && theme;
+    let color = propsColor;
+    if (useTheme) {
+        color = getHeaderTheme(theme).getAccentColor(theme);
+    } else if (inversed) {
+        color = '#333';
+    } else if (!propsColor) {
+        color = '#999';
+    }
```
ZakarFin, line 33: Should we just list the places that badge is used, default themed to true and check if we need non-themed ones and if we can use some colors from theme of constants for inversed color etc.

okauppinen: Removed inversed and themed prop. Added constant for zero. For now badge is rendered disabled like grey color if count is 0 or missing. Theme is used if color isn't given. Removed ThemedBadge.

Inversed true/false was used in layerlist. Changed it to use theme. Changes can be reverted (black badge):
`<Badge count={badgeText} />;`
=>
`<Badge color='#333' count={badgeText} />;`
in layerlist

PR #2414, opened 2023-09-06
display create and last login in mydata

DenverCoder544:
image


ZakarFin:
Looks good for merging, but lets do it in 2.13

PR #2419, opened 2023-09-08
Show single popup on marker tool on mobile.

Creaticide:
On mobile, close first popup when map is clicked and re-open tool after marker is added.


ZakarFin:
Looks good for merging, but lets do it in 2.13.

PR #2420, opened 2023-09-08
Hide nav on measure/marker tool open

Creaticide:


ZakarFin:
Looks good for merging, but lets do it in 2.13

PR #2422, opened 2023-09-08
Detect mouse, disable mouse coordinates.

Creaticide:
- Util function to detect if user has a mouse.
- Disable mouse coordinates on coordinate popup if no mouse detected.


ZakarFin:
**Before**
Mouse toggle always shown, but the coordinates are not updated when there is no mouse-like device to detect hovering over the map:

image

**After**
No mouse -> no toggle:
image

PR #2429, opened 2023-09-12
Layerlist show coverage

DenverCoder544:


bundles/mapping/mapmodule/plugin/layers/coveragetool/CoverageToolButton.jsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import { ThemeConsumer } from 'oskari-ui/util';
+import { getNavigationTheme } from 'oskari-ui/theme';
+import { Button } from 'oskari-ui';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+const StyledButton = styled(Button)`
```
ZakarFin, line 7: Could we use this instead: https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapmodule/MapModuleButton.jsx?

packages/mapping/ol/mapmodule/bundle.js
```diff
@@ -39,6 +39,13 @@ Oskari.clazz.define(
                     'type': 'text/javascript',
                     'src': '../../../../bundles/mapping/mapmodule/plugin/layers/LayersPlugin.ol.js'
                 },
+                /**
+                 * toggle layer coverage button
+                 */
+                {
+                    'type': 'text/javascript',
+                    'src': '../../../../bundles/mapping/mapmodule/plugin/layers/coveragetool/CoverageToolPlugin.js'
```
ZakarFin, line 47: Working towards removing/simplifying the packages/bundle.js: import in https://github.com/oskariorg/oskari-frontend/blob/master/bundles/mapping/mapmodule/AbstractMapModule.js#L35 instead. This also makes this available on the 3d mapmodule which might be broken without this import.

bundles/mapping/mapmodule/plugin/layers/coveragetool/CoverageToolButton.jsx
DenverCoder544: This won't work well when the content of the button is something else than an icon (=text). The text will overflow the tiny button and since additional styling would be necessary anyway then why not go all the way...

ZakarFin:
Useful when the layer covers a smaller area of the whole map:
image

PR #2436, opened 2023-09-15
Markers cluster

okauppinen:
Enable clustering for markers. Changed mapmodule generator cluster style to return style instead  of function to use same cache and style than others (WFS). WFS uses wrapClusterStyleFunction.

Not sure if cluestered markers are wanted everywhere so made it configurable. Maybe we should add prop + plugin modifier for backend.

One problem might be that nearest address markers can be very close.

image


ZakarFin:
There's a request for showing the Oskari infobox and pointing it to a specific marker. We should test if it works with clustering, but other than that this looks good to me.

PR #2438, opened 2023-09-15
Admin-layereditor json fields POC

okauppinen:
Add tab for JSON fields (options, capabilities, attributes, params). Every json have text area to review content. Tools for adding key value pairs to json and deleting key from json. Value is parsed so also Object or Array can be added. So everything but whole json should be added or deleted.

image

image

image

These could be removed:
image
'Show GetCapabilities response' opens xml to new tab so it's little dummy to change tab to compare xml and json
If attributes is removed, setting extra values inside data (e.g. noDataValue) might need ui component. or other solution

Also tempJSON handling can be removed from handler. BTW: not sure why there is so many setters. I often use common(key, value) setter instead of many named(value). 

I think that named keys in WFSLayerOptions.java and WFSLayerAttributes.java could have own inputs in layereditor
`maxFeatures, namespaceURL, noDataValue, commonId`
`labelProperty` -> StyleEditor text tab?


ZakarFin:
Looks good to me.

bundles/admin/admin-layereditor/resources/locale/fi.js
```diff
@@ -113,6 +117,12 @@ Oskari.registerLocalization(
                     "globalWithStyles": "Tasolle on määritetty vain yksi yleinen oletuskarttaselite, vaikka sillä olisi rajapintapalvelusta useita tyylejä käytettävissä. Poista oletuskarttaselite ja määritä mahdolliset tyylikohtaiset karttaselitteet."
                 }
             },
+            "jsonFields": {
```
ZakarFin, line 120: Do we have these translations for swedish?

PR #2440, opened 2023-09-15
Scroll indicator on sidebar.

Creaticide:
Show indicator if navigation can be scrolled for more content.

### DIFF
--- a/src/oskari.dom.js
+++ b/src/oskari.dom.js
@@ -137,8 +137,9 @@ const getNavigationEl = () => {
 }
 
 const setMenuScrollIndicator = () => {
-    const showIndicator = () => {
+    const indicatorNotNeeded = () => {
         const navEl = Oskari.dom.getNavigationEl();
+        if (!navEl) return true;
         const scrollTop = navEl.scrollTop;
         const scrollHeight = navEl.scrollHeight;
         const offsetHeight = navEl.offsetHeight;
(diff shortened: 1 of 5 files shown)


src/oskari.dom.js
```diff
@@ -131,6 +136,43 @@ const getNavigationEl = () => {
...
+    const setIndicator = (event) => {
+        window.requestAnimationFrame(() => {
+            if (indicatorNotNeeded()) {
+                getNavigationEl().classList.remove('show-scroll-icon');
+            } else {
+                getNavigationEl().classList.add('show-scroll-icon');
+            }
+        });
+    };
+    const navElement = document.querySelector('nav#maptools');
```
ZakarFin, line 158: Lets use `Oskari.dom.getNavigationEl();` instead for consistency.

src/react/theme/globalStyles.js
```diff
@@ -33,5 +35,34 @@ export const setGlobalStyle = (theme = {}) => {
...
+        .oskari-root-el > nav {
+            .scroll-indicator {
+                display: none;
+                background-color: ${navColor};
+                border-top: ${headerTheme.getAccentColor()} solid 2px;
+                height: 45px;
+                width: ${navigationDimensions.width}px;
+                pointer-events: none;
+                position: absolute;
+                left: ${navigationDimensions.left}px;
```
ZakarFin, line 48: Isn't left always 0 since the scroll indicator is inside the nav? Now if the navigation is on the right side the indicator will be off screen?

PR #2443, opened 2023-09-18
Layer list empty tool menu fix.

Creaticide:
Hide menu if there is no tools to show or publisher is active.


ZakarFin:
Fixes issues where layerlist is opened in publisher and some tools that are available for layers on geoportal don't work well with publisher trying to show a reduced/embedded map.

PR #2451, opened 2023-09-25
VectorLayer - registerOnly fix

okauppinen:
Fixes for: https://github.com/oskariorg/oskari-frontend/pull/2432

- Update layer options in _updateVectorLayer()
- Remove async layer in removeMapLayerFromMap()
- Async layers are stored on: request + showLayer, registered layer can be added from layerlist to map without request => get ol layer (creates if doesn't exist) and add to map if async layer isn't stored on maplayer add event


ZakarFin:
Fix for thematic maps/statistical data functionality that was not shown properly on map after the async-layer changes

PR #2457, opened 2023-09-29
Announcements publisher fixes

osqq:
- Show selected announcements from config
- Always show plugin if selected so from the tools to avoid confusion with noUI (at the moment only shows if announcements are selected)
- CSS fixes to plugin and tool popup

### DIFF
--- a/bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
+++ b/bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
@@ -62,12 +62,13 @@ Oskari.clazz.define('Oskari.admin.bundle.admin-announcements.publisher.Announcem
                 const toolPluginAnnouncementsConf = me._getToolPluginAnnouncementsConf();
                 if (toolPluginAnnouncementsConf !== null) {
                     me.getPlugin().updateAnnouncements(toolPluginAnnouncementsConf.config.announcements);
-                    toolPluginAnnouncementsConf.config.announcements.forEach(announcement => {
-                        const filteredAnnouncement = me.announcements.filter(ann => ann.id === announcement);
-                        if (me.isAnnouncementValid(filteredAnnouncement[0])) {
+                    toolPluginAnnouncementsConf.config.announcements.forEach(announcementId => {
+                        const announcement = me.announcements.find(ann => ann.id === announcementId);
+                        if (announcement !== undefined && me.isAnnouncementValid(announcement)) {
                             // make sure there are no duplicates
-                            if (me.selectedAnnouncements.filter(ann => ann.id === filteredAnnouncement[0].id).length === 0) {
-                                me.selectedAnnouncements.push(filteredAnnouncement[0]);
+                            const alreadyAdded = me.selectedAnnouncements.some(ann => ann.id === announcement.id);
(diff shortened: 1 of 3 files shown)


bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -65,7 +65,10 @@ Oskari.clazz.define('Oskari.admin.bundle.admin-announcements.publisher.Announcem
                     toolPluginAnnouncementsConf.config.announcements.forEach(announcement => {
                         const filteredAnnouncement = me.announcements.filter(ann => ann.id === announcement);
                         if (me.isAnnouncementValid(filteredAnnouncement[0])) {
```
ZakarFin, line 67: Using find and renaming the config array item would make this more readable:
```suggestion
                    toolPluginAnnouncementsConf.config.announcements.forEach(announcementId => {
                        const announcement = me.announcements.find(ann => ann.id === announcementId );
                        if (me.isAnnouncementValid(announcement )) {
```

bundles/admin/admin-announcements/publisher/AnnouncementsTool.js
```diff
@@ -65,7 +65,10 @@ Oskari.clazz.define('Oskari.admin.bundle.admin-announcements.publisher.Announcem
...
-                            me.selectedAnnouncements.push(filteredAnnouncement[0]);
+                            // make sure there are no duplicates
+                            if (me.selectedAnnouncements.filter(ann => ann.id === filteredAnnouncement[0].id).length === 0) {
+                                me.selectedAnnouncements.push(filteredAnnouncement[0]);
+                            }
                         }
```
ZakarFin, line 72: Related to the other suggestion about renaming: 
```suggestion
                            // make sure there are no duplicates
                            if (me.selectedAnnouncements.find(ann => ann.id === announcement.id).length === 0) {
                                me.selectedAnnouncements.push(announcement);
                            }
                        }
```

osqq: can't switch the filter to find, without adding undefined check, which would make it less readable in my opinion. The renaming is very valid, but I feel the possibility of undefined is more dangerous than filter returning an empty array.

ZakarFin: The ! at the beginning of my suggestion should be enough since if find doesn't return the announcement it will add it to the selectedAnnouncements. But it's a bit hard to spot at the start and maybe it's best to use array.some() instead to make it even clearer:
```suggestion
                            // make sure there are no duplicates
                            const alreadyAdded = me.selectedAnnouncements.some(ann => ann.id === announcement.id);
                            if (!alreadyAdded) {
                                me.selectedAnnouncements.push(filteredAnnouncement[0]);
                            }
                        }
```

ZakarFin: But it's not bad. The main thing bugging me was the naming/array usage for `filteredAnnouncement[0]` :)

osqq:
lint and my local formatter fighting but should be ok now

ZakarFin:
Thanks :)

PR #2458, opened 2023-10-03
New statsgrid search flyout.

Creaticide:
Statsgrid search flyout in React
Added components:
- SearchFlyout.jsx
- IndicatorParams.jsx
- IndicatorCollapse.jsx
- SearchHandler.js

Removed:
SearchFlyout.js
IndicatorList.js
IndicatorSelection.js
IndicatorParameters.js
IndicatorParametersHandler.js


ZakarFin:
Could we do this in a way that we don't copy/paste everything to the new folder (from statsgrid2016 -> statsgrid)?

We could keep the current bundle.js as-is, add a new one under `bundles/statistics/statsgrid/bundle.js` that would only link the instance.js and languages. Then import all the files in instance.js from either the new impl folder or the old one depending if they already have a React.js based version or not? That way it will be easier to see what has been rewritten and we can consider this somewhat complete when we don't link files from the old folder anymore?

ZakarFin:
The new instance.js would have:
```
import './view/SearchFlyout';
import '../statsgrid2016/TableFlyout'; 
```
or something along those lines?

PR #2465, opened 2023-10-10
fix text overflow in mapmoduletextbutton, when content is a <Message/>

DenverCoder544:
-added a property to use text-overflow ellipsis within a Message-component
image

### DIFF
--- a/bundles/framework/featuredata/plugin/FeatureDataPlugin.js
+++ b/bundles/framework/featuredata/plugin/FeatureDataPlugin.js
@@ -92,7 +92,6 @@ Oskari.clazz.define('Oskari.mapframework.bundle.featuredata.plugin.FeatureDataPl
                 <ThemeProvider value={this.getMapModule().getMapTheme()}>
                     <MapModuleTextButton
                         visible={layers?.length > 0}
-                        text={<Message messageKey='title' bundleKey='FeatureData'/>}
                         onClick={() => this.handler.openFlyout()}
                         active={flyoutOpen}
                         loading={loadingStatus.loading}
@@ -100,7 +99,9 @@ Oskari.clazz.define('Oskari.mapframework.bundle.featuredata.plugin.FeatureDataPl
                         $marginRight={marginRight}
                         $marginLeft={marginLeft}
                         $marginTop={'10'}
-                    />
+                    >
+                        {<Message messageKey='title' bundleKey='FeatureData'/>}
+                    </MapModuleTextButton>
                 </ThemeProvider>,
                 el[0]
             );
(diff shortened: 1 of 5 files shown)


bundles/mapping/mapmodule/MapModuleTextButton.jsx
```diff
@@ -55,7 +55,8 @@ const StyledButton = styled(Button)`
     }
 `;
 
-const ButtonText = styled('span')`
+const ButtonText = styled('div')`
```
ZakarFin, line 58: Should we add a comment here that these handles cases where the text is given directly as string and the duplicated ellipsis on Message-tag is because these don't work with for example Message-tag?

bundles/framework/featuredata/plugin/FeatureDataPlugin.js
```diff
@@ -92,15 +92,16 @@ Oskari.clazz.define('Oskari.mapframework.bundle.featuredata.plugin.FeatureDataPl
...
                         onClick={() => this.handler.openFlyout()}
                         active={flyoutOpen}
                         loading={loadingStatus.loading}
                         position={this.getLocation()}
                         $marginRight={marginRight}
                         $marginLeft={marginLeft}
                         $marginTop={'10'}
-                    />
+                    >
+                        {<Message messageKey='title' bundleKey='FeatureData'/>}
```
ZakarFin, line 103: Should work without the {}?
```suggestion
                        <Message messageKey='title' bundleKey='FeatureData'/>
```

bundles/mapping/mapmodule/plugin/layers/coveragetool/CoverageToolPlugin.js
```diff
@@ -59,13 +59,16 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.CoverageToolPlu
...
                             onClick={() => {
                                 this._coverageButtonClicked();
                             }}
                             active={false}
                             position={this.getLocation()}
-                        />
+                        >
+                            {<Message bundleKey={LOCALIZATION_BUNDLE_ID}
+                                messageKey='layerCoverageTool.removeCoverageFromMap'
+                                allowTextEllipsis={true}/>}
```
ZakarFin, line 70: Same here?
```suggestion
                            <Message bundleKey={LOCALIZATION_BUNDLE_ID}
                                messageKey='layerCoverageTool.removeCoverageFromMap'
                                allowTextEllipsis={true}/>
```

bundles/mapping/mapmodule/plugin/layers/BackgroundLayerSelection.jsx
```diff
@@ -48,14 +49,15 @@ export const BackgroundLayerSelection = ({ isMobile = false, layers, current, ma
...
-                        text={layer.title}
                         active={Number.parseInt(layer.id, 10) === current?.getId()}
                         loading={false}
                         $isDropdown={false}
                         $minWidth={BUTTON_WIDTH}
                         data-layerid={layer.id}
                         {...rest}
-                    />
+                    >
+                        {layer.title}
```
ZakarFin, line 59: Do we need the ellipsis handling added here?

DenverCoder544: added

ZakarFin:
Also adds allowEllipsis for Message-tag 👍

PR #2466, opened 2023-10-11
Statsgrid table in React + search fixes.

Creaticide:
Statsgrid table react component + fixes to search flyout.


bundles/statistics/statsgrid/FlyoutManager.js
```diff
@@ -89,7 +89,13 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.FlyoutManager', function (insta
             return;
         }
         if (type === 'search') {
-            this.searchHandler.getController().toggleSearchFlyout(true);
+            this.searchHandler.getController().toggleSearchFlyout(true, () => this.trigger('hide', type));
+            this.hide('table');
```
ZakarFin, line 93: Does this mean that both flyouts can't be open at the same time? Currently they can be.

Creaticide: Yes, did this because the react flyouts will open on top of each other currently.

ZakarFin:
The flyout handling/toggling with "extraOnclose"s etc is kinda messy here, but I think we can make it better once all the flyouts are React-based.

PR #2470, opened 2023-10-18
add pagination

DenverCoder544:
image

Seems to be efficient enough for at least a couple of thousands of features


ZakarFin:
Page size should probably be a bit bigger but we can tune it later.

PR #2472, opened 2023-10-19
Statsgrid diagram flyout

Creaticide:
Diagram flyout in React.


bundles/statistics/statsgrid/FlyoutManager.js
```diff
@@ -91,10 +87,17 @@ Oskari.clazz.define('Oskari.statistics.statsgrid.FlyoutManager', function (insta
...
             this.tableHandler.getController().toggleTableFlyout(true, () => this.trigger('hide', type));
             this.hide('search');
+            this.hide('diagram');
+            this.trigger('show', type);
+        } else if (type === 'diagram') {
+            this.diagramHandler.getController().toggleDiagramFlyout(true, () => this.trigger('hide', type));
+            this.hide('search');
+            this.hide('table');
             this.trigger('show', type);
         } else {
```
ZakarFin, line 102: Now that these are all React-based, could we do somehing like:
```
 this.flyoutInfo[type].controller.toggleFlyout(true, () => this.trigger('hide', type));
 Object.keys(this.flyoutInfo)
    .filter(typeId => typeId !== type)
    .forEach(typeId =>  this.hide(typeId)) ;
```

This assumes that this.flyoutInfo is changed from array to an object with the type/id of the previous object as key.

bundles/statistics/statsgrid/handler/DiagramHandler.js
```diff
@@ -0,0 +1,203 @@
...
+        this.updateState({
+            loading: true
+        });
+        const indicator = this.service.getStateService().getActiveIndicator();
+
+        if (!indicator) return;
+        this.getIndicatorData(indicator, (data) => {
+            const isUndefined = (element) => {
+                return element.value === undefined;
+            };
```
ZakarFin, line 99: This should be moved out of the other function. Maybe just declare it at the start of the file?

bundles/statistics/statsgrid/view/Diagram/Diagram.jsx
```diff
@@ -0,0 +1,357 @@
...
+            .attr('dx', transformX)
+            .attr('y', 0)
+            .attr('dy', isNumber ? '0.425em' : '0.32em')
+            .style('font-size', '11px')
+            .attr('fill', color)
+            .text(rendered);
+    });
+};
+
+export const Diagram = ({ state, controller }) => {
```
ZakarFin, line 326: Controller isn't used? And we could break down the state at this point to just receive chartData and sortOrder as props instead of the whole state.

bundles/statistics/statsgrid/view/Diagram/DiagramFlyout.jsx
```diff
@@ -0,0 +1,92 @@
...
+                    value={state.activeIndicator}
+                />
+                <StyledSelect
+                    filterOption={false}
+                    options={sortOptions}
+                    onChange={(value) => controller.setSortOrder(value)}
+                    value={state.sortOrder}
+                />
+           </Selections>
+           <Diagram state={state} controller={controller} />
```
ZakarFin, line 63: Controller isn't used and Diagram is pure visualization so we should remove the prop. Also we should just pass the data and sortOrder as props to make it easier to read.

PR #2474, opened 2023-10-20
Handle statsgrid diagram without data.

Creaticide:


bundles/statistics/statsgrid/view/Diagram/DiagramFlyout.jsx
```diff
@@ -52,15 +55,23 @@ const DiagramFlyout = ({ state, controller }) => {
...
                 />
            </Selections>
-           <Diagram chartData={state.chartData} sortOrder={state.sortOrder} />
+           {!state.chartData?.data ? (
+                <NoDataMessage>
+                    <Message messageKey='datacharts.nodata' />
+                </NoDataMessage>
+           ) : (
+                <Diagram chartData={state.chartData} sortOrder={state.sortOrder} />
+           )}
```
ZakarFin, line 74: Should we have this on Diagram-component instead so it can manage by itself even if it doesn't receive any data and doesn't rely on other components handling this?

bundles/statistics/statsgrid/view/Diagram/Diagram.jsx
```diff
@@ -342,16 +343,22 @@ export const Diagram = ({ chartData, sortOrder }) => {
 
     return (
         <Content>
-            <div
-                ref={labelsRef}
-                className='statsgrid-labels'
-            />
-            <Chart>
+            {!chartData?.data ? (
```
ZakarFin, line 346: We should just have another return statement rendering the no data message. It makes it easier to read than having the multiline if-statement in the middle of the JSX.

PR #2478, opened 2023-10-25
Fix populating indicator list (statsgrid search)

Creaticide:
Possibly fixes indicator list not loading correctly.

### DIFF
--- a/bundles/statistics/statsgrid/handler/SearchHandler.js
+++ b/bundles/statistics/statsgrid/handler/SearchHandler.js
@@ -106,51 +106,64 @@ class SearchController extends StateHandler {
         this.updateState({
             loading: true
         });
-        this.service.getIndicatorList(this.state.selectedDatasource, (err, result) => {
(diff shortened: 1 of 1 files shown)


bundles/statistics/statsgrid/handler/SearchHandler.js
```diff
@@ -106,51 +106,64 @@ class SearchController extends StateHandler {
...
+                        if (hasRegionSetRestriction) {
+                            const supportsRegionset = this.state.selectedRegionsets.some((iter) => {
+                                return ind.regionsets.indexOf(Number(iter)) !== -1;
+                            });
+                            if (!supportsRegionset) {
+                                disabledIndicatorIDs.push(ind.id);
+                            }
+                        }
+                        resolve();
                     });
```
ZakarFin, line 137: There's no async operations here and the promises are always resolved. Using promise here just makes this more complex than it needs to be.

PR #2487, opened 2023-11-02
fix featuredata d & d - handle in publisher

DenverCoder544:
add a wrapping div to MapModuleTextButton so publisher can add the dragging handle properly

image

### DIFF
--- a/bundles/mapping/mapmodule/MapModuleTextButton.jsx
+++ b/bundles/mapping/mapmodule/MapModuleTextButton.jsx
@@ -81,16 +81,18 @@ export const MapModuleTextButton = ({ visible, onClick, icon, children, active, 
         return null;
     };
 
-    return <ThemedButton
-        icon={icon || null}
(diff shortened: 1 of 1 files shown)


bundles/mapping/mapmodule/MapModuleTextButton.jsx
```diff
@@ -81,16 +81,18 @@ export const MapModuleTextButton = ({ visible, onClick, icon, children, active,
...
-        icon={icon || null}
-        onClick={onClick}
-        active={active}
-        position={position}
-        loading={loading}
-        {...rest}
-    >
-        {children}
-    </ThemedButton>;
+    return <div>
```
ZakarFin, line 84: Add a comment here that the div is so publisher can have a div-element to attach it's drag handle or something like that. Without any comment I can see myself removing the extra div as cleanup a couple of weeks from now :)

PR #2491, opened 2023-11-06
Fix flyout overflow.

Creaticide:
Fixes vertical scrollbar not appearing when needed.

### DIFF
--- a/src/react/components/window/Flyout.jsx
+++ b/src/react/components/window/Flyout.jsx
@@ -16,7 +16,8 @@ const Container = styled.div`
     min-width: 300px;
     border: 1px solid rgba(0, 0, 0, 0.2);
     max-width: 100vw;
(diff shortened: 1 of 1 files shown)


src/react/components/window/Flyout.jsx
```diff
@@ -16,7 +16,8 @@ const Container = styled.div`
     min-width: 300px;
     border: 1px solid rgba(0, 0, 0, 0.2);
     max-width: 100vw;
-    max-height: 100vh;
+    max-height: 100%;
```
ZakarFin, line 19: This means that the flyout can be bigger than theres space on the screen and we don't really want that.

ZakarFin:
For example thematic maps flyout for selecting indicators to add didn't show a scrollbar on small screens.

PR #2492, opened 2023-11-06
Timeseries year ordering fix.

Creaticide:
Always show earlier year in the "from" field.

### DIFF
--- a/bundles/statistics/statsgrid/handler/SearchHandler.js
+++ b/bundles/statistics/statsgrid/handler/SearchHandler.js
@@ -178,7 +178,11 @@ class SearchController extends StateHandler {
                             searchTimeseries: false
                         });
                     } else {
-                        selected = [selected, selectors[keyWithTime].values[selectors[keyWithTime].values.length - 1].id].sort((a, b) => (b - a));
+                        const selectValues = [
+                            selected,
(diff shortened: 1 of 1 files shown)


bundles/statistics/statsgrid/handler/SearchHandler.js
```diff
@@ -178,7 +178,7 @@ class SearchController extends StateHandler {
                             searchTimeseries: false
                         });
                     } else {
-                        selected = [selected, selectors[keyWithTime].values[selectors[keyWithTime].values.length - 1].id];
+                        selected = [selected, selectors[keyWithTime].values[selectors[keyWithTime].values.length - 1].id].sort((a, b) => (b - a));
```
ZakarFin, line 181: It's a bit hard to read what the data structure is here, but it looks like it's modifying values on the selectors themselves? Meaning the options that are available on the data API. As sorting is done "in place" (/modying the array it's called on) I think we should at least spread the array to make a copy before sorting to make a new list that is presented to the user instead of modifying the "raw data".

ZakarFin: Also do we need to add the selected value to the list again and should the default value be different depending if we are showing the start OR end time?

PR #2498, opened 2023-11-08
Layer rights icons.

Creaticide:
Display layer type icons on layer rights table.

Moved icon components from LayerList to oskari-ui/components/icons.


src/react/components/icons/LayerIcon.jsx
```diff
@@ -24,10 +24,10 @@ export const LayerIcon = ({ type, hasTimeseries = false, additionalTooltipKey, .
         return <DataLayerIcon {...rest} />;
     };
 
-    let tooltipTitle = (<Message messageKey={ `layerTooltipTitle.${type}` } />);
+    let tooltipTitle = (<Message messageKey={ `layerTooltipTitle.${type}`  } bundleKey='LayerList' />);
```
ZakarFin, line 27: We should move the tooltips to the generic translations instead of linking them from under LayerList if the icons are here.

PR #2507, opened 2023-11-13
Featuredata add id column

DenverCoder544:
Always add ID column. Hidden by default,

image

### DIFF
--- a/bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
+++ b/bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
@@ -7,6 +7,8 @@ import { filterFeaturesByPropertyFilter } from '../../../mapping/mapmodule/oskar
 import { cleanFilter } from 'oskari-ui/components/FeatureFilter';
 
 export const FEATUREDATA_DEFAULT_HIDDEN_FIELDS = ['__fid', '__centerX', '__centerY', 'geometry'];
+export const ID_FIELD = '__fid';
+export const ID_FIELD_LABEL = 'ID';
 
 export const SELECTION_SERVICE_CLASSNAME = 'Oskari.mapframework.service.VectorFeatureSelectionService';
 const EXPORT_FEATUREDATA_ROUTE = 'ExportTableFile';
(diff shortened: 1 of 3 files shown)


bundles/framework/featuredata/view/FeatureDataContainer.jsx
```diff
@@ -100,7 +100,7 @@ const createFeaturedataGrid = (features, selectedFeatureIds, showSelectedFirst,
 const createColumnSettings = (selectedFeatureIds, showSelectedFirst, sorting, visibleColumnsSettings) => {
     const { allColumns, visibleColumns, activeLayerPropertyLabels } = visibleColumnsSettings;
     return allColumns
-        .filter(key => !FEATUREDATA_DEFAULT_HIDDEN_FIELDS.includes(key) && visibleColumns.includes(key))
+        .filter(key => (!FEATUREDATA_DEFAULT_HIDDEN_FIELDS.includes(key) || key === ID_FIELD) && visibleColumns.includes(key))
```
ZakarFin, line 103: This is repeated so many times we should probably just add an export for something like `isFieldVisible(field, visibleColumns)` FeatureDataPluginHandler. Then we could comment on it why is ID_FIELD both in FEATUREDATA_DEFAULT_HIDDEN_FIELDS and handled separately?

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
```diff
@@ -7,6 +7,8 @@ import { filterFeaturesByPropertyFilter } from '../../../mapping/mapmodule/oskar
 import { cleanFilter } from 'oskari-ui/components/FeatureFilter';
 
 export const FEATUREDATA_DEFAULT_HIDDEN_FIELDS = ['__fid', '__centerX', '__centerY', 'geometry'];
+export const ID_FIELD = '__fid';
```
ZakarFin, line 10: I'm not sure why these need to be separate but we can use the constant to make it clear it's the same field:
```suggestion
export const ID_FIELD = '__fid';
export const FEATUREDATA_DEFAULT_HIDDEN_FIELDS = [ID_FIELD, '__centerX', '__centerY', 'geometry'];
```

bundles/framework/featuredata/view/FeatureDataContainer.jsx
DenverCoder544: Turns out we don't actually need the (historical) hidden_fields implementation for anything anymore. Removed that and it made things a whole lot simpler.

Also, didn't see the point of artificially cleaning out the id field from the list of properties in "select by properties"-dialog so selecting by id-field is now allowed.

bundles/framework/featuredata/plugin/FeatureDataPluginHandler.js
DenverCoder544: this was more or less a historical remainder. Got rid of it now.

PR #2508, opened 2023-11-13
Statsgrid timeseries checkbox width

Creaticide:


ZakarFin:
Fixes an issue where the timeseries is toggled on/off even when clicking on the empty area after the field. I'm pretty sure that without any comment the extra div will be "cleaned out" at some point. But I don't have any good tips how to make it obvious it's actually needed on the code.

PR #2509, opened 2023-11-13
Handle unknown indicators.

Creaticide:
Show "unknown indicator" text and icon if indicator metadata is not found through API.


bundles/statistics/statsgrid/view/search/IndicatorRow.jsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    return (
+        <Row>
+            <Tooltip title={indicator.labels?.full}>
+                <span>{getIndicatorText(indicator.labels)}</span>
+            </Tooltip>
+            {actions}
+        </Row>
+    );
```
ZakarFin, line 74: We should create a component just to show the indicator name as it's shown in this listing, but also in the table, diagram, classification UI and so on.

PR #2521, opened 2023-11-23
Metadata layout tweak

DenverCoder544:
some margin after the dropdown


ZakarFin:
Should we get rid of the empty file MetadataSearchResultListActionLink.jsx here as well?

PR #2522, opened 2023-11-24
Statsgrid state handling etc.

Creaticide:


ZakarFin:
This is now in its own branch under https://github.com/oskariorg/oskari-frontend/tree/statsgrid-state

PR #2524, opened 2023-11-29
hide announcements tile in mobile mode when there are none

DenverCoder544:


bundles/framework/announcements/instance.js
```diff
@@ -27,15 +29,14 @@ Oskari.clazz.define('Oskari.framework.bundle.announcements.AnnouncementsBundleIn
             if (me.started) {
                 return;
             }
-
+            this.tile = me.plugins['Oskari.userinterface.Tile'] ? me.plugins['Oskari.userinterface.Tile'] : null;
```
ZakarFin, line 32: DefaultExtension has `getTile()` that could be used here. We shouldn't store the reference here but use the getter instead.

PR #2526, opened 2023-12-05
Admin permissions efficiency issues

DenverCoder544:
-changed default page size from 50 to 20. After trying #all tricks in the book that was absolutely the only thing that seemed to make any difference whatsoever in speeding up the antd table re-rendering.
-some tweaking to column widths (layer name a bit wider, thos containing just check boxes narrower)

image


ZakarFin:
It appears there's some recognized issues with the AntD table: https://ant.design/docs/blog/render-times. Maybe we'll try again after updating AntD.

PR #2529, opened 2023-12-07
prevent searching with an empty search term

DenverCoder544:
^nuf said


ZakarFin:
We should check if some of the "advanced search" options have been selected even if there's no query from the free text input. If something/anything has been selected the search should proceed.

PR #2532, opened 2023-12-11
Fix layer icons.

Creaticide:
Show timeseries icon as well on layer permissions.

### DIFF
--- a/bundles/admin/admin-permissions/view/LayerRightsTable.jsx
+++ b/bundles/admin/admin-permissions/view/LayerRightsTable.jsx
@@ -88,6 +88,9 @@ export const LayerRightsTable = ThemeConsumer(({ theme, controller, state }) => 
             });
         return checked;
     };
+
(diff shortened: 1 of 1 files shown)


bundles/admin/admin-permissions/view/LayerRightsTable.jsx
```diff
@@ -97,6 +97,7 @@ export const LayerRightsTable = ThemeConsumer(({ theme, controller, state }) =>
             dataIndex: 'name',
             sorter: getSorterFor('name'),
             render: (title, item) => {
+                const layer = Oskari.getSandbox().getService('Oskari.mapframework.service.MapLayerService').findMapLayer(item.id);
```
ZakarFin, line 100: Get the reference to the MapLayerService before the render. The reference won't change between renders and it can be reused. Probably doesn't affect much but it reduces lookups as render is called multiple times.

PR #2534, opened 2023-12-12
Pinch zoom reset plugin

DenverCoder544:
-add a button for resetting viewport zoom on the map when in mobile mode & the viewport is pinched in
-avoid the situation where the user paints himself in the corner pinch zooming where no other elements besides the map can be seen

image

### DIFF
--- a/bundles/mapping/mapmodule/plugin/pinchzoomreset/PinchZoomResetPlugin.js
+++ b/bundles/mapping/mapmodule/plugin/pinchzoomreset/PinchZoomResetPlugin.js
@@ -124,9 +124,6 @@ Oskari.clazz.define('Oskari.mapframework.bundle.mapmodule.plugin.PinchZoomResetP
                 el[0]
             );
         },
-        teardownUI: function () {
-            this.removeFromPluginContainer(this.getElement());
-        },
         /**
          * @method _stopPluginImpl
          *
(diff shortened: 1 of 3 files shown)


bundles/mapping/mapmodule/plugin/pinchzoomreset/PinchZoomResetPlugin.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+`;
+
+/**
+ * @class Oskari.mapframework.bundle.mapmodule.plugin.PinchZoomReset
+ *
+ * Displays a button to reset pinch zoom on top of the map.
+ *
+ * This solves the situation where the user has pinch zoomed in sot that no other elements besides the map itself are no longer accessible.
```
ZakarFin, line 25:
```suggestion
 * This solves the situation where the user has pinch zoomed in so that no other elements besides the map itself are no longer accessible.
```

bundles/mapping/mapmodule/plugin/pinchzoomreset/PinchZoomResetPlugin.js
```diff
@@ -0,0 +1,163 @@
...
+         */
+        refresh: function () {
+            const el = this.getElement();
+            if (!el) {
+                return;
+            }
+
+            ReactDOM.render(
+                <StyledButton
+                    className='t_fullscreen'
```
ZakarFin, line 115:
```suggestion
                    className='t_pichzoom_reset'
```

PR #2540, opened 2023-12-18
use parseint when determining if visual viewport size != client size

DenverCoder544:
Added parseint for determining if visual viewport size is different from window (desktop emulator worked fine with integers, mobile uses floats). Will hide the zoom reset - tool on android.


ZakarFin:
These don’t work reliably on iOS (resize-event on visual viewport isn’t always triggered on iOS), but looks like even google maps can be zoomed to non-usable state on iOS Safari:

IMG_7360

ZakarFin:
Related to #2534

PR #2543, opened 2023-12-20
Swipe tool to publisher.

Creaticide:


bundles/framework/publisher2/tools/SwipeTool.js
```diff
@@ -0,0 +1,60 @@
...
+        this.handler.init(this.getTool().config);
+    }
+    getValues () {
+        if (!this.isEnabled()) {
+            return null;
+        }
+        const pluginConfig = this.getPlugin().getConfig();
+        const value = {
+            configuration: {
+                mapfull: {
```
ZakarFin, line 39: This should have configuration: { layerswipe: { /* something */ } } instead of mapfull. It should work like maplegend and similar where it's not a mapfull plugin that we are controlling but the whole layerswipe bundle.

bundles/framework/publisher2/handler/AdditionalToolsHandler.js
```diff
@@ -0,0 +1,37 @@
...
+            tools: []
+        });
+    };
+
+    init (data) {
+        const hasTools = super.init(data);
+        return hasTools;
+    }
+
+    stop () {
```
ZakarFin, line 20: Would it work if we added stop() to `ToolPanelHandler` and use it directly with a new instance for additional tools without extending it?

bundles/mapping/layerswipe/publisher/SwipeTool.js
```diff
@@ -0,0 +1,61 @@
...
+        super.stop();
+        this.handler.clearState();
+    }
+}
+
+// Attach protocol to make this discoverable by Oskari publisher
+Oskari.clazz.defineES('Oskari.LayerSwipe.SwipeTool',
+    SwipeTool,
+    {
+        'protocol': ['Oskari.mapframework.publisher.AdditionalTool']
```
ZakarFin, line 59: We should be able to use the common tool protocol/interface so we don't need a new one as `AdditionalTool`
```suggestion
        'protocol': ['Oskari.mapframework.publisher.Tool']
```

bundles/mapping/mapmodule/AbstractMapModule.js
```diff
@@ -29,6 +29,7 @@ import './plugin/realtime/RealtimePlugin';
 import './plugin/mylocation/MyLocationPlugin';
 import './plugin/zoombar/Portti2Zoombar';
 import './plugin/panbuttons/PanButtons';
+import '../layerswipe/plugin/LayerSwipePlugin';
```
ZakarFin, line 32: We don't need to import it here as it is it's own bundle. We can just start the layerswipe bundle instead on the embedded map.

src/react/components/icons/Swipe.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export const Swipe = () => (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32">
+        <line x1="15" y1="24" x2="16.5" y2="24" fill="none" stroke="#000" strokeLinejoin="bevel" strokeWidth="1.75"/>
```
ZakarFin, line 5: We'll need to have the colors defined with css/styled-components intead (`stroke="#000"` etc).

bundles/framework/publisher2/view/PanelAdditionalTools.js
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+import { LocaleProvider } from 'oskari-ui/util';
+import ReactDOM from 'react-dom';
+import { AdditionalTools } from './form/AdditionalTools';
```
ZakarFin, line 4: We can just use PublisherToolsList component directly without wrapping it here.

PR #2544, opened 2023-12-20
Swipe feature url param

DenverCoder544:
-Add layerSwipe = true to created link when said tool is activated.
-Turn on layerSwipe by url param.

image

### DIFF
--- a/bundles/mapping/layerswipe/instance.js
+++ b/bundles/mapping/layerswipe/instance.js
@@ -55,6 +55,9 @@ Oskari.clazz.define(
                 }
             };
             sandbox.request(this, addToolButtonBuilder('LayerSwipe', 'basictools', buttonConf));
+            if (this.state?.layerSwipe) {
+                this.setActive(true);
+            }
         },
 
         setActive: function (active) {
--- a/bundles/mapping/layerswipe/instance.js
+++ b/bundles/mapping/layerswipe/instance.js
@@ -55,7 +55,7 @@ Oskari.clazz.define(
                 }
             };
             sandbox.request(this, addToolButtonBuilder('LayerSwipe', 'basictools', buttonConf));
-            if (this.state?.layerSwipe) {
+            if (this.state?.active) {
                 this.setActive(true);
             }
         },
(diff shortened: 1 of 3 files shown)


ZakarFin:
There are some changes to layerswipe state handling (`state.active`) and activating it on startup on #2543. Lets wait merging this until that one is completed.

ZakarFin:
I think we should make layerswipe stateful and implement the getStateParams() function for it instead of adding the param to link generating function. But there are a bunch of changes for layerswipe and state handling already on https://github.com/oskariorg/oskari-frontend/tree/embedded-layerswipe so it's best to wait until that one is compeleted before merging or changing these.

bundles/mapping/layerswipe/instance.js
```diff
@@ -55,6 +55,9 @@ Oskari.clazz.define(
                 }
             };
             sandbox.request(this, addToolButtonBuilder('LayerSwipe', 'basictools', buttonConf));
+            if (this.state?.active) {
+                this.setActive(true);
+            }
```
ZakarFin, line 60: This is probably no longer needed after #2550?

bundles/mapping/toolbar/default-buttons.js
```diff
@@ -229,13 +229,14 @@ Oskari.clazz.category(
             const linkParams = sandbox.generateMapLinkParameters({});
             const viewUuid = this._getLinkUuid();
 
+            const layerSwipeActive = sandbox?.findRegisteredModuleInstance('LayerSwipe')?.active;
```
ZakarFin, line 232: And this is outdated as well after #2550. We should implement the getStateParameters() function for layerswipe instead as it is now stateful. This way we don't need to handle it separately on the link creation.

ZakarFin:
The parameter is now "swipe" on oskari-server: https://github.com/oskariorg/oskari-server/blob/develop/control-base/src/main/java/fi/nls/oskari/control/view/modifier/param/LayerSwipeParamHandler.java#L13

DenverCoder544:
Made a new pr on top of recent other changes on the matter, this pr is now obsolete.

https://github.com/oskariorg/oskari-frontend/pull/2551

PR #2546, opened 2023-12-22
add fallback for getting the visible columns from feature properties

DenverCoder544:
For some reason for some layers layer.properties is an empty array and in that case we gotta resort to getting the columns from feature.properties


ZakarFin:
Waiting to merge this until we can figure out why the layer.getProperties() is empty in the response from https://github.com/oskariorg/oskari-server/blob/master/control-base/src/main/java/org/oskari/control/layer/DescribeLayerHandler.java#L128

ZakarFin:
This makes it work in a similar way to the "GFI-popup" when clicking the map. So merging the change. We can still see from the server output if the properties are being passed for the layer or not and continue debugging.

PR #2550, opened 2024-01-09
Add swipetool to publisher #2

DenverCoder544:


bundles/mapping/layerswipe/handler/SwipeToolHandler.js
```diff
@@ -0,0 +1,65 @@
...
+    clearState () {
+        // plugin is created again on startup, so it's state doesn't need to be cleared
+        this.setState({
+            autoStart: false,
+            noUI: false
+        });
+    }
+    /*
+    onLayersChanged () {
+        // we could do setEnabled(false) here if Oskari.getSandbox().getMap().getLayers() is < 2
```
ZakarFin, line 38: We should add a ticket to implement this: if user removes layers while in publisher -> disable this if there aren't enough layers to compare? And maybe check that this can't be enabled with just one layer?

bundles/mapping/layerswipe/instance.js
```diff
@@ -39,25 +39,44 @@ Oskari.clazz.define(
 
         _startImpl: function (sandbox) {
             this.mapModule = Oskari.getSandbox().findRegisteredModuleInstance('MainMapModule');
+            this.publisherInstance = Oskari.getSandbox().findRegisteredModuleInstance('Publisher2');
```
ZakarFin, line 42: If layerswipe is started before publisher, this will be "undefined". We should fetch the reference when we need to do something with it to make sure it works even with funky bundle sort order.

bundles/mapping/layerswipe/plugin/LayerSwipePlugin.js
```diff
@@ -0,0 +1,154 @@
...
+                // no point in drawing the ui if we are not visible
+                return;
+            }
+
+            this.teardownUI();
+            this._element = this._createControlElement();
+            this.refresh();
+            this.addToPluginContainer(this._element);
+        },
+*/
```
ZakarFin, line 116: Remove the commented out code.

src/react/components/icons/index.js
```diff
@@ -12,3 +12,4 @@ export { Drag as DragIcon } from './Drag';
 export { LayerIcon } from './LayerIcon';
 export { EyeOpen } from './EyeOpen';
 export { EyeShut } from './EyeShut';
+export { Swipe as SwipeIcon } from './Swipe';
```
ZakarFin, line 15: Lets move the icon under layerswipe bundle.

bundles/mapping/layerswipe/handler/SwipeToolHandler.js
```diff
@@ -0,0 +1,65 @@
...
+
+    init (pluginConfig) {
+        this.updateState({
+            ...pluginConfig.conf,
+            autoStart: pluginConfig?.state?.active,
+            noUI: pluginConfig?.conf?.noUI
+        });
+    }
+
+    syncState () {
```
ZakarFin, line 22: We probably don't need this anymore? Check if it can be removed.

bundles/mapping/layerswipe/publisher/SwipeTool.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    init (data) {
+        const configuration = data?.configuration?.layerswipe || {};
+        // restore state to handler -> passing init data to it
+        this.handler.init(configuration);
+        if (configuration.conf) {
+            this.storePluginConf(configuration.conf);
+            this.setEnabled(true);
+            this.handler.syncState();
```
ZakarFin, line 37: Check if we can remove syncState() call here.

bundles/framework/publisher2/instance.js
```diff
@@ -280,11 +280,14 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.PublisherBundleInstan
...
             }
             if (blnEnabled) {
-                if (me.publisher) return;
+                if (me.publisher?.isEnabled()) return;
                 data = data || this.getDefaultData();
                 // trigger an event letting other bundles know we require the whole UI
+                // the latter SetStateRequest triggers UIChangeEvent as well so this is considered to be redundant
+                /*
                 var eventBuilder = Oskari.eventBuilder('UIChangeEvent');
                 this.sandbox.notifyAll(eventBuilder(this.mediator.bundleId));
```
ZakarFin, line 289: Lets check if `data.configuration` is falsy or an object without keys and send this instead of StateHandler.SetStateRequest if it is. As SetState wont do the reset if it receives an empty object or falsy value as state

bundles/mapping/layerswipe/handler/SwipeToolHandler.js
DenverCoder544: we do, unfortunately. The plugin might not exist yet where we could otherwise set the state, so this is needed for the ui state & the plugin state to stay 'n sync.

bundles/mapping/layerswipe/publisher/SwipeTool.js
DenverCoder544: Checked. Can't. Bummer.

bundles/framework/publisher2/view/PublisherSidebar.js
```diff
@@ -466,6 +487,9 @@ Oskari.clazz.define('Oskari.mapframework.bundle.publisher2.view.PublisherSidebar
             }
         },
 
+        isEnabled: function () {
+            return this._enabled;
+        },
```
ZakarFin, line 492: We probably don't need this._enabled or anything using it anymore?

bundles/mapping/layerswipe/plugin/LayerSwipePlugin.js
```diff
@@ -0,0 +1,135 @@
...
+            this.refresh();
+        },
+        isActive: function () {
+            return !!this.getInstance()?.isActive();
+        },
+        resetUI: function () {
+        },
+        getInstance: function () {
+            // we only need instance here since the flyout is operated by the instance
+            // TODO: we should move the flyout related code to this plugin
```
ZakarFin, line 55:
```suggestion
            // We need instance as it manages the `active` state.
```

bundles/mapping/layerswipe/publisher/SwipeTool.js
```diff
@@ -0,0 +1,76 @@
...
+            component: SwipeToolComponent,
+            handler: this.handler
+        };
+    }
+
+    init (data) {
+        const configuration = data?.configuration?.layerswipe || {};
+        if (configuration?.state?.active) {
+            this.setEnabled(true);
+        }
```
ZakarFin, line 34: setEnabled(true) is now called here and a few lines after again. Can we remove the other/this one?

bundles/mapping/toolbar/instance.js
```diff
@@ -340,8 +340,8 @@ Oskari.clazz.define('Oskari.mapframework.bundle.toolbar.ToolbarBundleInstance',
                 sandbox.request(me, Oskari.requestBuilder('ShowMapMeasurementRequest')(msg || '', false, null, null));
             },
             'UIChangeEvent': function (evt) {
-                // select default tool
-                this.getSandbox().postRequestByName('Toolbar.SelectToolButtonRequest');
+                // select default tool.
```
ZakarFin, line 343:
```suggestion
                // select default tool to reset UI state.
```

bundles/framework/publisher2/view/PublisherSidebar.js
DenverCoder544: We don't. Just scrap left over from two or three iterations ago 😛

PR #2551, opened 2024-01-12
add layerswipe to created link

DenverCoder544:
Add swipe to url when creating a link with link tool. Also, do not deactivate layerswipe-tool when either link-tool or save map view are clicked in toolbar.

When the tool is activated via a alink it doesn't currently get activated in the toolbar becaus of some default logic for selecting the hand tool. This is "a bit" confusing but we need to think how this _should_ work in case there were multiple tools activated via the link.

This pr replaces https://github.com/oskariorg/oskari-frontend/pull/2544, which can be closed after this one has been merged.

### DIFF
--- a/bundles/mapping/layerswipe/instance.js
+++ b/bundles/mapping/layerswipe/instance.js
@@ -290,6 +290,9 @@ Oskari.clazz.define(
         eventHandlers: {
             'Toolbar.ToolSelectedEvent': function (event) {
                 if (event.getToolId() !== 'LayerSwipe' && event.getToolId() !== 'link' && event.getToolId() !== 'save_view') {
+                    // This bundle generates state for both link and saving views, but only if it's active.
+                    // If we deactivate when link or save view tools are clicked the state is not stored as expected.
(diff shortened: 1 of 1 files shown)


bundles/mapping/layerswipe/instance.js
```diff
@@ -281,7 +289,7 @@ Oskari.clazz.define(
 
         eventHandlers: {
             'Toolbar.ToolSelectedEvent': function (event) {
-                if (event.getToolId() !== 'LayerSwipe') {
+                if (event.getToolId() !== 'LayerSwipe' && event.getToolId() !== 'link' && event.getToolId() !== 'save_view') {
```
ZakarFin, line 292:
```suggestion
                if (event.getToolId() !== 'LayerSwipe' && event.getToolId() !== 'link' && event.getToolId() !== 'save_view') {
                    // This bundle generates state for both link and saving views, but only if it's active.
                    // If we deactivate when link or save view tools are clicked the state is not stored as expected.
                    // So we need to keep the swipe tool active when these tools are clicked.
```

PR #2554, opened 2024-01-17
Antd v5 migration

DenverCoder544:
Initial pr where the packages have been updated and the most obvious failures have already been fixed.
* updated packages 
* fixed the console errors that were available straight from the get-go, upon app init
* some css fixes which were visible on the first glimpse
* started the "children to items"-saga with announcements bundle. Sequels on that are on their way.

### DIFF
--- a/src/react/components/InputGroup.jsx
+++ b/src/react/components/InputGroup.jsx
@@ -1,4 +1,4 @@
-import { Input } from 'antd';
+import { Space } from 'antd';
 import 'antd/es/input/style/index.js';
 
-export const InputGroup = Input.Group;
+export const InputGroup = Space.Compact;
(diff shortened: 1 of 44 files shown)


src/react/components/InputGroup.jsx
```diff
@@ -1,4 +1,3 @@
-import { Input } from 'antd';
-import 'antd/es/input/style/index.js';
+import { Space } from 'antd';
 
-export const InputGroup = Input.Group;
+export const InputGroup = Space.Compact;
```
ZakarFin, line 3: Is this copy/paste or valid?

DenverCoder544: It's valid as Input.Group was deprecated and there was a warning to replace it with Space.Compact.

PR #2557, opened 2024-01-18
notifications dialog - collapse uses items instead of children

DenverCoder544:


bundles/framework/announcements/view/flyout/FlyoutCollapse.jsx
```diff
@@ -13,23 +13,26 @@ export const FlyoutCollapse = ({
...
-                        extra={<CollapseTools toolController={toolController} announcementId={id}/>}>
-                        <AnnouncementsContent announcement={announcement}/>
-                        <Divider />
-                        <b><Message messageKey={'valid'} /></b>
-                        <p>{dateRange}</p>
-                    </CollapsePanel>
-                );
-            })}
-        </Collapse>
+        <Collapse accordion items={items}/>
```
ZakarFin, line 35: Do these have anything like renderItem-function available? We could try to change these to utilize that instead if one is available. Probably not any issue here, but some parts might benefit if we do the data-> jsx on renderItem "lazily" instead of eagerly generating the jsx for the items array.

ZakarFin: Specifically the permissions tab on admin-layereditor can be very heavy to render

DenverCoder544: Fortunately that one doesn't have collapses 😛 

Lazy rendering seems to be the default, i.e. when the collapse is closed the content isn't being rendered unless forceRender is set.

PR #2559, opened 2024-01-19
printoutpanel collapses to use items

DenverCoder544:


ZakarFin:
We should try making this a bit more readable with the new items-prop. Maybe separating the panels to own files that export both label and contents?

bundles/framework/printout/view/items/BasicViewSettingsCollapseItem.jsx
```diff
@@ -0,0 +1,22 @@
...
+const BasicViewPanelContent = ({ state, controller }) => {
+    return <AdditionalSettings state={state} controller={controller} />;
+};
+
+BasicViewPanelContent.propTypes = {
+    state: PropTypes.object.isRequired,
+    controller: PropTypes.object
+};
+
+export const getBasicViewCollapseItem = (key, state, controller) => {
```
ZakarFin, line 16: This is mostly a wrapper for AdditionalSettings. We should just rename `AdditionalSettings` to `BasicViewPanelContent` and add `getBasicViewCollapseItem()` to it.

ZakarFin: Lets use the AdditionalSettings as name and add the getter getAdditionalSettingsCollapseItem () to it and remove the BasicViewSettingsCollapseItem.jsx

PR #2565, opened 2024-01-25
Historic ortophotos timeseries mobile tweak

DenverCoder544:
Add a dropdown for selecting the year in mobile mode

image


ZakarFin:
Looks good, but we might need to add the "active year" to the dropdown as the available years change when the user pans the map and the selected year might not be one of the options after panning.

PR #2566, opened 2024-01-29
Add coordinates to GFI popup

osqq:
Show coordinate projection and the coordinates of a clicked point on map in GFI popup.

Requires _showCoordinates: true_ to be added to mapfull GetInfoPlugin cofig.

### DIFF
--- a/bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol.js
+++ b/bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol.js
@@ -165,7 +165,7 @@ Oskari.clazz.define(
             var me = this,
                 contentDiv = me._renderContentData(id, contentData),
                 sanitizedTitle = Oskari.util.sanitize(title),
-                popupContentHtml = me._renderPopupContent(id, sanitizedTitle, contentDiv, additionalTools,lonlat, options.showCoordinates),
+                popupContentHtml = me._renderPopupContent(id, sanitizedTitle, contentDiv, additionalTools, lonlat, options.showCoordinates),
                 popupElement = me._popupWrapper.clone(),
                 lonlatArray = [lonlat.lon, lonlat.lat],
                 colourScheme = options.colourScheme,
@@ -377,11 +377,10 @@ Oskari.clazz.define(
 
             headerWrapper.append(popupHeaderTitle);
 
-            // do we show coordinates?
+            // render coordinates to gfi header 
             if (showCoordinates) {
                 let mapModule = this.getMapModule();
                 let loc = Oskari.getLocalization('coordinatetool');
-
                 let crs = mapModule.getProjection();
 
                 let coordinateWrapper = jQuery('<div class="coordinateWrapper"></div>');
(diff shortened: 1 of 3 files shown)


bundles/mapping/infobox/resources/scss/infobox.ol.scss
```diff
@@ -127,20 +127,39 @@ div.popupHeader {
     margin: 0px;
     @include border-radius(6px 6px 0px 0px);
     display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    width: 390px;
```
ZakarFin, line 132: Can you test/provide some screenshots how this looks with a wide infobox? A long title or the content shown on the infobox might make the infobox wider than this. Does this need a set width?

osqq: Actually these were not actually added even tho git seems to think they are, only the flex direction is changed. But on that note the width needs to be removed from the class "popupHeaderTitle". The width on this popupHeader class seems to be overwritten by something at least on my local environment (so it might not be needed anyways). I can provide couple more screenshots.

osqq: [Line 565](https://github.com/oskariorg/oskari-frontend/blob/cb979b49bef1fc2197a7061fc14ea100f419d07b/bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol.js#L565) this seems to (always?) make the width '100%' so should the width be removed from the css or make it '100%', or might there be something that 390px starting point is for?

osqq:
Found that removing a layer from the selected makes the coords disappear so need to fix that.

bundles/mapping/infobox/resources/scss/infobox.ol.scss
ZakarFin: Oh, true. Most instances probably use the "adaptable: true" config for this. Well if the width isn't new addition I think it can stay there.

bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol.js
```diff
@@ -361,9 +373,53 @@ Oskari.clazz.define(
...
-                headerWrapper.append(additionalButton);
+                popupHeaderTitle.append(additionalButton);
             });
 
+            headerWrapper.append(popupHeaderTitle);
+
+            // render coordinates to gfi header
+            if (showCoordinates) {
+                let mapModule = this.getMapModule();
+                let loc = Oskari.getLocalization('coordinatetool');
```
ZakarFin, line 384: This assumes that we have coordinatetool bundle as part of the app. Maybe just copy/paste loc.display.compass.lat/lon/n/e to the shared oskariui locale-files instead (https://github.com/oskariorg/oskari-frontend/tree/master/bundles/framework/oskariui/resources/locale). Or move and modify coordinate tool to use these as well?

bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol.js
```diff
@@ -526,13 +582,13 @@ Oskari.clazz.define(
...
                 } else {
-                    const { colourScheme, font, title, lonlat } = popup;
+                    const { title, lonlat, options } = popup;
                     this._renderPopup(
                         popupId,
                         contentData,
                         title,
                         lonlat,
-                        { colourScheme, font },
+                        options,
```
ZakarFin, line 591: Just to make sure we don't try to access the keys on undefined object:
```suggestion
                        options || {},
```

bundles/mapping/infobox/plugin/openlayerspopup/OpenlayersPopupPlugin.ol.js
ZakarFin: Oh, the crs part is needed as well. But oskariui could have something like this:
```
            "coordinates": {
                "lon": "Lon",
                "lat": "Lat",
                "n": "N",
                "e": "E",
                "p": "N",
                "i": "E", 
                "crs": {
                  "EPSG:3067": "ETRS89-TM35FIN coordinates",
                  "EPSG:3575": "North Pole LAEA Europe coordinates",
                  "EPSG:3857": "WGS 84 / Pseudo-Mercator coordinates",
                  "default": "{crs} coordinates"
              },
            },
```

PR #2567, opened 2024-01-30
Historic ortophotos timeseries mobile tweak

DenverCoder544:
Added a new variable to the navigation theme for bg color. This needs to be added to the theme of the app using the historical ortophotos timeseries. For instance in paikkis ->

Oskari.app.getTheming().setTheme({
    "color": {
        "icon": "#3c3c3c",
        "accent": "#ffd400",
        "primary": "#fdf8d9",
        "text": "#3c3c3c"
    },
    "map": {
        "color": {
            "header": {
                "bg": "#fdf8d9",
                "icon": "#3c3c3c",
                "text": "#3c3c3c"
            },
            "accent": "#ffd400"
        },
        "navigation": {
            "color": {
                "text": "#FFFFFF",
                "accent": "#ffd400",
                "primary": "#141414",
                "bg": "#3c3c3c"
            },
            "roundness": 100,
            "opacity": 0.8
        },
        "font": "arial"
    }
});

Paikkis without setting the new variable (using the buttons' default background):
image

With the variable set to something lighter: 
image

...and with ASDI theme and no navigation bg set the result is still readable
image

### DIFF
--- a/bundles/framework/timeseries/view/TimeSeriesRange/TimeSeriesHeader.jsx
+++ b/bundles/framework/timeseries/view/TimeSeriesRange/TimeSeriesHeader.jsx
@@ -48,7 +48,7 @@ export const TimeSeriesHeader = ({ toggleMode, title, mode = 'year', loading = f
     const switchButtonMessage = <Message messageKey={switchButtonMessageKey} />;
     const modeIcon = mode === 'year' ? <LoginOutlined /> : <LogoutOutlined />;
     return (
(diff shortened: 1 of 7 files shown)


ZakarFin:
This adds new theme variable for navigation.color.bg that defaults to navigation.color.primary, but can be used to override the background for the main navigation and controls like these that are opened on top of the map (that don't use transparency like the buttons that make them look lighter).

PR #2575, opened 2024-02-16
Statsgrid state refactor

okauppinen:
Refactoring state:
- Use one handler (`StateHandler`) for maintaining state => remove `ClassificationHandler`,  `TableHandler` and `DiagramHandler`
- Use one handler (`ViewHandler`) for views => remove FlyoutManager, show/close from other handlers
- Gather all stuff  needed to render indicator (data, classification, classified data, labels,..) before adding indicator to state
- Store all indicator related stuff under it
- Remove Stastgrid events

Other Changes:
- Remove `ClassificationPlugin` as is was needed for rendering content to map plugin container. Now React movableContainer is used so no need for plugin.
- Refactor SeriesControl to use internal state and state handler instead of SeriesService => remove `SeriesService`
- Use same keys that stored state uses => indicator => id, datasource => ds, table => grid (search flyout and handler uses still datasource and indicator in same places)
- Copy locales, styles icons and publisher tools => new statsgrid doesn't use files under /statsgrid2016 anymore
- Clean bundles.js

Maybe we should remove StatisticsService and cache. Now data is requested from service less frequently => caching raw data and  addToQueue aren' t relevant any more. Also service contains backend calls + cache only.
=> move all backend calls to helpers and cache data in helpers


ZakarFin:
I guess we'll just need to merge this for testing, pretty difficult to see the changes through diffs 😅

PR #2594, opened 2024-03-04
Optional metadata url

osqq:
Allow user to add "metadataUrl" parameter to layer attributes for optional metadata url for that layer. If no optional url is set, we default to the one set in oskari-ext.properties.

Note! The ID used in the formed "GetCSWData" request is read from the layer administration's "Metadata file identifier". So in order to use optional metadata, set the url in attributes and the ID in "Replaced file identifier".

`{
"metadataUrl": "https://exampleurl.fi/metadata/csw?"
}`

relates to: https://github.com/oskariorg/oskari-server/pull/1052

### DIFF
--- a/bundles/catalogue/metadataflyout/service/MetadataLoader.js
+++ b/bundles/catalogue/metadataflyout/service/MetadataLoader.js
@@ -19,7 +19,7 @@ Oskari.clazz.define(
          * @param {function} errorCallback   Error callback function
          *
          */
-        getCSWData: function (uuid, lang, successCallback, errorCallback) {
+        getCSWData: function (uuid, lang, metadataUrl, successCallback, errorCallback) {
             var err = this._checkArgs(
                     uuid,
                     lang,
@@ -34,6 +34,10 @@ Oskari.clazz.define(
             uri = this.baseUrl + 'action_route=GetCSWData&uuid=' + uuid +
                 '&lang=' + lang +
                 '&srs=' + this.srs;
+
+            if (metadataUrl !== null) {
+                uri = uri + '&metadataUrl=' + metadataUrl;
+            }
 
             jQuery.ajax({
                 url: uri,
--- a/bundles/catalogue/metadataflyout/service/MetadataLoader.js
+++ b/bundles/catalogue/metadataflyout/service/MetadataLoader.js
@@ -14,14 +14,16 @@ Oskari.clazz.define(
          * @public @method getCSWData
          *
          * @param {string}   uuid            Metadata UUID
+         * @param {string}   layerId         Layer id
          * @param {string}   lang            Metadata language
          * @param {function} successCallback Success callback function
          * @param {function} errorCallback   Error callback function
          *
          */
-        getCSWData: function (uuid, lang, metadataUrl, successCallback, errorCallback) {
+        getCSWData: function (uuid, layerId, lang, successCallback, errorCallback) {
             var err = this._checkArgs(
                     uuid,
+                    layerId,
                     lang,
                     successCallback,
                     errorCallback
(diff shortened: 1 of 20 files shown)


osqq:
Is "metadataUrl" acceptable parameter name or should it be something like "optionalMetadataUrl"?

bundles/catalogue/metadataflyout/service/MetadataLoader.js
```diff
@@ -35,6 +35,10 @@ Oskari.clazz.define(
                 '&lang=' + lang +
                 '&srs=' + this.srs;
 
+            if (metadataUrl !== null) {
+                uri = uri + '&metadataUrl=' + metadataUrl;
+            }
```
ZakarFin, line 40: Sending an URL to be called is problematic from security perspective. We should just send the layer id instead (removing the uuid) and let server get the optional address and the uuid based on the layer id.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/Layer/LayerTools.jsx
```diff
@@ -15,6 +15,7 @@ const Tools = styled('div')`
     }
 `;
 
+//remove?
 const hasSubLayerMetadata = (layer) => {
```
osqq, line 19: This seems to be unused, remove?

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
```diff
@@ -237,8 +237,10 @@ class ViewHandler extends StateHandler {
         this.updateState({ selectedLayerIds });
     }
 
+    // Is this deprecated?
```
osqq, line 240: This seems to also not be used anywhere, remove?

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/Layer/LayerTools.jsx
ZakarFin: Nice catch. Looks like it isn't exported and not used on the file itself so you can go ahead and remove this function.

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
ZakarFin: It might be that the relevant logic has been moved to the MetadataIcon component. So if this is not called anywhere it can be removed.

ZakarFin:
Can you update this a bit as well (at least add a note that layer id is also an option): https://github.com/oskariorg/oskari-frontend/blob/master/api/mapping/metadataflyout/request/catalogue.showmetadatarequest.md

osqq:
> Can you update this a bit as well (at least add a note that layer id is also an option): https://github.com/oskariorg/oskari-frontend/blob/master/api/mapping/metadataflyout/request/catalogue.showmetadatarequest.md

Ah yes, I intended to do that but forgot!

bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapseHandler.js
osqq: The only reference to the same named function is in admin-layereditor, where we have a different function with the same name (in AdminLayerFormHandler.js). I'll remove it.

ZakarFin:
Thanks :)

ZakarFin:
Looks like this is not working properly on the metadata search:
image

^ The uuid is sent as layerId from `metadatasearch` bundle when clicking the search result.

PR #2597, opened 2024-03-05
Statsgrid2016 from master

okauppinen:
Revert stasgrid2016 changes from master


ZakarFin:
This restores the current statsgrid functionality to the level it was on the 2.12 release so it can be used as an alternative to the new one if we discover serious problems with the new one (react ilmplementation) after the release.

PR #2621, opened 2024-03-13
Statsgrid active region

okauppinen:
- show selected region in table
- select active region on table row click
- toggle selected region
- use theme color for active indicator (header) and selected region (row)
- handle active region change more reliable in regionset viewer

image

### DIFF
--- a/bundles/statistics/statsgrid/handler/StatisticsHandler.js
+++ b/bundles/statistics/statsgrid/handler/StatisticsHandler.js
@@ -109,8 +109,12 @@ class StatisticsController extends StateHandlerBase {
         return updated;
     }
 
-    setActiveRegion (activeRegion) {
-        this.updateState({ activeRegion });
+    setActiveRegion (value) {
+        // toggle if already selected
+        const activeRegion = this.getState().activeRegion === value ? null : value;
+        // don't use normal updateState and notify with updated key to optimize map rendering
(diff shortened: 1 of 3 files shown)


bundles/statistics/statsgrid/handler/StatisticsHandler.js
```diff
@@ -109,8 +109,12 @@ class StatisticsController extends StateHandlerBase {
...
     }
 
-    setActiveRegion (activeRegion) {
-        this.updateState({ activeRegion });
+    setActiveRegion (value) {
+        // toggle if already selected
+        const activeRegion = this.getState().activeRegion === value ? null : value;
+        // don't use normal updateState and notify with updated key to optimize map rendering
+        this.state = { ...this.getState(), activeRegion };
+        this.stateListeners.forEach(consumer => consumer(this.getState(), 'activeRegion'));
```
ZakarFin, line 117: This is a bit wild, but I don't have a better suggestion yet. Maybe we should add some other listener queue that RegionsetViewer could listen to instead, but lets let this simmer a bit and get back to it after 2.13.

ZakarFin: Suggested changes in https://github.com/okauppinen/oskari/pull/2

PR #2623, opened 2024-03-15
Add mentions of new/changed bundles in changelog

ZakarFin:
These are already mentioned in release notes, but since changelog is more about what if bundles/requests/events aka the API between bundles changes I think these should be mentioned here. We should figure out where to put these kinds of things so we don't need to repeat it everywhere and people know where to look for these kinds of changes.

Also we should figure out if we want to change the bundle id when it's a drop-in replacement or have the bundle.js in new path like with statsgrid.

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -1,4 +1,3 @@
-
 # Oskari API changelog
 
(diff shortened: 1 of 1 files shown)


ZakarFin:
In theory the bundle id shouldn't change if the API doesn't change (requests,events, conf, state, exposed services), but I think we did just the opposite here with changing the bundle ids where API didn't change and kept it with statsgrid that removed the events... :)

PR #2625, opened 2024-03-18
avoid caret jump in async state update

DenverCoder544:
do a sync state update before calling the (possibly) async one to avoid the caret jumping to end of input.

This does appear a bit hacky and also there might be some pitfall if the value is to be manipulated somehow in the latter onchange.  🤔

### DIFF
--- a/src/react/components/TextInput.jsx
+++ b/src/react/components/TextInput.jsx
@@ -1,6 +1,19 @@
 import styled from 'styled-components';
 import { Input } from 'antd';
 import 'antd/es/input/style/index.js';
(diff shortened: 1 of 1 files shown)


ZakarFin:
Related to the async statehandler:
- https://github.com/oskariorg/oskari-frontend/pull/2595
- https://github.com/oskariorg/oskari-frontend/pull/2602

We might want to change the StateHandler back to being sync instead and add an AsyncStateHandler that statsgrid could use.

ZakarFin:
Solved in #2626

PR #2630, opened 2024-03-27
Fine tune statsgrid async state to fix editing embedded map

ZakarFin:
Currently editing embedded maps results in js max stack error. This should fix it


ZakarFin:
This is related to #2626

PR #2637, opened 2024-04-19
Style effect support alpha

okauppinen:
OpenLayers supports [r,g,b(,a)] color so created first colorToArray. As theme and other components uses also getColorEffect changed util.getColorEffect to return hex  (has always `#` or rgba. For now alter

Not sure where EFFECT constant should be located. If better place isn't found and more constants are added, maybe some alias like oskari-ui could be added for cleaner imports.

Only getColorEffect (mapmodule, theme) used Oskari.util.alterBrightness. New util.getColorEffect doesn't use it because it doesn't support alpha and returns hex so maybe it could be removed or mark as deprecated.

Changes:
- mapmodule had (effect, color) and theme (color, effect)
- getColorEffect returns passed color if fails or no effect to apply (undefined, none, 0).
- getColorEffect returns always #-prefixed hex or rgba if alpha exists

Not sure how intended is to test returning black wihtout amount:
https://github.com/oskariorg/oskari-frontend/blob/2.13.1/src/util.test.js#L192-L200
Because red, green and blue gets NaN values
https://github.com/oskariorg/oskari-frontend/blob/2.13.1/src/util.test.js#L192-L200
And:
https://github.com/oskariorg/oskari-frontend/blob/2.13.1/src/util.test.js#L192-L200
returns 0 which is padded with leaading zeroes to 000000.
If we want black color for fallback/default, then it should be more visible/clearer.

Also getColorBrighness requires #-prefix for hex color but many other function doesn't.
https://github.com/oskariorg/oskari-frontend/blob/2.13.1/src/util.test.js#L192-L200

### DIFF
--- a/bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/VectorStyle/HoverModal.jsx
+++ b/bundles/admin/admin-layereditor/view/AdminLayerForm/VisualizationTabPane/VectorStyle/HoverModal.jsx
@@ -9,7 +9,7 @@ import { IconButton } from 'oskari-ui/components/buttons';
 import { Controller } from 'oskari-ui/util';
 import { getGeometryType } from '../../../LayerHelper';
 import { SUPPORTED_FORMATS } from 'oskari-ui/components/StyleEditor/constants';
-import { EFFECT } from '../../../../../../mapping/mapmodule/oskariStyle/constants.js';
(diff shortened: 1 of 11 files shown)


ZakarFin:
Still not sure if src/constants.js is where the color constants should be but this removes duplicated code from theme/map-colors so the result is better anyway. We can tune it later on if needed.

PR #2640, opened 2024-05-21
Statistics clipboard import region id

okauppinen:
Add check for region id


ZakarFin:
Previously regions could be referenced with name or id when importing from clipboard, but on the new thematic maps implementation the id references were not workin and name had to be used. This restores the functionality for using id to reference regions while importing from clipboard.

PR #2644, opened 2024-05-28
Reset cursor for olPopups to override map cursor setting

ZakarFin:
Fixes an issue where for example an RPC-application sets the map cursor with: https://oskari.org/examples/rpc-api/#/setCursorStyle

And opens an infobox with https://oskari.org/examples/rpc-api/#/infoBoxSimple

This results in the cursor being the one set for map even when hovering over an infobox.


ZakarFin:
image

PR #2645, opened 2024-05-28
Replace draft-js with Jodit

osqq:
Draft-js is deprecated and archived text editor, so replace it with Jodit.

### DIFF
--- a/bundles/framework/announcements/resources/scss/announcementsplugin.ol.scss
+++ b/bundles/framework/announcements/resources/scss/announcementsplugin.ol.scss
@@ -26,6 +26,7 @@ div.announcements-content {
 }
 
 div.mapplugin.announcements {
+    padding: 10px 10px 0 10px;
     clear:both;
     height: 37px;
     overflow: visible;
@@ -157,6 +158,7 @@ div.mapplugin.announcements {
     }
       
     .announcement-content {
+        text-align: left;
         max-height: 0;
         overflow: hidden;
         transition: padding 0.2s ease-out;
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "imports-loader": "0.8.0",
     "intl-messageformat": "9.11.4",
     "jest": "27.0.6",
+    "jodit-react": "^4.1.2",
     "jquery": "3.7.1",
     "jsts": "2.2.0",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -68,6 +68,7 @@
     "imports-loader": "0.8.0",
     "intl-messageformat": "9.11.4",
     "jest": "27.0.6",
+    "jodit-react": "^4.1.2",
```
ZakarFin, line 71: I would assume draft-js and draft-js-export-html should be removed with this?

bundles/framework/announcements/resources/scss/announcementsplugin.ol.scss
```diff
@@ -26,6 +26,7 @@ div.announcements-content {
 }
 
 div.mapplugin.announcements {
+    padding: 10px 10px 0 10px;
```
ZakarFin, line 29: Does the new editor generate HTML that now requires these padding/align changes (Should the banner/popups be changed as well) or is this some unrelated tuning for the plugin?

osqq: No, the announcements plugin is not yet in line with the other plugins in regards of styling, so this would be a temporary quick fix so the plugins above this wont touch.

package.json
osqq: yes, forgot to remove

ZakarFin:
Thanks :) It might take a while before we can test this change, but we are looking into this.

ZakarFin:
npm run build for this fails with:
```
ERROR in ./node_modules/oskari-frontend/bundles/admin/admin-announcements/view/AnnouncementsForm.jsx
Module not found: Error: Can't resolve 'draft-js/dist/Draft.css'
```

Looks like there's a reference to Draft.css in here that needs to be removed as well: https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/admin/admin-announcements/view/AnnouncementsForm.jsx#L12

ZakarFin:
Thanks, I'll merge this to a new branch called wysiwyg for testing purposes before we can decide if this can be merged to develop.

ZakarFin:
Looks good for merging. Thanks!

PR #2681, opened 2024-09-20
Lodash lib remove

okauppinen:
Lodash global `_` expose was removed in: https://github.com/oskariorg/oskari-frontend/pull/2666
As library isn't needed anymore => remove lodash from libraries

When something is imported from lodash (npm dependency), it exposes global `_`.  However we can't rely on it because no common (always existing) bundle imports/exposes global `_`. 

Replaced lodash usage with native functions. When lodash has to be used (mainly outdated jQuery based components use: template, bind,..), imported needed functions from lodash.

Now only MetadataPanel.js templates and sprite.js uses `_` and them import/require lodash.

Lodash 4.x => Removed _.pluck in favor of _.map with iteratee shorthand
`const obj = [{foo: 'bar'}, {foo: 'baz'}];`
`_.pluck(obj, 'foo')` => `_.map(obj, 'foo')`
=> `['bar', 'baz']`

Native Array.map:
`obj.map(item => item.foo)`


ZakarFin:
Nice, it was very fiddly which lodash version was actually used as it could change after some bundle was loaded. Now we know it's the one referenced in package.json

PR #2684, opened 2024-10-07
escape path separator in lazy loader on windows

DenverCoder544:
espace path separator on win


ZakarFin:
Fixes issues from #2659 for Windows environments

PR #2686, opened 2024-10-09
AntD5 MapButton hover & size fixes

okauppinen:
- Fix icon sizes by setting font-size style for icon
- Move svg strike hover effect/style to MapButton (highlight prop). Fixes issue where swipe icons important none fill was overrided  by MapButtons important fill. Also I think that is cleaner to give svg only without styling for MapModuleButton icon prop regardless MapButton gets more complex
- Use iconSize prop instead of font-size, height, width styling
- fix MapModuleTextButton hover by inceasing specificity
- use delete icon instead of DeleteButton for coverage clear

Tried to get rid of all size styling in MapButton and let AntD set sizes by passing font-size for Icon. If icon is given via props to button (not child) AntD sets icon-only, vertical align `em`, padding.. But own svg icons and XY text icon needs font-size. I will try creating TextIcon component and using Icon component={svg} in own PR. And use TextIcon for small minus and plug zoom slider icons as AntD is changed shape-rendering for Plus/MinusOutlined.

AntD4:
antd4

AntD5:
antd5


bundles/mapping/mapmodule/plugin/zoombar/ZoomSlider.jsx
```diff
@@ -141,38 +141,40 @@ export const ZoomSlider = ({ changeZoom, zoom = 0, maxZoom, isMobile = false, ..
...
-                        onChange={value => {
-                            changeZoom(value);
-                        }}
-                        {...rest}
-                    />
-                </ThemeProvider>
-            )}
+
+            <ThemeProvider value={mapModule.getMapTheme()}>
+                <ThemedSlider
```
ZakarFin, line 155: Does this now always render the slider? It was previously only rendered on large screens. Also the +/- buttons were larger on smaller screens.

okauppinen: There is isMobile check before this: https://github.com/oskariorg/oskari-frontend/blob/develop/bundles/mapping/mapmodule/plugin/zoombar/ZoomSlider.jsx#L119

PR #2687, opened 2024-10-09
Metadata flyout react implementation

okauppinen:
New react based implementation for showing parsed and localized metadata in flyout. Added util functions for prettify, linkify, localize codes, languages, format data => rendering is handled mostly with common `LabeledItem`. Especially code localization `{ label, description }` , urls/links and splitting lines for own paragraphs would require extra logic (in every component) if every content type has own component. Also label is rendered only when reasonable content is available.  Decided to hide/handle complexity in util functions and preprocess data for common rendering.

Differences between new and metadataflyout:
- not registered as stateful and extension (tile, flyout)
- format dates
- handle a-tags (metadataflyout replaces urls with a-tag and if content has a-tag already replace won't work properly)
- show InfoIcon if description is available (old has title without any visual identification)
- handles only one identification (couldn't find any metadata which have more than one identification)
- didn't copy: layerSpec, tile, Flyout, Extension, AddTabRequest[Handler], some old localization files 
- ShowMetaDataRequest handles only layerId and uuid. Not RS_Identifier_Code, RS_Identifier_CodeSpace, additionalMetadata

`catalogue.ShowMetadataRequest` works with `uuid` or `layerId`:

Used same `metadataflyout` so requires only main.js:
`import 'oskari-lazy-loader?metadataflyout!oskari-frontend/packages/catalogue/bundle/metadataflyout/bundle.js';`
==>
`import 'oskari-lazy-loader?metadataflyout!oskari-frontend/bundles/catalogue/metadata/bundle.js'`


bundles/catalogue/metadata/instance.js
```diff
@@ -0,0 +1,169 @@
...
+            this.sandbox = sandbox || Oskari.getSandbox(this.conf.sandbox || 'sandbox');
+            this.mapmodule = this.sandbox.findRegisteredModuleInstance('MainMapModule');
+
+            this.sandbox.register(this);
+
+            this.requestHandler = Oskari.clazz.create('Oskari.catalogue.bundle.metadata.request.ShowMetadataRequestHandler', this);
+            this.sandbox.requestHandler('catalogue.ShowMetadataRequest', this.requestHandler);
+
+            Object.getOwnPropertyNames(this.eventHandlers).forEach(p => this.sandbox.registerForEventByName(this, p));
+
```
ZakarFin, line 40: I think removing the stateful from this might be a good idea, but we need to handle state anyway on startup to make sure we can show the requested metadata on startup. There's an URL-param that can be passed to this as state and it's expected to be shown on startup if it receives the uuid on state.

bundles/catalogue/metadata/view/tab/components/ResponsibleParties.jsx
```diff
@@ -0,0 +1,30 @@
...
+    }
+    return (
+        <Fragment>
+            <Label labelKey={dataKey} />
+            <List>
+                {parties.map((party, i) => (
+                    <li key={i}>
+                        {party.organisationName}
+                        <List>
+                            {party.electronicMailAddresses.map((address, i) => <li key={i}>{address}</li>)}
```
ZakarFin, line 18: Do we need to check if party has electronicMailAddresses before mapping?

PR #2692, opened 2024-10-11
Statgrid publisher edit

okauppinen:
Add common statistics tool which shows 'Open data search' button and has common getValues for statistics.

- allow to hide tool checkbox in Publisher2
- use tool localization from statsgrid bundle
- rename group for 'statsgrid' (let data as it is because statsgrid2016 uses it)


bundles/framework/publisher2/handler/ToolPanelHandler.js
```diff
@@ -39,8 +39,7 @@ class UIHandler extends StateHandler {
                     component: toolComponent.component,
                     handler: toolComponent.handler,
                     publisherTool: tool,
-                    id: tool.getTool().id,
-                    title: tool.getTool().title
+                    ...tool.getTool()
```
ZakarFin, line 42: We should just add `tool.hideCheckbox` here instead so it's clear (for maintenance purposes) what is/can be passed.

ZakarFin: Object spread -> explicit set.

PR #2695, opened 2024-10-14
Publisher tools reactification

DenverCoder544:
Scalebar & Logo es class impl


packages/framework/bundle/publisher2/bundle.js
```diff
@@ -95,7 +95,7 @@ Oskari.clazz.define("Oskari.mapframework.bundle.publisher2.PublisherBundle", fun
         },
         {
             "type": "text/javascript",
-            "src": "../../../../bundles/framework/publisher2/tools/ScalebarTool.js"
+            "src": "../../../../bundles/mapping/mapmodule/plugin/scalebar/publisher/ScalebarTool.js"
```
ZakarFin, line 98: This is kinda weird with published picking some parts under mapmodule. It helps with code not being packaged for the client when publisher isn't present though (embedded maps for example). 

How about if we create a `mapmodule/publisher/tools.js` that only imports the tools that mapmodule is providing for the publisher:
```
import '../plugin/scalebar/publisher/ScalebarTool.js';
import '../plugin/logo/publisher/LogoTool.js';
```
Publisher would then have some kind of a documentable "API" how it could in some cases pick tools from other bundles?

ZakarFin: The src here would then be just one line 
```suggestion
            "src": "../../../../bundles/mapping/mapmodule/publisher/tools.js"
```

bundles/mapping/mapmodule/plugin/scalebar/publisher/ScalebarTool.js
```diff
@@ -0,0 +1,58 @@
...
+        super(...args);
+        this.index = 1;
+        this.group = 'additional';
+        this.config = null;
+    }
+
+    getTool () {
+        return {
+            id: SCALEBAR_TOOL_ID,
+            title: Oskari.getMsg('Publisher2', 'BasicView.maptools.ScaleBarPlugin'),
```
ZakarFin, line 15: Maybe we should bring these localizations over to mapmodule if we want to have the tools here?

ZakarFin: With this it's kind of all over the place with localizations being under publisher, tools under plugins and the actual import is still done on publisher. Maybe we should have everything under the bundles that have the code (in mapmodule in this case). Best case scenario would be if we had some sort of an API how publisher would be able to hook into these without bundling everything in, even when the publisher is NOT present on the app.

packages/framework/bundle/publisher2/bundle.js
DenverCoder544: Sounds reasonable. Let's move to this direction. 👍

bundles/mapping/mapmodule/plugin/scalebar/publisher/ScalebarTool.js
DenverCoder544: Moved these under mapmodule localisations in publishertools segment. At least it's more logical now, maybe not more easily automatically found though...

PR #2703, opened 2024-10-21
Publisher tools reactification

DenverCoder544:
MyLocation - tool es/react implementation


bundles/mapping/mapmodule/resources/locale/en.js
```diff
@@ -161,6 +161,18 @@ Oskari.registerLocalization(
         },
         "publisherTools": {
             "ScaleBarPlugin": "Scale bar",
+            "MyLocationPlugin": {
```
ZakarFin, line 164: Should these be removed from under publisher?

PR #2707, opened 2024-10-25
AdminUsers role select fix

okauppinen:
Roles was splitted to roles and systemRoles in previous PR and forget to made changes to user form select => role select options were empty.

Changes:
- Added own RoleSelect component to fix issue.
- Added props validation and some eslint fixes
- Added show users icon for admin role
- Added edit user icons for users by role tab. If fetched users doesn't include selected user, triggers search.
- Fixed IconButton hover background

### DIFF
--- a/bundles/framework/admin-users/view/RoleBlock.jsx
+++ b/bundles/framework/admin-users/view/RoleBlock.jsx
@@ -1,14 +1,23 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { TextInput, Message } from 'oskari-ui';
 import { UserOutlined } from '@ant-design/icons';
 import { Block, Button, ButtonContainer } from './styled';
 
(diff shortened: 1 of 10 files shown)


bundles/framework/admin-users/view/RoleBlock.jsx
```diff
@@ -1,14 +1,23 @@
...
+const ADMIN = 'admin';
+
 export const RoleBlock = ({ role, controller, isSystemRole, editingRole }) => {
-    const { id, name } = role;
+    const { id, name, type } = role;
     if (isSystemRole) {
         return (
             <Block>
                 <span>{name}</span>
+                { type === ADMIN &&
```
ZakarFin, line 15: Only admin users can be shown like this? This might be better with `where type not in (GUEST, LOGGED_IN)` just to get the long listings removed?

PR #2742, opened 2024-11-25
Close message on click

ZakarFin:
Allows user to click a message to close it. Example:

image


ZakarFin:
Not sure if we want to add this for all notifications (actually messages). But it looks like onClick can't be passed by the caller: https://github.com/oskariorg/oskari-frontend/blob/2.13.2/src/react/util/extras/Messaging.js#L43

PR #2759, opened 2024-12-05
Publisher sidebar reactification

DenverCoder544:
in this draft pr:
*The sidebar container is mostly es / react
* the first & second panel are reactified and using our react Collapse - component
* state handling / validation as well as rendering might be a bit quirky and might need some work.

this is what it would look like:
image

image


bundles/framework/publisher2/view/dialog/ReplaceConfirmDialogContent.jsx
```diff
@@ -0,0 +1,27 @@
...
+import PropTypes from 'prop-types';
+import { Button, Message } from 'oskari-ui';
+import { SecondaryButton } from 'oskari-ui/components/buttons';
+import { ButtonContainer, DialogContentContainer } from './Styled';
+import { LocaleProvider } from 'oskari-ui/util';
+
+export const ReplaceConfirmDialogContent = ({ okCallback, closeCallback }) => {
+    return <LocaleProvider value={{ bundleKey: 'Publisher2' }}>
+        <DialogContentContainer>
+            {<Message messageKey='BasicView.confirm.replace.msg'/>}
```
ZakarFin, line 11: Are the {} needed?
```suggestion
            <Message messageKey='BasicView.confirm.replace.msg'/>
```

bundles/framework/publisher2/view/dialog/ReplaceConfirmDialogContent.jsx
```diff
@@ -0,0 +1,27 @@
...
+            </ButtonContainer>
+        </DialogContentContainer>
+    </LocaleProvider>;
+};
+
+ReplaceConfirmDialogContent.propTypes = {
+    message: PropTypes.string,
+    okCallback: PropTypes.func,
+    closeCallback: PropTypes.func,
+    replaceButtonTitle: PropTypes.string
```
ZakarFin, line 26: Only these two?
```suggestion
    okCallback: PropTypes.func,
    closeCallback: PropTypes.func
```

bundles/framework/publisher2/view/PanelGeneralInfo.js
```diff
@@ -1,178 +1,112 @@
...
-     * @static
-     * @param {Object} sandbox
-     * @param {Object} localization
-     *       publisher localization data
-     */
-    function (sandbox, localization) {
-        this.loc = localization;
-        this.sandbox = sandbox;
+export const PANEL_GENERAL_INFO_ID = 'panelGeneralInfo';
+export class PanelGeneralInfo extends React.Component {
```
ZakarFin, line 6: We should try making these function components. Would that mean that the panel is passed some onChange function that is called when the user changes stuff? We can probably move the name/domain validation to some helper functions instead of having them on the component itself.

bundles/framework/publisher2/view/PanelGeneralInfo.js
```diff
@@ -1,178 +1,112 @@
...
-                    errors.push({
-                        field: name,
-                        error: me.loc.error.domainStart
-                    });
-                    return errors;
-                }
-                return errors;
-            };
+    getLabel () {
+        return this.localization?.domain?.title || '';
```
ZakarFin, line 40: I think we only need to have something like getLabel() for the tool-panels. Maybe not even for them as they should know the label to render when they are created?

bundles/framework/publisher2/view/dialog/ReplaceConfirmDialogContent.jsx
DenverCoder544: fixed this

bundles/framework/publisher2/view/dialog/ReplaceConfirmDialogContent.jsx
DenverCoder544: fixed this 2

bundles/framework/publisher2/view/PanelGeneralInfo.js
DenverCoder544: Yeah. The first draft was somewhat quirky. Took a step back and separated the component + state handling in different places so now we have a function component paired with a separate statehandler/controller along with some additional logic to make it all work with publisher sidebar and antd-collapse. So far this applies only to generalinfopanel but I'll do map sizes panel next.

bundles/framework/publisher2/view/PanelGeneralInfo.js
DenverCoder544: Now that a step has been taken towards function components and statehandlers this is now useless.

bundles/framework/publisher2/handler/PanelMapPreviewHandler.js
```diff
@@ -0,0 +1,162 @@
...
+        const selectedMapSize = MAP_SIZES.find(size => size.id === selectedId);
+        if (selectedId === CUSTOM_MAP_SIZE_ID) {
+            selectedMapSize.width = data?.metadata?.size?.width || '';
+            selectedMapSize.height = data?.metadata?.size?.height || '';
+        }
+        this.updateMapSize(selectedMapSize);
+    }
+
+    updateMapSize (mapSize) {
+        console.log('updateMapSize: ', mapSize);
```
ZakarFin, line 59:
```suggestion
```

bundles/framework/publisher2/handler/PanelMapPreviewHandler.js
```diff
@@ -0,0 +1,162 @@
...
+            }
+        };
+
+        if (!isNaN(parseInt(selected.width)) && !isNaN(parseInt(selected.height))) {
+            values.metadata.size = {
+                width: selected.width,
+                height: selected.height
+            };
+        }
+        console.log('values: ', values);
```
ZakarFin, line 127:
```suggestion
```

bundles/framework/publisher2/view/PublisherSideBarHandler.js
```diff
@@ -0,0 +1,177 @@
...
+    updateGeneralInfoPanel () {
+        const newCollapseItems = this.getState().collapseItems.map(item => item);
+        const generalInfoPanel = newCollapseItems.find(item => item.key === PANEL_GENERAL_INFO_ID);
+        generalInfoPanel.children = this.renderGeneralInfoPanel();
+        this.updateState({
+            collapseItems: newCollapseItems
+        });
+    }
+
+    renderGeneralInfoPanel () {
```
ZakarFin, line 39: We should move the render functions to PublisherSidebar.js directly since they are always there. Maybe we need to do something like this with the tool panels though?

bundles/framework/publisher2/view/form/MapPreviewForm.jsx
```diff
@@ -31,7 +29,7 @@ const RowContainer = styled('div')`
 
 const Label = (props) => {
     const { option } = props;
-    let text = Oskari.getMsg(MESSAGE_BUNDLE_KEY, 'BasicView.sizes.' + option.id);
+    let text = Oskari.getMsg(PUBLISHER_BUNDLE_ID, 'BasicView.sizes.' + option.id);
     if (option.id !== CUSTOM_MAP_SIZE_ID && !isNaN(parseInt(option.width)) && !isNaN(parseInt(option.height))) {
         text += ' (' + option.width + ' x ' + option.height + 'px)';
     }
```
ZakarFin, line 35: Something like this might work so we don't have to keep passing the bundle id and use Oskari.getMsg() directly:
```suggestion
    let extraInfo = '';
    if (option.id !== CUSTOM_MAP_SIZE_ID && !isNaN(parseInt(option.width)) && !isNaN(parseInt(option.height))) {
        extraInfo = '(' + option.width + ' x ' + option.height + 'px)';
    }
    return <Message key={ 'BasicView.sizes.' + option.id}>{ extraInfo }</Message>
```

bundles/framework/publisher2/view/PublisherSideBarHandler.js
ZakarFin: Lets return to this once it's more ready. Probably easier to test things when jQuery panels are no longer in the way.

ZakarFin:
Lets go with this as baseline and try tuning it when we get all the panels as React impls.

PR #2760, opened 2024-12-10
Fix postinstall script for non-devmode installs

ZakarFin:
Running `npm run build` with defaults on for example sample-application doesn't work correctly without this. Only works when oskari-frontend is installed as "dev mode" from a folder next to an app with `npm install ../oskari-frontend`. The change forces install to throw an error if the postinstall script can't find Cesium and do its thing. This makes installing oskari-frontend  on an app throw an error instead of finding out afterwards when running `npm run build` for the application.

Apparently this can be added for an app to make npm run build work:
https://www.npmjs.com/package/@open-wc/webpack-import-meta-loader

### DIFF
--- a/webpack/hacks/postinstall.js
+++ b/webpack/hacks/postinstall.js
@@ -11,6 +11,7 @@ if (!fs.existsSync(pathToCesium)) {
 // try again with non-dev-mode
 if (!fs.existsSync(pathToCesium)) {
     // something went terribly wrong
+    throw new Error('Couldn\'t find @cesium');
 } else {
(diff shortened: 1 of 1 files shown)


ZakarFin:
Maybe #2761 instead

ZakarFin:
Merged #2761 as better solution. Eventhough the added dependency looks "troubling" as 0.4.7 from 5 years ago, it's pretty simple and removes the need to check if the Cesium file that the postinstall hack overwrote has changed when updating the library.

PR #2761, opened 2024-12-11
Add webpack-import-meta-loader

tuomari:
Add import.meta babel support to webpack4, which does not support it natively
import.meta is used for example in new cesium versions, and when the loader is added, the hacks are no longer needed.


ZakarFin:
Thanks! The 0.4.7 version from 5 years ago doesn't really sound that good (https://www.npmjs.com/package/@open-wc/webpack-import-meta-loader), but I'll think about this a bit. We might need to do a hotfix as we've noticed the vectortile decluttering isn't working properly and adding new layers with url-templates like {x]/{y}/{z} doesn't seem to work properly.

webpack/config.js
```diff
@@ -122,6 +122,10 @@ const BABEL_LOADER_RULE = {
 const getModuleRules = (isProd = false, antThemeFile) => {
     const styleFileRules = getStyleFileRules(isProd, antThemeFile);
     const rules = [
+        {
+            test: /\.js$/,
+            loader: require.resolve('@open-wc/webpack-import-meta-loader')
+        },
```
ZakarFin, line 128:
```suggestion
/*
webpack-import-meta-loader fixes this error for building Cesium:
ERROR in ../oskari-frontend/node_modules/@cesium/engine/Source/Core/buildModuleUrl.js 42:43
Module parse failed: Unexpected token (42:43)
*/
        {
            test: /\.js$/,
            loader: require.resolve('@open-wc/webpack-import-meta-loader')
        },
```

PR #2772, opened 2025-01-14
Remove old metadatacatalogue, replaced by metadatasearch

ZakarFin:
This would remove `AddSearchResultActionRequest` that is not implemented on `metadatasearch`


ZakarFin:
Related to: https://github.com/oskariorg/oskari-documentation/issues/73

PR #2773, opened 2025-01-14
Remove metadataflyout, replaced by metadata bundle

ZakarFin:
`catalogue/metadataflyout` replaced by `catalogue/metadata`.

Also removes unused `catalogue.AddTabRequest` that was not implemented on the new metadata bundle.


ZakarFin:
Related to: https://github.com/oskariorg/oskari-documentation/issues/73

PR #2775, opened 2025-01-17
Metadata stateful

okauppinen:
changes for stateful


ZakarFin:
Cherry-picked to 2.14.2

PR #2776, opened 2025-01-22
LayerList sub group/collapse tools fix

okauppinen:
Previously sub group/collapse used same LayerCollapsePanel than normal groups:
https://github.com/oskariorg/oskari-frontend/blob/2.13.2/bundles/framework/layerlist/view/LayerViewTabs/LayerList/LayerCollapse/LayerCollapsePanel.jsx#L123

When rendering was splitted to own files SubGroupList didn't have collapse extra (tools). Moved getCollapseItems to own helper file to get same content for sub groups. Renamed SubGroupCollapse because it's Collapse not List.


ZakarFin:
Solves https://github.com/oskariorg/oskari-documentation/issues/74

PR #2783, opened 2025-02-03
add collapse tooltips

DenverCoder544:
these here
image


ZakarFin:
Maybe use this instead https://github.com/oskariorg/oskari-frontend/blob/master/src/react/components/icons/Info.jsx for the i-icons (looks like it's the quetion mark icon, but it's better to have that as SVG instead of the old png with "i"?

PR #2784, opened 2025-02-04
Statsgrid panel react collapse

DenverCoder544:
image

### DIFF
--- a/bundles/framework/publisher2/view/PublisherSidebar.js
+++ b/bundles/framework/publisher2/view/PublisherSidebar.js
@@ -43,7 +43,7 @@ class PublisherSidebar {
         this.handler.init(this.data, this.publisherTools);
         const content = <LocaleProvider value={{ bundleKey: 'Publisher2' }}>
             <ThemeProvider>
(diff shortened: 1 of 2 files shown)


bundles/framework/publisher2/view/PublisherSidebar.js
```diff
@@ -66,49 +65,8 @@ class PublisherSidebar {
...
-                panel.addClass('t_' + group);
-                accordion.addPanel(panel);
-            }
-        });
-
-        // -- render to UI and setup buttons --
-        const contentDiv = container.find('div#jqueryAccordions');
-        accordion.insertTo(contentDiv);
         // disable keyboard map moving whenever a text-input is focused element
+        const contentDiv = container.find('div#publisherWrapperDiv');
```
ZakarFin, line 69: Looks like the only reason to have a reference is to disable keyboard navigation of map when typing on input-fields. Let's try removing the lines instead and see if it's still needed or if we still need to prevent event bubbling from the React-based inputs.

PR #2786, opened 2025-02-11
toggle statsgrid panel when layer is toggled

DenverCoder544:
toggle statsgrid-panel on/off in publisher when statsgrid layer is added / removed.

### DIFF
--- a/bundles/framework/publisher2/handler/StatsGridPanelHandler.js
+++ b/bundles/framework/publisher2/handler/StatsGridPanelHandler.js
@@ -1,0 +1,69 @@
+import { controllerMixin } from 'oskari-ui/util';
+import { ToolPanelHandler } from './ToolPanelHandler';
+const STATSGRID_LAYER_ID = 'STATS_LAYER';
+class UIHandler extends ToolPanelHandler {
+    constructor (tools, sandbox, toggleStatsGridPanel) {
+        super(tools);
+        this.sandbox = sandbox;
+        this.toggleStatsGridPanel = toggleStatsGridPanel;
+        this.eventHandlers = {
+            /**
+             * @method AfterMapLayerAddEvent
+             * @param {Oskari.mapframework.event.common.AfterMapLayerAddEvent} event
+             */
+            AfterMapLayerAddEvent: function (event) {
+                if (!!this.toggleStatsGridPanel && event.getMapLayer().getId() === STATSGRID_LAYER_ID) {
+                    this.toggleStatsGridPanel(true);
+                }
+            },
+
+            /**
+             * @method AfterMapLayerRemoveEvent
+             * @param {Oskari.mapframework.event.common.AfterMapLayerRemoveEvent} event
+             */
+            AfterMapLayerRemoveEvent: function (event) {
+                if (!!toggleStatsGridPanel && event.getMapLayer().getId() === STATSGRID_LAYER_ID) {
+                    toggleStatsGridPanel(false);
+                }
+            }
+
+        };
+    }
+
+    initTools () {
+        const hasTools = super.init(this.data, true);
+        return hasTools;
+    }
+
+    init (data) {
+        this.data = data;
+        Object.getOwnPropertyNames(this.eventHandlers).forEach((event) => {
+            this.sandbox.registerForEventByName(this, event);
+        });
+
+        return this.initTools();
+    }
+
+    stop () {
+        super.stop();
+        Object.getOwnPropertyNames(this.eventHandlers).forEach(event => this.sandbox.unregisterFromEventByName(this, event));
+    }
+
+    onEvent (event) {
+        const handler = this.eventHandlers[event.getName()];
+        if (!handler) {
+            return;
+        }
+        return handler.apply(this, [event]);
+    }
+
+    getName () {
+        return 'Oskari.mapframework.bundle.publisher2.view.PanelMapLayers';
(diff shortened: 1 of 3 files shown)


bundles/framework/publisher2/view/PublisherSideBarHandler.js
```diff
@@ -289,9 +289,40 @@ class PublisherSidebarUIHandler extends StateHandler {
...
+            key: PANEL_STATSGRID_ID,
+            label: Oskari.getMsg('Publisher2', 'BasicView.statsgrid.label'),
+            children: this.renderStatsGridPanel()
+        };
+    }
+
+    toggleStatsGridPanel (visible) {
+        const newCollapseItems = this.getState().collapseItems
+            .filter(item => item.key !== PANEL_STATSGRID_ID)
+            .map(item => item);
```
ZakarFin, line 303: The map() can be removed?

bundles/framework/publisher2/view/PublisherSideBarHandler.js
```diff
@@ -289,9 +289,40 @@ class PublisherSidebarUIHandler extends StateHandler {
...
+            }
+        }
+
+        this.updateState({
+            collapseItems: newCollapseItems
+        });
+    }
+
     updateStatsgridPanel () {
         const newCollapseItems = this.getState().collapseItems.map(item => item);
```
ZakarFin, line 320: It's basically this?
```suggestion
        const newCollapseItems = [...this.getState().collapseItems];
```

bundles/framework/publisher2/view/PublisherSideBarHandler.js
```diff
@@ -289,9 +289,40 @@ class PublisherSidebarUIHandler extends StateHandler {
...
         const panel = newCollapseItems.find(item => item.key === PANEL_STATSGRID_ID);
+
+        if (!panel) {
+            return;
+        }
         panel.children = this.renderStatsGridPanel();
         this.updateState({
             collapseItems: newCollapseItems
```
ZakarFin, line 328: Only thing that get's changed is the children of the statsgrid collapseItem, but this forces re-rendering?

bundles/framework/publisher2/handler/ToolPanelHandler.js
```diff
@@ -12,26 +12,30 @@ class UIHandler extends StateHandler {
         }
     }
 
-    init (data) {
+    init (data, silent) {
         this.data = data;
-        while (this.toolsToInit.length) {
-            const tool = this.toolsToInit.shift();
```
ZakarFin, line 18: The shift is used to move tools from "toolsToInit" to state.tools. If we don't shift() we have the list in both variables. Is this intended?

ZakarFin: Maybe it is to enable initializing them after the publisher was opened so we have the full set of tools that could be shown and then on state the tools that are actually shown. Maybe rename `toolsToInit` to something like `allAvailableTools`?

bundles/framework/publisher2/handler/StatsGridPanelHandler.js
```diff
@@ -0,0 +1,69 @@
...
+            },
+
+            /**
+             * @method AfterMapLayerRemoveEvent
+             * @param {Oskari.mapframework.event.common.AfterMapLayerRemoveEvent} event
+             */
+            AfterMapLayerRemoveEvent: function (event) {
+                if (!!toggleStatsGridPanel && event.getMapLayer().getId() === STATSGRID_LAYER_ID) {
+                    toggleStatsGridPanel(false);
+                }
```
ZakarFin, line 27: Make `AfterMapLayerAddEvent` and `AfterMapLayerRemoveEvent` consistent with how they reference `toggleStatsGridPanel()`. If we don't need to save a reference to it with `this.toggleStatsGridPanel` we shouldn't.

bundles/framework/publisher2/handler/ToolPanelHandler.js
DenverCoder544: The case where we might have to (re)-initialise the panel later hadn't been thought of before. But this is in fact the case when we toggle panels dynamically. Renamed the variable for this to be a little more intuitive.

bundles/framework/publisher2/view/PublisherSideBarHandler.js
DenverCoder544: yes!

bundles/framework/publisher2/view/PublisherSideBarHandler.js
DenverCoder544: had been using the map() - syntax throughout the file to achieve the same effect. Changed all of those to use spread.

bundles/framework/publisher2/view/PublisherSideBarHandler.js
DenverCoder544: This mechanics is due to the way we have to feed individual panels to antd collapse but basically we gotta (re)render the items in order to reflect the changes to state in ui. Not pretty, agreed.

bundles/framework/publisher2/handler/StatsGridPanelHandler.js
DenverCoder544: yeah, you're right. Removed the reference.

PR #2787, opened 2025-02-12
try to determine checked for maplegend even when there are async layers

DenverCoder544:
seems to work for a simple async case where nothing surprising happens.

I'm guessing that if we open a published map for editing, that originally has
* map legend - checkbox checked 
* some layer in the original config that had a maplegend but which doesn't exist anymore and loading fails
* we later add a map layer that has a legend 

=> checkbox might be turned on by surprise but maybe this is something of a corner case...


ZakarFin:
Tries to solve this: https://github.com/oskariorg/oskari-documentation/issues/81

PR #2791, opened 2025-02-18
Oskari 3.0 bundle loader

ZakarFin:
The idea is that we can get rid of packages folder and simplify the `bundle.js` files that were located in the packages-folder (which makes things easier to document etc) .

The included `bundles/admin/admin/index.js` is an example of what the bundle.js replacement could be. In the applications main.js this would change the import from:
```
import 'oskari-loader!oskari-frontend/packages/admin/bundle/admin/bundle.js';
```
to:
```
import 'oskari-bundle!oskari-frontend/bundles/admin/admin';
```
This makes everything simpler by removing lot of the copy/pasted parts of bundle.js and only leaving the essentials:
- what happens when the application wants to use a bundle with given id. This is done with:
```
Oskari.bundle('myBundle', () => /* return a bundle instance for bundle "myBundle" that can be started */ );
```

Everything else remains the same (the instance start() function is called and configuration is injected). The `bundle-loader` figures out the localization files for the bundle by searching files in the `resources/locale/` folder next to the index.js. This way we only need to add the files and not reference them in the index.js. If we do need to manually setup locales, we could add something like locales.json next to index.js that can refer to files that are not in the default localization folder.


ZakarFin:
Note! The new loader doesn't support the expose-functionality that is used on some bundles.

PR #2792, opened 2025-02-18
Oskari 3.0 bundle lazy-loader

ZakarFin:
Continues: #2791 and adds `oskari-lazy-bundle` loader that works with the new `oskari-bundle` loader.

Now we have both loaders with new versions for application main.js usage:
- `oskari-bundle` replaces oskari-loader and supports more streamlined bundle registrations
- `oskari-lazy-bundle` replaces oskari-lazy-loader for adding support to lazy-load bundles with the streamlined bundle registration
 
Moved some internal'ish functions around the core (from releasenotes):

These changes shouldn't really affect your app, but things that have changed inside the "engine":

- New core component `src/BundleRegister` for managing bundles and exposes Oskari.bundle() (previously part of src/loader.js) and Oskari.lazyBundle() functions.
- Oskari.bundle_manager functions removed:
    - registerDynamic() - replaced by BundleRegister.lazyBundle() that is exposed as Oskari.lazyBundle()
    - loadDynamic() - moved to src/loader.js as private function as it doesn't need to be exposed

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -8,6 +8,33 @@ Removed bundles:
 - `catalogue/metadatacatalogue` replaced by `catalogue/metadatasearch`
 - `catalogue/metadataflyout` replaced by `catalogue/metadata`
 
+### Changes for bundle registrations 
+
+In preparation of removing the packages-folder from oskari-frontend and migrating any bundle.js files under it to bundles-folder as index.js files.
+
+import 'oskari-loader!oskari-frontend/packages/admin/bundle/admin/bundle.js';
+
+New loaders for application main.js usage:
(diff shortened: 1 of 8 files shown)


ReleaseNotes.md
```diff
@@ -8,6 +8,33 @@ Removed bundles:
 - `catalogue/metadatacatalogue` replaced by `catalogue/metadatasearch`
 - `catalogue/metadataflyout` replaced by `catalogue/metadata`
 
+### Changes for bundle registrations 
+
+In preparation of removing the packages-folder from oskari-frontend and migrating any bundle.js files under it to bundles-folder as index.js files.
+
+import 'oskari-loader!oskari-frontend/packages/admin/bundle/admin/bundle.js';
```
ZakarFin, line 15:
```suggestion
```

PR #2798, opened 2025-02-26
Packages/bundle.js migration (to bundles/index.js)

ZakarFin:
Includes the commit from #2800 to fix an issue with removing the findbycoordinates additional styles file as unnecessary.


ZakarFin:
This includes pretty much everything else but excludes:
- rpc (requires some change on how JSChannel is imported(
- bundles under mapping (these might need some tweaking on how to expose the 2d ol vs 3d cesium map engine modules)
- jQuery UI-components (divmanazer, ui-components, oskariui/these share localizations so they need some tweaking)

PR #2799, opened 2025-02-26
Remove admin-publish-transfer as app specific bundle

ZakarFin:
The tool requires React-migration and is an app specific (very fiddly to use) way of moving embedded maps appsetups between Oskari instance environments. The bundle was moved to the app that actually uses this and removed from oskari-frontend as it's very hard to use correctly (admin-only functionality).

Because of timing pressure we decided to leave the jsondiff dependency in the libraries folder for now, but added a OskariReadme.md to warn people against building new stuff relying on it.

---
Separated from #2798 (includes its commits) as the transfer tool requires react-migration as well as bundle.js migration.


ZakarFin:
The transfer-tool will most likely be removed from oskari-frontend. It's an application specific thing we have added that is really easily misused to break things and hard to document. The plan is to move it to the app repository that uses it.

PR #2806, opened 2025-03-03
Remove selectlist & chosen.js

ZakarFin:
Remove SelectList component and chosen.js as it was the component to use it.

Continues #2805. 

Related to https://github.com/oskariorg/oskari-documentation/issues/84


ZakarFin:
Looks like this is still used on pti-frontend. We might need to restore it or move the code to https://github.com/nls-oskari/pti-frontend

PR #2807, opened 2025-03-04
Add an RPC function for getting layer groups

ismogis:
Added a new RPC function to retrieve the groups of layers in the embedded map. The function returns groups in a hierarchical and sorted structure based on the hierarchy and `orderNumber` defined in Oskari. Each group includes its ID and name, along with the IDs of layers within the group and its subgroups.


ismogis:
Removed the trailing whitespaces and rebased the branch.

ZakarFin:
Nice implementation 👍

PR #2811, opened 2025-03-04
Layer remove getInspireName

okauppinen:
Layer can have more than one group so getInspireName doesn't work reliably. Remove getInspireName and use getGroups.

Using getGroups for search index fixes issue where search by group doesn't find all matching layers
=> search: 'roads' and layer's groups: ['some group', 'roads']  => should found

### DIFF
--- a/bundles/framework/layerlist/model/LayerGroup.js
+++ b/bundles/framework/layerlist/model/LayerGroup.js
@@ -144,7 +144,7 @@ export class LayerGroup {
     }
 
     clone () {
-        const clone = new LayerGroup(this.id, this.type, this.name, this.description, this.parentId);
+        const clone = new LayerGroup(this.id, this.groupMethod, this.name, this.description, this.parentId);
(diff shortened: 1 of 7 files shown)


bundles/framework/layerlist/model/LayerGroup.js
```diff
@@ -124,8 +142,9 @@ export class LayerGroup {
         }
         return terms.every(key => searchableIndex.indexOf(key.toLowerCase()) !== -1);
     }
+
     clone () {
-        const clone = new LayerGroup(this.id, this.groupMethod, this.name, this.description, this.parentId);
+        const clone = new LayerGroup(this.id, this.type, this.name, this.description, this.parentId);
```
ZakarFin, line 147: Was the second parameter changed by accident or do we have the type in both groupMethod and type variables?

okauppinen: accident. fixed

PR #2815, opened 2025-03-07
Add layer descriptions to RPC function retrieving layers

ismogis:
Added the layer descriptions to the function retrieving layers in the embedded map.


ZakarFin:
Nice, thanks!

PR #2824, opened 2025-04-07
fix the no UI flag for maprotator, getinfo & announcements

DenverCoder544:
fix no ui flag


ZakarFin:
Commit was cherry-picked to hotfix/3.0.1 branch

PR #2826, opened 2025-04-11
Publisher popups react

okauppinen:
- Refactor publisher popups to use react showPopup()
- MyData bundle had already react popup for iframe snippet => moved it to publisher
- changed to use integer height and width (MapPublishedEvent)
- some localization => loc() and eslint updates

Not sure what is the right place for SnippetPopup


ZakarFin:
SnippetPopup is ok under publisher, what we should do is move the mydata published maps tab implementation to publisher as well. But that's another issue.

PR #2834, opened 2025-04-15
Fix publisher adding multiples of the same announcement id to the conf

osqq:


ZakarFin:
Cherry-picked d4cda922832a3efa2a855cb000de187a7a30ec4c to hotfix/3.0.1

PR #2835, opened 2025-04-15
Add styling for printing (with browser functionality)

ZakarFin:
Fixes an issue where browser printing looks like this:
image


ZakarFin:
After, looks like this:
image

PR #2842, opened 2025-04-17
Coordinatetool rpconly fix

DenverCoder544:
fixed a bug where xy-tool never was toggled visible again in publisher when the map was initially saved with "rpc only" flag set.


ZakarFin:
Cherry-picked 7d274dfa0a35737683b71a2356d71d2e97549767 and 3891493d1b792de4223ce9c0c4829f331de7e510 to hotfix/3.0.1

PR #2847, opened 2025-04-23
Publisher side panel and handlers refactoring

okauppinen:
Refactor to use react `showSidePanel` to render publisher panel. Move functionalities from sidebar to handler and service. Service handles transition between geoportal and publisher (layers, plugins, fetch app setup) and sidebar handler handles app setup to be published. Added `setPublishModeImpl` (sidebar, service) which should do all tricks for transition. Service might be registered to sandbox if panels (or tools) are wanted to be added without using protocol.

Refactor `PublisherSidebarHandler` to iterate panels and handlers. Add `getPanelContent` for every handler.  Refactor to use controller from react (some components used handler functions instead of controller).

minor refactoring:
- simplify request handler (remove callback factory and use scheduled function only on startup)
- remove jQuery usage
- remove ColourSchemeChangedEvent
- update loc keys to match (tool) groups
- move sidebar handler and layout form to correct place
- remove internal state and effect from layout panel
- remove unused localization
- remove some controller functions (e.g. stop, getValues, ..) which should never be called from react components
- add props validation
- use tool as is (don't map to custom object)
- use localization key in validation error (localize on render/one place)
- move FONTS to theme constants

bug fixes:
- domain validation
- style presets don't override theme icon hover color with yellow
- fix ToolEnabledChangedEvent listener (plugin/tool added/enabled while editing layout => drag handler added) 

changes:
- show validation error before confirm on edit
- show only one primary button on edit (save as new is normal button now)


ZakarFin:
Looking good, but impossible to say if everything is ok with such a large PR :) Lets merge and test it out.

PR #2763, opened 2024-12-13
Update OpenLayers 9.2.4 -> 10.3.1

ZakarFin:
Updated:
- OpenLayers 9.2.4 -> 10.3.1
- olcs 2.20 -> 2.22.1

PR #2849, opened 2025-04-24
Ol 10 upgrade

DenverCoder544:
Based on this pr https://github.com/oskariorg/oskari-frontend/pull/2763

Adds latest OpenLayers & Cesium + fixes jest config for unit tests


ZakarFin:
Note, affected by this https://github.com/openlayers/openlayers/issues/16878. Changing the opacity on dev.oskari.org for the default layer (vectortile) results in the error about globalAlpha.

PR #2852, opened 2025-04-24
Publisher panel protocol

okauppinen:
Add `AbstractPublisherPanel` and get extra publisher panels using protocol. Before this extra panels were made for extra tool groups using localization (label + tooltip). After this extra panel doesn't require localization to be inside publisher and custom UI handler can be provided.

Move `StatsgridHandler` under statsgrid bundle.

For now statsgrid `open data search` button is made with extra tool (hide check box + button component + handler). Could use handler `getPanelContent` => ToolList + Button and removing tool + handler. I think that hide checkbox was added only for Statsgrid


ZakarFin:
We need to document the adding of panels when the refactoring is completed

PR #2853, opened 2025-04-30
Publisher handlers remove react

okauppinen:
- Remove react from handlers. `getPanelContent` => `getPanelComponent` and move shoModal to popups (confirm, error).
- Store uuid to service
- toggle panel by `visible` state (previously return null if tools empty)
- add all intialized tools to state and render visible tools only


ZakarFin:
The publisher is starting to look really nice.

PR #2856, opened 2025-06-27
Fixing coordinates tool on guided tour

AnnikaEloranta:
This fixes the bug on the last page of the guided tour. Form more info check this ticket: https://jira.nls.fi/browse/OS-3700

The coordinates tool should now open during the guided tour. In addition, the user should now be able to hide/show the coordinates tool by pressing the button on guided tour. The button should have an appropriate label in Finnish, English and Swedish.


ZakarFin:
Good job 👍

PR #2857, opened 2025-07-28
Bug/OS-3678 larger close button on mobile

AnnikaEloranta:
The flyout tool's close button is now more suitable for mobile use: The button looks the same but the click handler is bound to a larger div around the icon, making it easier to click on mobile. 

Fixes this bug: https://jira.nls.fi/browse/OS-3678

### DIFF
--- a/bundles/framework/divmanazer/extension/ExtraFlyout.js
+++ b/bundles/framework/divmanazer/extension/ExtraFlyout.js
@@ -100,9 +100,7 @@ Oskari.clazz.define('Oskari.userinterface.extension.ExtraFlyout',
                     me.options.container.append(popup);
                 }
 
-                var closeButton = popup.find('.oskari-flyouttool-close');
-                closeButton.empty().append(popup.find('.icon-close'));
-                closeButton.on('click', function () {
+                popup.find('.oskari-flyouttool-close').on('click', function () {
                     me.hide();
                 });
 


resources/icons.css
```diff
@@ -219,12 +219,14 @@
     height: 16px;
     background-image: url('icons.png') !important;
     background-repeat: no-repeat !important;
-    background-position: -554px 0px !important;
+    background-position: -552px 2px !important;
```
ZakarFin, line 222: The icons.css is a generated file (based on image-files in the resources/icons folder) so lets try finding another way to fix this instead.

bundles/framework/divmanazer/extension/ExtraFlyout.js
```diff
@@ -97,7 +97,7 @@ Oskari.clazz.define('Oskari.userinterface.extension.ExtraFlyout',
                 } else {
                     me.options.container.append(popup);
                 }
-                popup.find('.icon-close').on('click', function () {
+                popup.find('.oskari-flyouttool-close').on('click', function () {
```
ZakarFin, line 100: Could we have the `.icon-close` element inside the `.oskari-flyouttool-close` element so the bigger element has the click handler and the smaller one is used to clip the icon from the sprite without changes to the icons.css file?

bundles/framework/divmanazer/extension/ExtraFlyout.js
```diff
@@ -97,9 +99,13 @@ Oskari.clazz.define('Oskari.userinterface.extension.ExtraFlyout',
                 } else {
                     me.options.container.append(popup);
                 }
-                popup.find('.icon-close').on('click', function () {
+
+                var closeButton = popup.find('.oskari-flyouttool-close');
+                closeButton.empty().append(popup.find('.icon-close'));
```
ZakarFin, line 104: This row looks like it doesn't really do anything. It removes the icon div and appends it back? Is it needed for some reason or could it be removed? Maybe add a comment why it's there if it's needed? Otherwise it's will be removed by accident as it looks like it's dead code.

PR #2873, opened 2025-09-24
copy myplacesimport to new bundle and rename stuff and call new actio…

DenverCoder544:
…n route

this is basically just a copy of myplacesimport bundle and everything in it. Filenamse / variablenames have been changed to nnnMyFeaturesnnn and the action rout used to import the data uses the new action route but apart from those nothing much changed compared to myplacesimport.


bundles/framework/myfeatures/request/ShowMyFeaturesDialogRequest.js
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * @method create called automatically on construction
+     * @static
+     * @param {Object} id User layer id
+     */
+    function (id) {
+        this._id = id;
+    }, {
+        /** @static @property __name request name */
+        __name: 'MyFeatures.ShowMyFeatureDialogRequest',
```
ZakarFin, line 18: Naming is always hard, but lets at least remove the duplicated reference to `myfeatures`. Not sure if we should try:
- Getting rid of the "bundle reference" before the dot
- Making it match the bundle id
- ???
If we do keep it, we could have the name after the prefix more generic without fear of conflicting naming. But making a "similar but written differently" version of the bundle id is always a bit weird and makes documenting things harder.

```suggestion
        __name: 'myfeatures.ShowLayerDetailsRequest',
```

Also, not sure if this is a request to open layer details or not, but the name should give some indication what we are requesting.

DenverCoder544: Went with prefix + dot + ShowLayerDialogRequest[Handler] 'cos essentially it opens up a dialog. 🤷

PR #2925, opened 2025-11-17
Antd select deprecation fix

DenverCoder544:
replace antd deprecated <Option ..../> with options={...} in all of our `<Select/>`s

image


ZakarFin:
We'll need to add something about this on the releasenotes like: Due to AntD deprecations we have removed the `<Option>` component (used with `<Select>`) from `oskari-ui`. Instead of using:
```javascript
import { Select, Option } from 'oskari-ui';
...
return <Select><Option></Select>
```
Use this instead:
```javascript
import { Select} from 'oskari-ui';
...
const options = [{
   'value': 'actual value',
   'label': 'label to show for users,
   'data-value': 'actual-value'
}];
return <Select options={ options } />
```
The `data-value` is used to make testing easier, but writing this made me think that maybe we should bake it in the Select-component itself (adding data-value for options if it's not set)?

DenverCoder544:
Moved the data-value to be set to actual value inside select component if no explicit other value is given. Definitely better this way.

ZakarFin:
This is ready for merging, but I'll wait a bit so we don't start using these accidentally in apps that are meant to use Oskari 3.2.0 (mainly PTI)

PR #2931, opened 2025-12-09
Myfeatures content editor

DenverCoder544:
first draft of myfeatures bundle playing with a modified copy of content editor. Still has it's glitches but the basic functionality for adding / modifying / deleting features on a myfeatures layer are there.

Some minor changes needed for the backend as well:
https://github.com/oskariorg/oskari-server/pull/1227

image

image

image


bundles/framework/myfeatures/handler/MyFeaturesHandler.js
```diff
@@ -212,6 +237,95 @@ class MyFeaturesHandler extends StateHandler {
         }
     }
 
+    // strip myf_ from the beginning of layerId for db
+    getLayerUUID(layer) {
+        if (!layer?.id?.length > 4) {
+            return null;
+        }
+
+        return layer?.id?.substring(4, layer.id.length) || null;
```
ZakarFin: On the layer operations the server-side code recognizes this and the full layer id with `myf_`-prefix can be used. I would do that here as well so the frontend code doesn't have to modify the layer id that server uses.

bundles/framework/myfeatures/handler/MyFeaturesHandler.js
```diff
@@ -212,6 +237,95 @@ class MyFeaturesHandler extends StateHandler {
...
+        // keep prefix -> use in app.
+        const layerId = layer?.id || null;
+        const layerUUID = this.getLayerUUID(layer);
+        if (!layerUUID) {
+            return;
+        }
+        const url = Oskari.urls.getRoute('MyFeaturesFeature', {
+            layerId: layerUUID,
+            id: featureId,
+            crs: this.sandbox.getMap().getSrsName()
```
ZakarFin: I don't think delete should care about crs?

ZakarFin: But maybe it's mandatory on the `MyFeaturesFeature` route?

src/react/components/FeatureEditor/resources/scss/style.scss
```diff
@@ -0,0 +1,17 @@
+#contentMap {
```
ZakarFin: Lets move these to styled-components definitions.

src/react/components/FeatureEditor/view/SideContentEditor.js
```diff
@@ -0,0 +1,240 @@
+import React from 'react';
```
ZakarFin: This SideContentEditor is probably not used now on `myfeatures` bundle? It should either be deleted or migrated to not use Oskari.clazz.define() and maybe use the showSidePanel() in oskari-ui: https://github.com/oskariorg/oskari-frontend/blob/3.2.1/src/react/components/window/index.js#L322

bundles/framework/myfeatures/handler/MyFeaturesHandler.js
DenverCoder544: Moved this logic to backend. It's definitely better this way. See https://github.com/oskariorg/oskari-server/pull/1227/commits/f62e0113b54ec0bf084b086c74a8732255ce6863

bundles/framework/myfeatures/handler/MyFeaturesHandler.js
DenverCoder544: It's not needed, this was a copypaste thing.

src/react/components/FeatureEditor/resources/scss/style.scss
DenverCoder544: I removed these altogether. They were not being used any more.

src/react/components/FeatureEditor/view/SideContentEditor.js
DenverCoder544: Had it around for reference but the logic is now inside FeatureEditorPanelHandler so this wasn't needed any more.

PR #2933, opened 2026-01-05
Addfeaturetool

DenverCoder544:
*add a "feature tools" - category for layers
*add feature editor feature tool for myfeatures layer
* add a button in infobox for feature's whose layer supports feature editor
* open feature editor with the feature's data

image

image


bundles/framework/myfeatures/request/ShowFeatureEditorRequest.js
```diff
@@ -0,0 +1,47 @@
+/**
```
ZakarFin, line 1: The filename should have the .js suffix :)

bundles/framework/myfeatures/service/MyFeaturesService.js
```diff
@@ -74,6 +75,12 @@ export class MyFeaturesService {
         editLayerTool.setTitle(toolName);
         editLayerTool.setCallback(() => this.sandbox.postRequestByName('myfeatures.ShowLayerDialogRequest', [mapLayer.getId()]));
         mapLayer.addTool(editLayerTool);
+
+
+        const featureEditorTool =  Oskari.clazz.create('Oskari.mapframework.domain.Tool');
+        featureEditorTool.setName(FEATURE_EDITOR_TOOLNAME);
```
ZakarFin, line 81: We should provide title and callback here as we know what should happen when the user clicks the tool (and what the tool should be named for the end-user).

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -519,11 +520,14 @@ Oskari.clazz.define(
...
 
-            if (fragments?.length) {
-                contentData.html = this._renderFragments(fragments);
-                contentData.layerId = fragments[0].layerId;
+            fragments?.forEach((fragment) => {
+                const layerId = fragment.layerId;
+                contentData.html = this._renderFragments([fragment]);
+                contentData.layerId = layerId;
+                contentData.featureId = data?.featureId || null;
+                this.addFeatureEditorTool(contentData);
```
ZakarFin: Lets make this more generic with something like `addFeatureTools(contentData)`.

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -541,6 +545,25 @@ Oskari.clazz.define(
...
+                    // TODO: name is used as the key in a hash and as the label of the button? Ungreat.
+                    name: Oskari.getMsg('myfeatures', 'featureEditor.title'),
+                    action: () => {
+                        const { layerId, featureId } = contentData;
+                        const sandbox = this.getSandbox();
+                        sandbox.postRequestByName('ShowFeatureEditorRequest', [layerId, featureId]);
+                    }
+                }];
+            }
+        },
```
ZakarFin: We don't really need nor want to know what the tool does here or how it's labeled:
```suggestion
        addFeatureTools: function(contentData) {
            const layer = this.getSandbox().findMapLayerFromAllAvailable(contentData?.layerId);
            const { layerId, featureId } = contentData;
             contentData.actions = layer.getFeatureTools().map(tool => {
                return {
                    name: tool.getTitle(),
                    action: tool.callback(layerId, featureId)
                };
            });
        },
```

src/react/components/FeatureEditor/index.js
```diff
@@ -1 +1 @@
-export { FeatureEditorPanel } from './FeatureEditorPanel';
+export { FeatureEditorPanel } from './FeatureEditorPanel.jsx';
```
ZakarFin, line 1: Does this need the suffix on import?

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
```diff
@@ -11,6 +11,7 @@ import './request/SwipeStatusRequest';
 import './request/SwipeStatusRequestHandler';
 
 import { getGfiContent, getGfiResponseType, hasGfiData } from './GfiHelper';
+import { FEATURE_EDITOR_TOOLNAME } from '../../../../framework/myfeatures/constants';
```
ZakarFin: We shouldn't need this here if we do it in more generic way.

bundles/framework/myfeatures/constants.js
```diff
@@ -20,3 +20,5 @@ export const ERRORS = {
     NO_SRS: 'unknown_projection',
     FORMAT: 'format_failure'
 };
+
+export const FEATURE_EDITOR_TOOLNAME = 'FeatureEditor';
```
ZakarFin: Should we prefix this with `myfeatures.FeatureEditor`? in case we want to use the same tool name for content-editor at some point.

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
ZakarFin: Should we check that featureId is present in the contentData and only then add the featuretools?

bundles/mapping/mapmodule/domain/AbstractLayer.js
```diff
@@ -765,6 +768,45 @@ Oskari.clazz.define(
                 return tool.getName() === toolName;
             });
         },
+        /**
```
ZakarFin, line 771: These should logically be in the `wfslayer` model instead of the `AbstractLayer`. But then we would have to check if the functions are present before using them or checking the `instanceof` layer object. Not sure if it's worth it to move them. Also VectorLayer (runtime features) or VectorTileLayer might not extend the wfslayer model so this might be a better place after all 🤔

bundles/framework/myfeatures/request/ShowFeatureEditorRequest.js
ZakarFin: Or we might not actually need it if we just pass a callback function for the feature tools. Let's remove the request instead for now.

bundles/mapping/mapmodule/plugin/getinfo/GetInfoPlugin.js
DenverCoder544: Lets what now? :D EDIT: OK, after the edit it's making more sense :D

bundles/framework/myfeatures/request/ShowFeatureEditorRequest.js
DenverCoder544: I added the .js suffix and the tool callback now calls the request. I think we should keep the request, since this probably isn't the only occasion where it might be useful to be able to open up feature editor.

bundles/mapping/mapmodule/domain/AbstractLayer.js
DenverCoder544: Keeping this as is for the time being then :)

src/react/components/FeatureEditor/index.js
DenverCoder544: There seems to be some glitch (in our webpack configs probably) which makes the dev server whine about a missing export when the suffix isn't there.

PR #2934, opened 2026-01-15
Add new myfeatures layer

DenverCoder544:
Please note, vänligen observera that this pr contains everything implemented in https://github.com/oskariorg/oskari-frontend/pull/2933 so it might be a good idea to merge that first.

In this pr there's functionality to add a new empty myfeatures layer. Also added a button to add a new feature to an already saved layer, so that adding the first feature to a new layer is possible somehow, before we add functionality to do this from the toolbar. We might wanna remove the button later on. 

image

image

image

image

### DIFF
--- a/bundles/framework/myfeatures/handler/MyFeaturesHandler.js
+++ b/bundles/framework/myfeatures/handler/MyFeaturesHandler.js
@@ -173,7 +173,8 @@ class MyFeaturesHandler extends StateHandler {
                 },
                 style: {
                     ...layerJson?.options?.styles?.default?.featureStyle
-                }
+                },
+                layerFields: layerJson?.layerFields
             };
             this.showLayerDialog(values);
         } catch (err) {
--- a/bundles/framework/myfeatures/handler/MyFeaturesHandler.js
+++ b/bundles/framework/myfeatures/handler/MyFeaturesHandler.js
@@ -54,7 +54,8 @@ class MyFeaturesHandler extends StateHandler {
         const save = values => this.importFile(values);
         const update = values => this.updateLayer(id, values);
         const onOk = isNew ? newCallback : isImport ? save : update;
-        this.popupControls = showLayerForm(values, conf, onOk, () => this.popupCleanup());
+        const addNewFeature = () => { this.popupCleanup(); this.showFeatureEditorDialog(id); };
+        this.popupControls = showLayerForm(values, conf, onOk, () => this.popupCleanup(), addNewFeature);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


bundles/framework/myfeatures/handler/MyFeaturesHandler.js
```diff
@@ -50,10 +50,12 @@ class MyFeaturesHandler extends StateHandler {
             unzippedMaxSize: this.getMaxSize() * 15,
             isImport
         };
+        const newCallback = (values) => this.createLayer(values);
         const save = values => this.importFile(values);
         const update = values => this.updateLayer(id, values);
-        const onOk = isImport ? save : update;
-        this.popupControls = showLayerForm(values, conf, onOk, () => this.popupCleanup());
+        const onOk = isNew ? newCallback : isImport ? save : update;
```
ZakarFin: This is a bit wild, but with enough head scratching it compiles eventually ;)

bundles/framework/myfeatures/view/LayerForm/LayerFieldsTab.jsx
```diff
@@ -0,0 +1,112 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Select, Message, TextInput } from 'oskari-ui';
+import { Table } from 'oskari-ui/components/Table';
+import { PrimaryButton, DeleteButton } from 'oskari-ui/components/buttons';
+import styled from 'styled-components';
+const types= ['Boolean', 'Integer', 'Double', 'String', 'Date', 'Timestamp', 'UUID'];
```
ZakarFin, line 7: We might want to have these in some constant on the featureEditor/vector feature/describeFeature side so we can use the same values for formatting/input type for feature properties. But before we figure out where these should be, we can have them here. Just have to remember we have these here as well.

bundles/framework/myfeatures/handler/MyFeaturesHandler.js
```diff
@@ -50,10 +50,24 @@ class MyFeaturesHandler extends StateHandler {
...
+            if (isNew) {
+                return newCallback;
+            };
+
+            if (isImport) {
+                return saveCallback;
+            }
+
+            return updateCallback;
+        };
```
ZakarFin: Not sure if we need to have the variables?
```suggestion
        const getOkCallback = () => {
            if (isNew) {
                return (values) => this.createLayer(values);
            };
            if (isImport) {
                return (values) => this.importFile(values);
            }
            return (values) => this.updateLayer(id, values);
        };
```

DenverCoder544: Probably not. I'm not sure which version is easier to read but neither one is rocket science.

PR #2938, opened 2026-01-26
keep temp feature on map until explicitly discarded or saved

DenverCoder544:
keep updated temp geometry on map until excplicitly discarded or saved. The layout changed a bit and replaced the action link with a normal button.

image

image

image

multi-geometry mode layout stayed the same apart from the link -> button change
Multipolygon:

image

image

### DIFF
--- a/src/react/components/FeatureEditor/DrawingHelper.js
+++ b/src/react/components/FeatureEditor/DrawingHelper.js
@@ -20,13 +20,19 @@ const fakeModule = {
 };
 
 const startDrawing = (type, isMulti = false, currentGeometry, listener) => {
+
+    const sandbox = Oskari.getSandbox();
+
+    // clear old drawing in case there might a previous sketch on the map
+    sandbox.postRequestByName('DrawTools.StopDrawingRequest',
+        [DRAW_OPERATION_ID, true, true]);
+
     const drawParams = {
         allowMultipleDrawing: true,
         showMeasureOnMap: true,
         geojson: currentGeometry,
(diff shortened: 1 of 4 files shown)


src/react/components/FeatureEditor/DrawingHelper.js
```diff
@@ -20,24 +20,31 @@ const fakeModule = {
 };
 
 const startDrawing = (type, isMulti = false, currentGeometry, listener) => {
+
+    const sandbox = Oskari.getSandbox();
+
+    // clear old drawing in case there might a previous sketch on the map
+    sandbox.postRequestByName('DrawTools.StopDrawingRequest',
+        [DRAW_OPERATION_ID, true, true]);
```
ZakarFin: Can we just call the existing function to do this?
```suggestion
    stopDrawing(true);
```

src/react/components/FeatureEditor/styled.jsx
```diff
@@ -8,10 +8,22 @@ export const StyledSpace = styled(Space)`
 
 export const StyledContainer = styled('div')`
     display: inline-flex;
-    justify-content: space-between;
+    justify-content: space-between;
```
ZakarFin:
```suggestion
    justify-content: space-between;
```

src/react/components/FeatureEditor/DrawingHelper.js
DenverCoder544: Glad you asked. Yes, yes we definitely could! :)

PR #2943, opened 2026-02-02
Featureeditor from featuredata

DenverCoder544:
Open feature editor from featuredata table. The createFeatureToolsColumn - method is not might be something we wanna make more general at some point if we have more feature-specific tools to add etc. But that requires for instance that the tools share the same arguments to their callback functions etc.

image

### DIFF
--- a/bundles/framework/featuredata/view/FeatureDataContainer.jsx
+++ b/bundles/framework/featuredata/view/FeatureDataContainer.jsx
@@ -11,7 +11,6 @@ import { ExportButton } from './ExportData';
 import { CompressedView } from './CompressedView';
 import { FEATURE_EDITOR_TOOLNAME } from '../../myfeatures/constants';
 import { IconButton } from 'oskari-ui/components/buttons';
-import { EditOutlined } from '@ant-design/icons';
 
 export const FEATUREDATA_BUNDLE_ID = 'FeatureData';
 export const FEATUREDATA_WFS_STATUS = { loading: 'loading', error: 'error' };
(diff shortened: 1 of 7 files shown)


bundles/mapping/mapmodule/service/map.layer.js
```diff
@@ -299,10 +299,17 @@ Oskari.clazz.define('Oskari.mapframework.service.MapLayerService',
...
-            this.addLayer(newLayer, true);
+            const shouldAddToSelectedMapLayers = !!this.getSandbox().findMapLayerFromSelectedMapLayers(layerId);
+            this.removeLayer(layerId, true, false);
+            this.addLayer(newLayer, false);
+
             const mapLayerEvent = Oskari.eventBuilder('MapLayerEvent');
             this.getSandbox().notifyAll(mapLayerEvent(layerId, 'update'));
+
+            if (shouldAddToSelectedMapLayers) {
+                this.getSandbox().postRequestByName('AddMapLayerRequest', [layerId]);
```
ZakarFin: I think we should make what ever requires this to listen to MapLayerEvent (with update operation) instead. This will change the order of layers. It might be used when admin updates the layer, but at that point it doesn't matter as much if the order changes. However the end-user might get confused why it changed

bundles/framework/featuredata/view/FeatureDataContainer.jsx
```diff
@@ -171,6 +190,13 @@ const createColumnSettings = (selectedFeatureIds, showSelectedFirst, showCompres
                 ellipsis: true
             };
         });
+
+    if (layer?.getFeatureTool(FEATURE_EDITOR_TOOLNAME)) {
```
ZakarFin: Lets make this more generic with if there are any featureTools (not just the specific one) -> create the column.

bundles/framework/featuredata/view/FeatureDataContainer.jsx
```diff
@@ -141,9 +144,25 @@ const createFeaturedataGrid = (features, selectedFeatureIds, showSelectedFirst,
...
-const createColumnSettings = (selectedFeatureIds, showSelectedFirst, showCompressed, sorting, visibleColumnsSettings) => {
+const createFeatureToolsColumn = (layer) => {
+    const featureEditorTool = layer?.getFeatureTool(FEATURE_EDITOR_TOOLNAME);
+    return {
+        align: 'left',
+        title: <Message bundleKey={FEATUREDATA_BUNDLE_ID} messageKey='table.featureTools.title' />,
+        render: (item) => {
+            return <IconButton
+                title={featureEditorTool.getTitle()}
+                icon={<EditOutlined />}
```
ZakarFin: Could we add some generic icon reference to the tool (if it doesn't have already) and make this more generic in a way that if we recognize the icon "type" (might be called class on the tools) -> use a predefined icon, otherwise show the title of the tool? Then the tools handling could work with any feature tools with either showing the edit button or the title for non-edit tools (like removal tool as a later implementation of a feature tool).

bundles/mapping/mapmodule/service/map.layer.js
DenverCoder544: The problem wasn't in listening the event but the fact that nobody notifies the map itself of the changes happening. I implemented a rudimentary mergeLayer - functionality to map state which at least does the bare minimum we need for the featuretools to appear and the layerType to be correct.

PR #2947, opened 2026-02-16
add field presentation segment to fields tab

DenverCoder544:
Backend pr: nnn

add fields formatting/localisation section to myfeatures. These affect eg. infobox and featuredata but not (as of yet) feature editor. 

image

image

image

image

image


src/react/components/VectorLayerPresentation/LayerHelper.js
```diff
@@ -0,0 +1,225 @@
+// oskari-server WFSLayerAttributes
+export const DATA = {
```
ZakarFin: This file is propably not needed (if so, lets remove it from here). Otherwise looks good for merging 👍

ZakarFin:
Related server-side PR: https://github.com/oskariorg/oskari-server/pull/1238

src/react/components/VectorLayerPresentation/LayerHelper.js
DenverCoder544: yeah, that one we didn't use

PR #2953, opened 2026-02-27
Myfeatures ux

DenverCoder544:
Includes commits from a prior pr: https://github.com/oskariorg/oskari-frontend/pull/2952 

option to add new layer from featureeditor

image


src/react/components/FeatureEditor/LayerSelectionPanel.jsx
```diff
@@ -12,14 +12,14 @@ const Row = styled('div')`
     display: flex;
     justify-content: flex-start;
     gap: 1em;
+    padding-bottom: 1em;
 `;
 
 const BUNDLE_KEY = 'oskariui';
 
-export const LayerSelectionPanel = ({ setCurrentLayer }) => {
+export const LayerSelectionPanel = ({ myFeaturesLayers, setCurrentLayer, addNewLayer = null }) => {
```
ZakarFin: I think we should actually remove the layer selection panel from the FeatureEditor as it doesn't really need to be part of the *feature editor* and references to `myfeatures` functionality start to creep up on the feature editor because of it. Also the layer selection only really makes sense with the `myfeatures` functionality at the moment. If we were to add other kinds of layers to the draw buttons on the toolbar, it still shouldn't be part of the generic FeatureEditor UI-component from `oskariui`. The layer selection has a better fit for being closer to the functionality whose features are being added (in this case `myfeatures`).

Ideally the feature editor should not care about where the vector feature that is being edited came from nor what happens to the feature once it has been edited/"committed" as ready or canceled from the editor. As such it shouldn't really care about the layers or such things, only the metadata about vector feature properties of the layer that the edited feature came from (so we can use the information about the property types, labels and such. I'm not even sure if we should pass the OskariLayer to the editor or only the vector feature to be edited and any relevant metadata that might be present on the layer object. This way it's much easier to use the same UI for the content-editor functionality as well.

PR #2960, opened 2026-03-09
move images under request and update paths

DenverCoder544:
move images / update paths to work in oskari.org


ZakarFin:
Should fix issues in https://oskari.org/documentation/api/requests/unreleased/AddMarkerRequest 
image

PR #2962, opened 2026-03-10
FeatureEvent: check for touch event/hitTolerance when gathering features on click

ZakarFin:
Use getFeaturesAtPixel() instead of forEachFeatureAtPixel() to allow using touch and refactor accordingly. The layer handling is kinda weird. We check for type, but at least wfs-layers don't have a specific handler to call (maybe mvt does). It's the only reason to map layerId -> ol layer -> back to layerId.

getFeaturesAtPixel() already does the cluster check and attaches all features on cluster to the list so this simplified a bit, but the downside is the awkward layer fiddling.

### DIFF
--- a/bundles/mapping/mapmodule/service/VectorFeatureService.ol.js
+++ b/bundles/mapping/mapmodule/service/VectorFeatureService.ol.js
@@ -296,18 +296,29 @@ Oskari.clazz.defineES('Oskari.mapframework.service.VectorFeatureService',
          */
         _onMapClicked (event) {
             const me = this;
-            let clickHits = [];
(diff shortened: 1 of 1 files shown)


ZakarFin:
This works like before. I'm likely to make another change in the near future that would filter out features from hover layer etc. from `mapmodule.getFeaturesAtPixel()` so we wouldn't need to filter out features that don't have a layer tied to them here.

ZakarFin:
There's a problem with this change when there are layers on the map that have been added through RPC (runtime layers) that are not registered to maplayer service. The features returned by map have a layerId, but getOlMapLayer() returns null for those as it uses sandbox.findMapLayerFromSelectedLayers().

PR #2963, opened 2026-03-13
Timeslider fix

okauppinen:
If dataPoints is missing (empty list) => sliderPoints [] => can't find snapPoint (undefined) => js error

Fix admin-layereditor timeseries metadata layer select's search. No children => js error
Could also use own `filterOption={(input, option) => option.label?.toLowerCase().includes(input.toLowerCase())}` but antd default search is case insensitive


ZakarFin:
The WMS-T timeseries UI breaks kinda spectularly as well if the timeseries has only one value. Not sure if this UI handles the situation any better.

PR #2964, opened 2026-03-13
FeatureEvent: check for touch event - fix clicks for features on runtime layers

ZakarFin:
Fixes an issue raised by #2962 where clicks on features located on runtime layers caused a js error as the layer the feature "clicked" was not on the `sandbox.findMapLayerFromSelectedLayers()` / MapLayerService (most runtime layers are not registered there). Added a note that `MapModule.getOLMapLayers()` does NOT usually return any runtime layers.

`MapModule.getFeaturesAtPixel()` now returns the layer implementation (OL layer) in addition to the feature implementation (OL VectorFeature), feature properties and layerId. Note that layerId can be null for internal layers like the layer we use for visualizing features that are hovered over by the cursor. Added special handling for the additional hover feature so it's not returned on `getFeaturesAtPixel()` as the actual feature that the hover is duplicating is already in the list and has the correct layer id.

Added a bit of documentary comments about the lessons learned with this change.

### DIFF
--- a/bundles/mapping/mapmodule/MapModuleClass.ol.js
+++ b/bundles/mapping/mapmodule/MapModuleClass.ol.js
@@ -22,7 +22,7 @@ import 'ol/ol.css';
 import { OskariImageWMS } from './plugin/wmslayer/OskariImageWMS';
 import { getOlStyles, getOlStyleForLayer, setDefaultStyle } from './oskariStyle/generator.ol';
 import { STYLE_TYPE } from './oskariStyle/constants';
-import { LAYER_ID } from '../mapmodule/domain/constants';
+import { LAYER_ID, LAYER_HOVER } from '../mapmodule/domain/constants';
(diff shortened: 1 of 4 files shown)


bundles/mapping/mapmodule/MapModuleClass.ol.js
```diff
@@ -361,6 +362,11 @@ export class MapModule extends AbstractMapModule {
             opts.hitTolerance = 10;
         }
         this.getMap().forEachFeatureAtPixel([x, y], (feature, layer) => {
+            if (layer.get(LAYER_HOVER) === true) {
+                // we don't want to process hits for our hover layer as
+                //  it duplicates features that are under it/oin the same pixel on the _actual layer_
```
ZakarFin:
```suggestion
                //  it duplicates features that are under it/on the same pixel on the _actual layer_
```

PR #2965, opened 2026-03-16
Myfeatures add new layer

DenverCoder544:
-swap input / table placement
-use real names of datatypes in table
-field type always string by default
-when creating a layer from scratch, add field "name" and it's localisations by default so creating a layer is possible even without visiting fields - tab.
-fix some bugs related to state handling between presentation popups and the rest of the dialog.

image

image

image

### DIFF
--- a/bundles/framework/myfeatures/view/LayerForm/LayerFormContent.jsx
+++ b/bundles/framework/myfeatures/view/LayerForm/LayerFormContent.jsx
@@ -49,11 +49,28 @@ const getDefaultAttributes = (layerFields) => {
     };
 };
 
+export const DEFAULT_FIELD = 'name'
+export const DEFAULT_TYPE = 'String';
+
+const getDefaultLayerFields = () => {
+    return [{
+        name: DEFAULT_FIELD,
+        type: DEFAULT_TYPE
+    }];
+}
+
 export const LayerFormContent = ({ values, config, onOk, onCancel, error, addFeature }) => {
     // TODO: refactor this thing as it's getting way overly complicated
     const { maxSize, unzippedMaxSize, isImport } = config;
     const { style = Oskari.custom.generateBlankStyle(), locale = {} } = values || {};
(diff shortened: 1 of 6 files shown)


bundles/framework/myfeatures/view/LayerForm/LayerFormContent.jsx
```diff
@@ -35,16 +35,34 @@ const getTabTitle = (isValid, messageKey) => (
...
+    return [{
+        name: DEFAULT_FIELD,
+        type: DEFAULT_TYPE
+    }];
+};
+
+const getDefaultLocale = () => {
+    return {
+        en: {
+            name: 'Name'
```
ZakarFin: We should get these from localizations, but not sure how we could do that currently :)

src/react/components/VectorLayerPresentation/VectorLayerPresentation.jsx
```diff
@@ -58,6 +60,24 @@ export const VectorLayerPresentation = ({ layer, updateAttributes, updateFeature
         featureFilter
     });
 
+    useEffect(() => {
```
ZakarFin, line 70: Can we make an attempt to try this some other way?

bundles/framework/myfeatures/view/LayerForm/LayerFormContent.jsx
DenverCoder544: "hacky fallback fix" to en locale made.

src/react/components/VectorLayerPresentation/VectorLayerPresentation.jsx
DenverCoder544: If we want to continue using the same component for admin tools and myfeatures we should probably be moving most of the current state handling to a new wrapper component to be able to get rid of the useEffect magic. This would also probably ease the pain regarding maintaining draft / baseline states. Currently syncinc those is a pain and very prone to error.

However, until we've come to a decision of where we want to go with this functionality it is probably not wise to use any more effort refactoring this.

PR #2972, opened 2026-04-09
Myfeatures UI vol many

DenverCoder544:
Refactor the way adding a new myfeatures layer works for the nth time where n is plenty.

*toggle field visibility and moving up / down now handled directly from the list
*separate visibility settings for different locales can no longer be handled from this view
*open locale/formatting modal directly ffrom table, by default only show this one field with the possibility to allow editing all other localizations as well

image

image

image

image


ZakarFin:
There's a lot of statehandling code in `bundles/framework/myfeatures/view/LayerForm/LayerFieldsTab.jsx‎` that we should try to refactor away. But lets merge for now and get feedback on current version.

PR #2982, opened 2026-04-29
show external source icon for rss annoucenments

DenverCoder544:
Indicate announcement from an external feed with an icon on the list (for admins). No other changes yet. We need to figure out what we should be able to edit, what happens when we delete and when the scheduled job updates from rss. 

image


ZakarFin:
Related to https://github.com/oskariorg/oskari-server/pull/1266

PR #2983, opened 2026-05-04
UI fixes

DenverCoder544:
* fix the bug of drawing not ending correctly when button pressed
* fix modal padding that antd update broke.
image

*replace the ugly buttons with links and add group
image

* bonus: sort announcements by begin date

### DIFF
--- a/src/react/components/FeatureEditor/DrawingHelper.js
+++ b/src/react/components/FeatureEditor/DrawingHelper.js
@@ -49,20 +49,13 @@ const startDrawing = (type, isMulti = false, currentGeometry, listener) => {
     sandbox.registerForEventByName(fakeModule, EVENT_NAME);
 
 };
+
 const stopDrawing = (clearPrevious = false, finishDrawing = false) => {
     const sandbox = Oskari.getSandbox();
-    if (finishDrawing) {
-        // Call DrawPlugin.stopDrawing() directly with suppressEvent=false.
-        // This runs forceFinishDrawing() (handles in-progress sketch, trims ghost point)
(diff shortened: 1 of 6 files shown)


src/react/components/FeatureEditor/DrawingHelper.js
```diff
@@ -39,13 +49,22 @@ const startDrawing = (type, isMulti = false, currentGeometry, listener) => {
...
-    sandbox.unregisterFromEventByName(fakeModule, EVENT_NAME);
-    drawListener = null;
+    if (finishDrawing) {
+        // Call DrawPlugin.stopDrawing() directly with suppressEvent=false.
+        // This runs forceFinishDrawing() (handles in-progress sketch, trims ghost point)
+        // and then sendDrawingEvent(true) — the same path as a double-click on the map.
+        // fakeModule.onEvent receives the isFinished=true event, updates geometry and does cleanup.
+        const drawPlugin = sandbox.findRegisteredModuleInstance('DrawTools')?.getPlugin();
+        drawPlugin?.stopDrawing(DRAW_OPERATION_ID, clearPrevious, false);
+        // Fallback: if event was not sent (e.g. geometry too short), clean up listener here.
```
ZakarFin: Could this be accomplished with the request instead or is there a reason we need to call the plugin directly?

DenverCoder544: Yes it could, as it turns out. The logic was a bit messy and there was an async bug with an unnecessary cleaning of the listener which should now be fixed.

PR #2989, opened 2026-05-13
Fix publiccode.yml error and add CI workflow for validation

bfabio:


publiccode.yml
```diff
@@ -16,7 +16,9 @@ categories:
   - web-collaboration
   - data-visualization
 
-organisation: "National Land Survey of Finland"
+organisation:
+  name: "National Land Survey of Finland"
+  uri: "https://www.maanmittauslaitos.fi"
```
bfabio, line 21: If you have a better URI to represent your organization, like some sort of centrally registered code or URN, feel free to use it and replace this one.

ZakarFin:
Thanks!

PR #2990, opened 2026-05-15
Layer admin url params

DenverCoder544:
* Add a separate params - section in layeradmin to manage url params for layer.
* Reserved parameter SERVICE, VERSION & REQUEST are cleaned out as before and added to outbound requests under the hood. Adding said params isn't allowed using the table either.
* removed the params - collapse from under JSON tab.

TODO - regarding copy to clipboard functionality: should we be adding said params to url when copying and should we display those inside the params table (not editable)?

Add a new fresh layer:
image
   
params snatched from url input field:
image

layer selected and in edit view:
image

reserved params not allowed
image

### DIFF
--- a/bundles/admin/admin-layereditor/resources/locale/en.js
+++ b/bundles/admin/admin-layereditor/resources/locale/en.js
@@ -41,7 +41,11 @@ Oskari.registerLocalization(
                 },
                 "opacity": "Opacity",
                 "params": {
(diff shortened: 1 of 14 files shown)


ZakarFin:
I think we can just safely drop the  reserved parameters from the table and as such they wouldn't be added to the url from copy to clipboard button.

The clipboard button should give full url with protocol and parameters from the params table.

PR #3000, opened 2026-05-27
show warning message when there might be more features on the layer t…

DenverCoder544:
show warning message when there might be more features on the layer than were loaded

image

### DIFF
--- a/bundles/mapping/mapmodule/plugin/wfsvectorlayer/WfsVectorLayerPlugin/impl/VectorLayerHandler.ol.js
+++ b/bundles/mapping/mapmodule/plugin/wfsvectorlayer/WfsVectorLayerPlugin/impl/VectorLayerHandler.ol.js
@@ -7,6 +7,7 @@ import { createXYZ } from 'ol/tilegrid';
 
 import { AbstractLayerHandler, LOADING_STATUS_VALUE } from './AbstractLayerHandler.ol';
 import { RequestCounter } from './RequestCounter';
+import { Messaging } from 'oskari-ui/util';
 
 import olPoint from 'ol/geom/Point';
 import olMultiPoint from 'ol/geom/MultiPoint';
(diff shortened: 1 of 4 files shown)


bundles/mapping/mapmodule/plugin/wfsvectorlayer/WfsVectorLayerPlugin/impl/VectorLayerHandler.ol.js
```diff
@@ -205,6 +208,21 @@ export class VectorLayerHandler extends AbstractLayerHandler {
         };
     }
 
+    _showMaybeMoreFeaturesWarning (layer, jqXHR) {
+        const layerId = layer.getId();
+        if (this.maybeMoreFeaturesWarned.has(layerId)) {
```
ZakarFin, line 215: We should probably reset this at some point. It makes sense to only show it once per viewport, but if the user looks at another location on the map that also has the header, the note should pop up again.

I'm not sure what's a good check for this. We could store the extent that triggered this and use it as key besider the layerId, but that would mean the maybeMoreFeaturesWarned Set will grow as long as the user is on the page. We could try resetting this when on the `updateLoadingStatus(LOADING_STATUS_VALUE.COMPLETE);` or the MapStartedMoving/AfterMapMoveEvents?

ZakarFin: This one triggers when the map starts moving (maybe reset the notification at that point?) https://github.com/oskariorg/oskari-frontend/blob/3.3.0/bundles/mapping/mapmodule/MapModuleClass.ol.js#L166

Worst case we could reset the set with some timer, but that's not a good option either. We don't want it to be too spammy (trigger multiple notifications per view port), but also it's a bit weird if we only notify once in any session per layer

DenverCoder544: Yeah, good point. I'm not sure what would be the ideal way of resetting this as it might also get annoying if we reset the warning on every pan. But let's try MapStartedMoving for resetting the set and see how that pans out.

PR #3005, opened 2026-06-04
Feature editor bringtofront

DenverCoder544:
change feature editor from flyout to popup to ensure it stays on top even when there are other (jquery) flyouts opened. Also change the confirmation dialog which was antd's notification to oskari modal.

image

image

### DIFF
--- a/src/react/components/FeatureEditor/EditConfirmation.jsx
+++ b/src/react/components/FeatureEditor/EditConfirmation.jsx
@@ -1,22 +1,41 @@
 import React from 'react';
-import { notification } from 'antd';
+import { showModal } from 'oskari-ui/components/window';
 import { Button, Message } from 'oskari-ui';
+import { styled } from 'styled-components';
+
+const Buttons = styled('div')`
+    display: flex;
+    flex-direction: row;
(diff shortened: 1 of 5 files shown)


src/react/components/FeatureEditor/EditConfirmation.jsx
```diff
@@ -1,22 +1,41 @@
...
+    const content = (
+        <Content>
+            <Message bundleKey={'oskariui'} messageKey={'FeatureEditorView.editConfirm.msg'}/>
+            <Buttons>
+                <Button type='primary' onClick={() => { onConfirm(true); close(); }}>
+                    <Message bundleKey={'oskariui'} messageKey={'FeatureEditorView.buttons.yes'}/>
+                </Button>
+                <Button onClick={close}>
+                    <Message bundleKey={'oskariui'} messageKey={'FeatureEditorView.buttons.cancel'}/>
+                </Button>
```
ZakarFin, line 35: There might be predefined buttons in `oskari-ui` that have the message included so we wouldn't need to use the child Message-tag with the buttons, but lets go with this and maybe update in a future pr.

PR #3009, opened 2026-06-08
fix mergelayer to use correct opacity and visibility

DenverCoder544:
preserve opacity and visibility in case my features layer was already in selectedlayers and we needed to merge values after refresh.

image

### DIFF
--- a/bundles/mapping/mapmodule/service/map.state.js
+++ b/bundles/mapping/mapmodule/service/map.state.js
@@ -423,10 +423,15 @@ import { UnsupportedLayerReason } from '../domain/UnsupportedLayerReason';
         mergeLayer: function(id, newLayer) {
             const index = this.getLayerIndex(id);
             if (index > -1) {
-                // do this to preserve layer specific stuff such as getFeatureCount on myfeatures layer
-                // when cloning
+                const current = _selectedLayers[index];
+                const opacity = current.getOpacity();
+                const visibilityInfo = current.getVisibilityInfo();
                 const emptyLayer = Object.create(Object.getPrototypeOf(newLayer));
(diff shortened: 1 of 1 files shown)


bundles/mapping/mapmodule/service/map.state.js
```diff
@@ -423,10 +423,15 @@ import { UnsupportedLayerReason } from '../domain/UnsupportedLayerReason';
...
             if (index > -1) {
-                // do this to preserve layer specific stuff such as getFeatureCount on myfeatures layer
-                // when cloning
+                const current = _selectedLayers[index];
+                const opacity = current.getOpacity();
+                const visibilityInfo = current.getVisibilityInfo();
                 const emptyLayer = Object.create(Object.getPrototypeOf(newLayer));
-                _selectedLayers[index] = Object.assign(emptyLayer, newLayer);
+                const merged = Object.assign(emptyLayer, newLayer);
+                // TODO: we should move to a direction where maplayer and selectedlayer are in fact the same instance so there would be no need for this hacky and error prone copying props one by one
```
ZakarFin, line 431: I think we should separate the layer information on MapLayerService and the state information for layers that are on the map. The maplayerservice would then hold the information about the layers as they exist with the defaults and such where as runtime state like opacity and visibility state would be handled separately. That way it would be easier to reset runtime state for layers to defaults when the layer is added to the map. This is related to a known issue when a layer that was hidden from the map is  _removed_ and the _re-added_ to the map -> the layer is still hidden, but the user might be confused why the added layer is not shown on the map (until they open the selected layers and see it's on the map but hidden).

But that's a bigger task to do and we would have to evaluate and test what might be caused by such a change.

PR #3012, opened 2026-06-11
Userguide react

DenverCoder544:
implement userguide as es/react. Also add a link to open the original user guide url in a new window.

image


bundles/framework/userguide/instance.js
```diff
@@ -1,228 +1,143 @@
...
-        /**
-         * @property {String[]} protocol
-         * @static
-         */
-        'extend': ['Oskari.userinterface.extension.DefaultExtension']
-    });
+    }
+}
+
+Oskari.clazz.defineES('Oskari.mapframework.bundle.userguide.UserGuideBundleInstance', UserGuideBundleInstance);
```
ZakarFin: We could just export UserGuideBundleInstance instead and import it on index.js?

PR #3014, opened 2026-06-15
fix organisationname not showing for myf layers and add wrapper

DenverCoder544:
make sure organisationname is shown for myfeatures layers as well, even if they were already in view when the view was (re)loaded.

wrap to a <span class="myfeatures_organisation-name"/> to be able to override / hide this in other apps.

layer.getSource never seems to be truthy, so the condition of checking whether to use source or organisationname always seems to fall back to organisation. But I didn't dare remove that just in case there might be some side effects I hadn't considered.

image

### DIFF
--- a/bundles/mapping/mapmodule/plugin/logo/DataProviderPopup.jsx
+++ b/bundles/mapping/mapmodule/plugin/logo/DataProviderPopup.jsx
@@ -13,21 +13,31 @@ const DataSection = styled('div')`
 `;
 
 const createLink = (item) => {
-    if (typeof item === 'string') return item;
-    if (!item || !item.name) return null;
-    if (!item.url) return item.name;
-    return <a href={item.url} target='_blank'>{item.name}</a>;
+    if (typeof item === 'string') {
+        return item;
+    }
+    if (!item || !item.name) {
+        return null;
+    }
+    if (!item.url && !item.className) {
+        return item.name;
+    }
+    if (!item.url) {
+        return <span className={item.className}>{item.name}</span>;
+    }
+    return <a className={item.className} href={item.url} target='_blank'>{item.name}</a>;
 };
 
 const formatSource = (src) => {
     let source = [];
-    if (!src) return source;
-    if (Array.isArray(src)) {
-        source = src.map(s => createLink(s));
+    if (!src) {
         return source;
-    } else {
-        source.push(createLink(src));
     }
+    if (Array.isArray(src)) {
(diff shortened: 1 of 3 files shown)


bundles/mapping/mapmodule/plugin/logo/DataProviderPopup.jsx
```diff
@@ -13,21 +13,31 @@ const DataSection = styled('div')`
...
+        return item;
+    }
+    if (!item || !item.name) {
+        return null;
+    }
+    if (!item.url && !item.className) {
+        return item.name;
+    }
+    if (!item.url) {
+        return <span className={item.className}>{item.name}</span>;
```
ZakarFin, line 26: Does `item.name` include both the "Omat aineistot" and the source the user has given? The purpose of the CSS-hidable part was to be able to show the source user has given without the "Omat aineistot" prefix.

bundles/mapping/mapmodule/plugin/logo/LogoPlugin.js
```diff
@@ -127,11 +122,18 @@ Oskari.clazz.define(
                 }
             };
         },
+        getMyFeaturesSource: function (layer) {
+            return { name: layer.getOrganizationName(), className: 'myfeatures-organisation-name' };
+        },
         getItemFromLayer: function (layer) {
+            const isMyFeatures = layer.getLayerType && layer.getLayerType() === 'myf';
```
ZakarFin, line 129: Is there a way we could do this on the myfeatures bundle instead so we would not need to reference myfeatures in the logoplugin?

ZakarFin:
I feel like we need to revisit the specs for this change. The main purpose is to show what the user used in the "source" field of the layer and be able to hide the additional "My features" text with CSS  if the application using the functionality so chooses (wrap the optionally hideable part in a span with a class so it can be easily targeted with styling).

However, showing the source information for the myfeatures layers is mostly useful when they are shown as part of a published map. Myfeatures layers are presented as WFS-layers for the frontend (by the server) when being shown on a published map. So we can't tie this to layers of type `myfeatures`. Additionally, even for layers of type `myfeatures` apparently the source is not available for the frontend even if the user added it. So there are multiple issues related to this that we need to solve.

This PR looks ok for merging, but it doesn't solve the issues we are trying to solve. So before merging this we need to figure out how we can solve the core issues related to this.

ZakarFin:
Implemented in #3018

PR #3031, opened 2026-08-20
Updated localizations for myfeatures

ZakarFin:
Initial changes to localizations for myfeatures.


ZakarFin:
There should be more coming, but lets merge now to have a look how it feels.

PR #3040, opened 2026-09-02
Jquery component cleanup

DenverCoder544:
remove a bunch of jQuery ui components not used anymore.


ZakarFin:
Nice 💪 I'll let this sit for a while and think if we can merge it for 3.4 or should we wait until Oskari 4.0

PR #3042, opened 2026-09-10
Myfeatures bugs

DenverCoder544:
- fix sorting for dcesc- field.
- use the value provided in "data source" field for the layer in the list of myfeatures layers. Not quite sure if this breaks stuff elsewhere given the fuzzy logic by which organisation name or datasource are determined elsewhere.  
-also add value for "created" field

image

image


ZakarFin:
I think using the orgName as key in layer json and setOrganizationName() breaks or will break the layer listing functionality. Maybe we should add some field next to featureCount here https://github.com/DenverCoder544/oskari-frontend/blob/2f8534c6ad3761d21e92f71a37c6d9267e8285e1/bundles/framework/myfeatures/domain/MyFeaturesLayer.js#L11 and use that instead of get/setOrganizationName (orgName).

PR #3043, opened 2026-09-14
Getlocalization cleanup

DenverCoder544:
**Contains commits from https://github.com/oskariorg/oskari-frontend/pull/3040 so we need to merge that first.**

This one replaces (at least most) calls to Oskari.getLocalization with Oskari.getMsg().


ZakarFin:
Blocked by #3040 - merging for Oskari 4.0

PR #3044, opened 2026-09-15
align input fields in featureeditor

DenverCoder544:
align fields in feature editor form

image


ZakarFin:
Not sure how this works with long labels, but the previous one probably didn't work any better :)

PR #3045, opened 2026-09-15
optionally suppress warnings

DenverCoder544:
clean up the "no builder found for nn" warnings caused by rpc in to browser console. 

If we still want something to appear in console we could have a single log message for all available / all non-available events/requests on rpc side.


ZakarFin:
Probably makes sense that missing requestBuilders are reported by default while events are not. It should be really rare case where something triggers an event that is not provided by that same bundle. But some bundle trying to send a request that is provided by another bundle is fairly common and the logged message helps with debugging.

PR #3046, opened 2026-09-15
one console message for missing requests in rpc

DenverCoder544:
add a single log message for missing request builders


bundles/framework/rpc/instance.js
```diff
@@ -216,6 +216,7 @@ Oskari.clazz.define(
                     available.push(allowedRequests[i]);
                 }
             }
+            Oskari.log('Messages').warn('RPC missing builder for following requests: ', allowedRequests.filter(req => !available.includes(req)).join(', '));
```
ZakarFin: Would this be more user/developer friendly?

```suggestion
            Oskari.log('RPC').warn('Following requests are not available with current configuration: ', allowedRequests.filter(req => !available.includes(req)).join(', '));
```

 Though I wonder if we need to say this at all. There's a function on the RPC-client that can be used to check which requests are available.

PR #3049, opened 2026-09-18
Feature editor improvements

DenverCoder544:
Minor improvements to myfeatures / feature-editor.

* when changing to another feature in feature-editor only ask for confirmation if the feature is actually modified

image

*  add a new tool to myfeatures layerlist, which opens feature-editor for that layer

image

----------------------->>>>>>>

image

* highlight the feature being edited on map as well

image


bundles/framework/myfeatures/handler/MyFeaturesHandler.js
```diff
@@ -392,6 +392,35 @@ class MyFeaturesHandler extends StateHandler {
...
+    }
+
+    async openFeatureData (layerId) {
+        const layer = this.instance.getMapLayerService().findMapLayer(layerId);
+
+        if (!this.sandbox.isLayerAlreadySelected(layerId)) {
+            this.addLayerToMap(layerId);
+        }
+
+        if (!this.layerPropertiesLoaded(layer)) {
```
ZakarFin: DescribeLayer etc are done when the layer is added to the map if it was not there already. Is it necessary to check here since the alternative is to open the feature data table immediately. Maybe we can just always postpone it for 500ms? It doesn't guarantee that the describe layer would be done at that point either so would it be better to fix this somewhere else? Maybe have featuredata handle missing properties and schedule a check for them later on? Does featuredata break if the describe is not there when it's opened?

bundles/framework/myfeatures/service/MyFeaturesService.js
```diff
@@ -98,7 +98,7 @@ export class MyFeaturesService {
 
         const featureEditorTool =  Oskari.clazz.create('Oskari.mapframework.domain.Tool');
         featureEditorTool.setName(FEATURE_EDITOR_TOOLNAME);
-        featureEditorTool.setTitle(Oskari.getMsg('myfeatures', 'featureEditor.title'));
+        featureEditorTool.setTitle(Oskari.getMsg('myfeatures', 'featureEditor.tooltip'));
```
ZakarFin: Is the tooltip shown somewhere else and we want to reuse it here? Should we rename the localization key to something else instead?

bundles/framework/myfeatures/handler/MyFeaturesHandler.js
DenverCoder544: I'll see to this. We can probably remove this. Was kinda using this for a fallback for a completely different "feature" which caused some head-scracthing.

bundles/framework/myfeatures/MyFeaturesList.jsx
```diff
@@ -52,10 +52,16 @@ export const MyFeaturesList = ({ data = [], controller, loading }) => {
             render: (title, item) => {
                 return (
                     <ToolsContainer>
+                        <IconButton
```
ZakarFin: Should we have some isFeatureDataAvailable that returns sandbox.hasHandler('ShowFeatureDataRequest') and only show the button if the functionality is part of the application?

